PR #10, opened 2018-02-05
Fixing renewal of sessions

stefanorr:
Currently, sessions can not be renewed as the response for these is not handled correctly. The response is currently cast to a NullDataResponse, which fails and throws an exception due to the data field being an object not a string. I created a new response class and applied it to the response. I also made the session object reflect the change in token generated by snowflake.

There is also a minor change to fix logging index error.

Fixes issue #16

### DIFF
--- a/Snowflake.Data/Core/HttpUtil.cs
+++ b/Snowflake.Data/Core/HttpUtil.cs
@@ -13,27 +13,14 @@ namespace Snowflake.Data.Core
 {
     class HttpUtil
     {
-        static private HttpClient httpClient;
-
-        static public HttpClient getHttpClient()
-        {
-            if (httpClient == null)
-            {
-                initHttpClient();
-            }
-            return httpClient;
-        }
-
-        static private void initHttpClient()
+        static public HttpClient initHttpClient(TimeSpan timeout)
         {
             // enforce tls v1.2
             ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
             ServicePointManager.UseNagleAlgorithm = false;
             ServicePointManager.CheckCertificateRevocationList = true;
 
-            httpClient = new HttpClient(new RetryHandler(new HttpClientHandler()));
-            // default timeout for each request is 16 seconds
-            //httpClient.Timeout = TimeSpan.FromSeconds(16);
+            return new HttpClient(new RetryHandler(new HttpClientHandler())) { Timeout  = timeout };
         }
 
         class RetryHandler : DelegatingHandler
@@ -50,19 +37,21 @@ namespace Snowflake.Data.Core
                 HttpResponseMessage response = null;
                 int backOffInSec = 1;
 
-                int httpTimeout = (int)requestMessage.Properties["TIMEOUT_PER_HTTP_REQUEST"];
+                TimeSpan httpTimeout = (TimeSpan)requestMessage.Properties["TIMEOUT_PER_HTTP_REQUEST"];
 
                 CancellationTokenSource childCts = null;
-                if (httpTimeout != -1)
-                {
-                    childCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
-                    childCts.CancelAfter(httpTimeout);
-                }
 
                 while (true)
                 {
                     try
-                    {   
+                    {
+                        childCts = null;
+
+                        if (!httpTimeout.Equals(Timeout.InfiniteTimeSpan)) {
+                            childCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                            childCts.CancelAfter(httpTimeout);
+                        }
+
                         response = await base.SendAsync(requestMessage, childCts == null ? 
                             cancellationToken : childCts.Token);
                     }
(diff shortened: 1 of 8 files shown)


howryu:
@stefanorr Thanks for your contribution! And sorry I think I missed renew session part testing.

And right now appveyor will only accepts the branch within snowflake repository since we don't want anyone to abuse our deployments. So please understand that. I will wrap up your change and push to a branch in our repo and run tests. If all looks good, I will push the change.

Snowflake.Data/Core/SFSession.cs
```diff
@@ -185,14 +185,18 @@ internal void renewSession()
...
                 SnowflakeDbException e = new SnowflakeDbException("", 
                     sessionRenewResponse.code, sessionRenewResponse.message, "");
                 logger.Error("Renew session failed", e);
                 throw e;
-            }
+            } 
+            else 
+            {
+				sessionToken = sessionRenewResponse.data.sessionToken;
+			}
```
howryu: I think I used space instead of tab and I forgot it's two spaces or four spaces, but please follow the format. Thanks.

Snowflake.Data/Core/RestResponse.cs
```diff
@@ -78,6 +83,25 @@ internal class AuthnResponseData
...
     }
 
+
+	internal class RenewSessionResponseData {
+
+		[JsonProperty(PropertyName = "sessionToken", NullValueHandling = NullValueHandling.Ignore)]
+		internal string sessionToken { get; set; }
+
+		[JsonProperty(PropertyName = "validityInSecondsST", NullValueHandling = NullValueHandling.Ignore)]
+		internal Int16 masterTokenValidityInSeconds { get; set; }
```
howryu: This is actually validity in seconds for session token (ST)

Line 99 is validity in seconds for master token (MT)

stefanorr: Oops, yeah I though that must have been the case. I just names things incorrectly. In case you didn't see I also incorrectly (or insufficiently) named validityInSecondsMT to validityInSeconds further down. I can change it on this branch for reference if you would like or you can just do it with your new branch. Thanks

Snowflake.Data/Core/SFSession.cs
stefanorr: Will do, thanks.

stefanorr:
Hi there, I have fixed some more issues. I accidentally committed to my master branch so it got added to this pull request, but I guess the process will be the same for you if you approve of it

CPavanKishore:
Hey @stefanorr now its working,I Can able to get all the chunks with out timeout exception,you are awesome man, can you make GetSchemaTable() as well with this

stefanorr:
@CPavanKishore Glad I could help :) I'm a little unsure what you mean for GetSchemaTable(). Are you wanting to be able to get the schema for a specific table name? This isn't possible to implement as you can have multiple tables that are the same inside different schemas. If you want to know what tables are in a given schema that's possible. Or if you want to know what schema the session is in (therefore what schema will be relevant to any table you want to use based on just running queries on that table through your connection), that's also possible.

stefanorr:
@CPavanKishore I have not tried to run this in a web application sorry

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -13,27 +13,14 @@ namespace Snowflake.Data.Core
 {
     class HttpUtil
     {
-        static private HttpClient httpClient;
-
-        static public HttpClient getHttpClient()
```
howryu: I generally don't like this change. You changed it in such a way that every http request will create a new HttpClient class, which could use up a lot of resources. The original implementation will only create a single http client and that class is being shared between http requests. According to https://msdn.microsoft.com/en-us/library/system.net.http.httpclient(v=vs.110).aspx#Anchor_5, httpClient class is thread safe, so concurrent http requests should not be a problem. 

However, there is another bug in original implementation, which is that the initialization to http client singleton is not synchronized, which will lead to a null pointer exception due to the race condition. I will fix this issue in a separate pr.

Snowflake.Data/Core/RestRequestImpl.cs
```diff
@@ -94,36 +89,26 @@ public JObject get(SFRestRequest getRequest)
...
             return JObject.Parse(jsonString.Result);
         }
 
-        private HttpResponseMessage sendRequest(HttpRequestMessage requestMessage, CancellationToken cancellationToken)
+        private HttpResponseMessage sendRequest(HttpRequestMessage requestMessage, TimeSpan timeout)
         {
             try
             {
-                var response = HttpUtil.getHttpClient().SendAsync(requestMessage, cancellationToken)
+                var response = HttpUtil.initHttpClient(timeout).SendAsync(requestMessage)
```
howryu: The timeout here is passed down to the constructor of http client and set to Timeout property of HttpClient. https://msdn.microsoft.com/en-us/library/system.net.http.httpclient.timeout(v=vs.110).aspx
According to the doc, this timeout looks to me is still the per http request, (per http request timeout is passed down using a map HttpRequestMessage.Properties) This timeout is the total timeout when accessing a snowflake rest endpoint (including all the retries). So I think a cancellation token will make more sense here.

howryu:
@stefanorr I revert some of your changes due to the comments I put there for now. If you disagree with me, we can always continue the discussion and apply your change if reasonable.

stefanorr:
@howryu Looks good, thanks.

PR #12, opened 2018-02-14
Update to Readme.md - adding .NET version

billyprodev:
I thought that it would be important to mention which version of dot net you are using to compile this because my .DLL project need to match the same version in order to be able to import from NuGet.
Best,
Billy

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Build
 Prerequisites
 -------------
 This project is developed under Visual Studio 2015. All other version of visual studio is not supported.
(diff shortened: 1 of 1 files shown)


howryu:
Thanks for doing that. To be honest, we plan to port this project to be complied againt NETStandard, which can be used by both .net core or .net framework projects.

howryu:
This is fixed in e7807bcd69c37f92784068719f8ef5b18ff7685f.

Issue #13, opened 2018-02-14
Connector breaks on NULL values


stefanorr:
Every time a null value is received by the connector it breaks due to null value exceptions.

I have submitted a small change in https://github.com/snowflakedb/snowflake-connector-net/pull/10 which fixes this for my purposes, and I think it's the best place to handle it, but I am not knowledgeable enough about the connector to assert this position.

PR #14, opened 2018-02-26
Adjusting type processing to appropriately handle null values

akrock:
This is an effort to fix issue #13 

Made some changes to the SFBaseResultSet to return DBNull.Value when the column contains null.
Sorry I am unable to currently run the test cases that have been created as it seems some files are missing?

### DIFF
--- a/Snowflake.Data/Core/SFBaseResultSet.cs
+++ b/Snowflake.Data/Core/SFBaseResultSet.cs
@@ -1,4 +1,4 @@
-﻿/*
+﻿/*1/
  * Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
  */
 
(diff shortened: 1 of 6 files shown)


Snowflake.Data/Core/SFBaseResultSet.cs
```diff
@@ -24,134 +24,33 @@ abstract class SFBaseResultSet
...
+            switch (type)
             {
                 case SFDataType.DATE:
-                    return SFDataConverter.toDateString(getDateTime(columnIndex), 
+                    var val = GetValue(columnIndex);
+                    if (val == null)
+                        return null;
+                    return SFDataConverter.toDateString((DateTime)val, 
                         sfResultSetMetaData.dateOutputFormat);
+                //TODO: Feels like were missing some implementations here, at least for time?
```
howryu: Yes. Because snowflake provide a parameters like TIMESTAMP_NTZ_OUTPUT_FORMAT, which allowed user to specify output format (detailed in https://docs.snowflake.net/manuals/user-guide/date-time-input-output.html#output-formats). However, it requires some effort to correct implement parser and formatter of the snowflake timestamp format, I decide to defer the implementation when some users actually asked for it.

Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -19,95 +19,100 @@ public enum SFDataType
...
+        internal static object ConvertToCSharpVal(string srcVal, SFDataType srcType, Type destType)
         {
-            logger.TraceFormat("src value: {0}, srcType: {1}, destType: {2}", srcVal, srcType, destType);
-            if (destType == typeof(Int16) 
-                || destType == typeof(Int32)
-                || destType == typeof(Int64)
+            Logger.TraceFormat("src value: {0}, srcType: {1}, destType: {2}", srcVal, srcType, destType);
+            
+            if (srcVal == null)
+                return DBNull.Value;
```
howryu: It looks like this line is the actual fix, right? I will add testcase. Thanks for fixing!

Snowflake.Data/Core/SFResultSetMetaData.cs
```diff
@@ -81,31 +81,46 @@ internal int getColumnIndexByName(string targetColumnName)
...
+            var column = rowTypes[targetIndex];
+            var dataType = GetSFDataType(column.type);
+            var nativeType = GetNativeTypeForColumn(dataType, column);
+
+            return Tuple.Create(dataType, nativeType);
+        }
 
+        private SFDataType GetSFDataType(string type)
+        {
+            if (Enum.TryParse(type, true, out SFDataType rslt))
```
howryu: In my env, this line cannot be compiled. I have to rewrite it to 
```
SFDataType rslt;
if (Enum.TryParse(type, true, out rslt))
```

howryu:
Pushed from #18

Snowflake.Data/Core/SFBaseResultSet.cs
```diff
@@ -24,134 +24,33 @@ abstract class SFBaseResultSet
...
-
-            switch(sfDataType)
+            var type = sfResultSetMetaData.getColumnTypeByIndex(columnIndex);
+            switch (type)
             {
                 case SFDataType.DATE:
-                    return SFDataConverter.toDateString(getDateTime(columnIndex), 
+                    var val = GetValue(columnIndex);
+                    if (val == null)
+                        return null;
```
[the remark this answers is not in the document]
akrock: Yes these should have probably DBNull, looks like I missed them.

Issue #4, opened 2018-01-18
Support async query execution


howryu:
https://msdn.microsoft.com/en-us/library/system.data.common.dbcommand.executedbdatareaderasync(v=vs.110).aspx

PR #15, opened 2018-03-02
Fix async support

akrock:
This pull request implements OpenAsync and all ExecuteXXXAsync methods for DbCommand.

It builds upon the changes made by stefanorr in PR #10 .

Issues fixed:
#4 
#9 (this was actually fixed by stefanorr, it was an issue with the value provided for the timeout when downloading chunks).

Although it was not directly reported the base implementation of sending a request in RestRequestImpl.cs would have caused issues with ASP.NET applications due to deadlock caused by using .Result.

All Synchronous calls should now be safe for use in ASP.NET applications, although Asynchronous calls should be preferred when available.

``` 
var response = HttpUtil.getHttpClient().SendAsync(requestMessage, cancellationToken)
                    .Result.EnsureSuccessStatusCode();
```

### DIFF
--- a/Snowflake.Data/Core/SFChunkDownloader.cs
+++ b/Snowflake.Data/Core/SFChunkDownloader.cs
@@ -6,6 +6,7 @@ using System;
 using System.IO.Compression;
 using System.IO;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -23,9 +24,7 @@ namespace Snowflake.Data.Core
         private List<SFResultChunk> chunks;
 
         private string qrmk;
-
-        private int colCount;
-
+        
         private int nextChunkToDownloadIndex;
 
         private int nextChunkToConsumeIndex;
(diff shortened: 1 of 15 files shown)


Snowflake.Data/Core/SFChunkDownloader.cs
```diff
@@ -42,94 +41,70 @@ class SFChunkDownloader
...
+                timeout = TimeSpan.FromHours(1),
+                httpRequestTimeout = TimeSpan.FromSeconds(16),
                 chunkHeaders = downloadContext.chunkHeaders
             };
 
-            HttpResponseMessage httpResponse = restRequest.get(downloadRequest);
+            var httpResponse = await restRequest.GetAsync(downloadRequest);
             Stream stream = httpResponse.Content.ReadAsStreamAsync().Result;
             IEnumerable<string> encoding;
+            //TODO this shouldn't be required.
```
howryu: Can you explain why this should not be required?

Snowflake.Data/Core/SFResultSet.cs
```diff
@@ -2,91 +2,83 @@
...
             sfResultSetMetaData = new SFResultSetMetaData(responseData);
 
             updateSessionStatus(responseData);
             isClosed = false;
         }
 
-        internal override bool next()
+        internal override Task<bool> NextAsync()
+        {
+            return Task.FromResult(Next());
```
howryu: So I am assuming that this is synchronous operation, right? In most cases I guess synchronous Next() is okay. 

However, if the next line is start of a new chunk and download from s3 has not finished yet, it is possible that this consumer thread should be blocked so that waiting download being finished.

Snowflake.Data/Core/SFChunkDownloader.cs
akrock: HttpClient supports Automatic Decompression of responses.

See: https://msdn.microsoft.com/en-us/library/system.net.http.httpclienthandler.automaticdecompression(v=vs.118).aspx

Need to confirm if it is enabled by default of if it needs to be requested. I did not get around to testing that yet.

Snowflake.Data/Core/SFResultSet.cs
akrock: It is possible that this call may be synchronous under those situations. However to make it not synchronous by wrapping Next in a Task.Run() results in so much overhead that it causes performance problems.

We would need to duplicate (or decompose) the code between Next and NextAsync and make sure that only the call to _chunkDownloader.GetNextChunk is wrapped in a task (or made asynchronously)

howryu:
Close this one since all discussion will be moved to #22

PR #22, opened 2018-03-07
Pr 15

### LINKED ISSUE #4: Support async query execution

howryu:
This pr is used to run #15, which tried to fix #4.

I improved following from #15.

- Improved SnowflakeDbDataReader.NextAsync() method. Basically, in most cases synchronized execution of Next is okay. However, if current chunk consumption is done and new chunk has not been downloaded, main thread will blocked on waiting for download. Wrap this waiting download logic in a separate thread so that main is not blocked.

- Honor cancellation token. If a cancellation token is passed from upper layer. Honor those cancel signal. 

@smtakeda @jianshenghuang @akrock Can you review the change?

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -7,6 +7,7 @@ namespace Snowflake.Data.Tests
     using NUnit.Framework;
     using Snowflake.Data.Client;
     using System.Data;
+    using System;
 
     [TestFixture]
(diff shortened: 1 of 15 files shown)


howryu:
I think .net driver is more a programming interface, so application should be responsible for catching ctrl-c signal and hook it into cancellation token.

PR #25, opened 2018-03-16
Port to Net standard

### LINKED ISSUE #7: Release .Net Core 2 Library
It would be amazing if you also port/integrate this .Net connector to .Net Core 2 version

howryu:
In this pr, I make the .net driver compile against .net standard 2.0 and .net framework 4.6 
For .net framework 4.6.1 and after, should be compatible with .net standard 2.0
See the link for detail of .net standard version mapping 
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

This should resolve #7 
@smtakeda  @jianshenghuang  Can you guys review it? Mostly, the change happened in removing common.logging and switched to log4net. This is because common.logging is not supported in .net standard.
cc @hkapre


Snowflake.Data.Tests/App.config
```diff
@@ -3,23 +3,19 @@
 Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
 -->
 <configuration>
-  <configSections>
+  <!--<configSections>
     <sectionGroup name="common">
       <section name="logging" type="Common.Logging.ConfigurationSectionHandler, Common.Logging" />
     </sectionGroup>
     <section name="log4net" type="log4net.Config.Log4NetConfigurationSectionHandler, log4net"/>
+  </configSections> -->
```
smtakeda: If not used, can you remove it?

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -30,8 +30,19 @@ public SFBaseTest()
         [OneTimeSetUp]
         public void SFTestSetup()
         {
-            string testConfigString = Encoding.UTF8.GetString(Properties.Resources.parameters);
+#if NET46
+            log4net.GlobalContext.Properties["framework"] = "net46";
+            log4net.Config.XmlConfigurator.Configure();
 
+#else
```
smtakeda: So this means two test binaries is built? one for .Net framework and other for .Netcore? If so, are you running tests for both binaries?

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -1,106 +1,38 @@
...
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <TargetFrameworkProfile />
+    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
+		<Title>Snowflake.Data</Title>
+    <Description>Snowflake Connector for .NET</Description>
+    <Company>Snowflake Computing, Inc</Company>
+    <Product>Snowflake Connector for .NET</Product>
+    <Copyright>Copyright (c) 2012-2018 Snowflake Computing Inc. All rights reserved.</Copyright>
+		<Version>0.3.0</Version>
```
smtakeda: If this is the only version info, this is fine. But if multiple places have the version info, it would be better to have a separate version file so that all can share the same.

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -1,106 +1,38 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{3E6328DD-E284-4F23-B55A-472BB7EA2C25}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>Snowflake.Data</RootNamespace>
-    <AssemblyName>Snowflake.Data</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <TargetFrameworkProfile />
+    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
```
smtakeda: Ok. this means both .Netcore 2.0 and .Net Framework 4.6 are supported?

Snowflake.Data.Tests/App.config
howryu: Will remove it.

Snowflake.Data.Tests/SFBaseTest.cs
howryu: Yes. Running tests for both.

Snowflake.Data/Snowflake.Data.csproj
howryu: Right. Note this is the project file for .net driver itself. So I used the netstandard2.0, which did not have a runtime. (But should be good for compilation) Customer (or application) needs to use netcoreapp2.0 as target framework since .net core will have a runtime. Note the tests binaries are using .netcoreapp2.0 since ultimately, for tests to be run, a runtime is required.

Snowflake.Data/Snowflake.Data.csproj
howryu: This should be the only place for version. 

The other version should not be needed since that is the project file for tests binaries.

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -1,118 +1,36 @@
...
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <TargetFrameworkProfile />
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
+	  <Title>Snowflake.Data.Tests</Title>
+    <Description>Snowflake Connector for .NET</Description>
+    <Company>Snowflake Computing, Inc</Company>
+    <Product>Snowflake Connector for .NET</Product>
+    <Copyright>Copyright (c) 2012-2018 Snowflake Computing Inc. All rights reserved.</Copyright>
+		<Version>0.3.0</Version>
```
howryu: @smtakeda  Will remove this one since this is the project file for tests.

Snowflake.Data/Snowflake.Data.csproj
smtakeda: ok, by production, add a log including the driver name and version so that the given log file tell what version of driver is used.

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -30,8 +30,19 @@ public SFBaseTest()
         [OneTimeSetUp]
         public void SFTestSetup()
         {
-            string testConfigString = Encoding.UTF8.GetString(Properties.Resources.parameters);
+#if NET46
```
polita: Is there a #define or build target for 4.6.1 specifically, or can we assume everyone with 4.6 has been upgraded to 4.6.1?

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -1,118 +1,36 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{97E793E8-62A4-4FF2-9406-F5277269EEC2}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>Snowflake.Data.Tests</RootNamespace>
-    <AssemblyName>Snowflake.Data.Tests</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <TargetFrameworkProfile />
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
```
polita: Same question on the target framework: do we need to explicitly target 4.6**.1**, or will targeting 4.6 ensure all customers have 4.6.1 at least?

Snowflake.Data/Core/SFSessionProperty.cs
```diff
@@ -65,20 +65,19 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
                             SFSessionProperty p = (SFSessionProperty)Enum.Parse(
```
polita: A couple lines above, we split using the String.Split overload that takes a string array but pass in just a character. We should use the overload that takes a char[]. It will be slightly faster.

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -1,106 +1,38 @@
...
+    <Product>Snowflake Connector for .NET</Product>
+    <Copyright>Copyright (c) 2012-2018 Snowflake Computing Inc. All rights reserved.</Copyright>
+		<Version>0.3.0</Version>
+    <DebugType>Full</DebugType>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
+	
+	<ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
```
polita: Current for Newtonsoft.Json is 11.0.2. Might be worth updating while you're in the code.

appveyor.yml
```diff
@@ -1,7 +1,7 @@
-version: 0.1.0.{build}
-configuration: Release 
+version: '{branch}-{build}'
+configuration: Debug 
 platform: Any CPU
-
+image: Visual Studio 2017
```
polita: Do we need to have the secret in line 7 checked in?

Snowflake.Data.Tests/SFBaseTest.cs
howryu: No tests build against .net 4.6.1 currently. I could add one.

Assumption is that if library a.k.a snowflake .net connector package is build against 4.6, so that every application using 4.6.1 should be able to use our connector right?

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
howryu: Same question from me. If our library against 4.6, so that everyone using 4.6 or 4.6.1 should be able to use our connector? Do I miss anything?

appveyor.yml
howryu: This is not our secret. I encrypt config file (containing sensitive info like password) and commit to github. 
Line 7 is "secure variable" in appveyor for this CI to decrypt our config files. 
Details found in: https://www.appveyor.com/docs/how-to/secure-files/

Snowflake.Data.Tests/SFBaseTest.cs
polita: I meant this as a general question about 4.6 versus 4.6.1. If memory serves, we (on the Azure team) used to build explicitly for 4.6.1 or greater, rather than for 4.6.

PR #31, opened 2018-05-11
Logging improve

howryu:
1. How to enable logging? See last section of READMDE.md

2. Does the logging have os version, framework version etc? Yes and No.
Client environment is added to authentication request body, and logged. However, it is really difficult to get framework version at runtime. MS has a ticket for it. https://github.com/dotnet/corefx/issues/17452

3. Sample logging. https://ci.appveyor.com/project/howryu/snowflake-connector-net/build/logging_improve-135/artifacts

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbCommand.cs
+++ b/Snowflake.Data/Client/SnowflakeDbCommand.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Snowflake.Data.Log;
 
 namespace Snowflake.Data.Client
 {
@@ -22,8 +23,11 @@ namespace Snowflake.Data.Client
 
         private SnowflakeDbParameterCollection parameterCollection;
 
+        private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
+
         public SnowflakeDbCommand(SnowflakeDbConnection connection)
         {
+            logger.Debug("Constucting SnowflakeDbCommand class");
             this.connection = connection;
             this.sfStatement = new SFStatement(connection.SfSession);
             // by default, no query timeout
@@ -124,6 +128,7 @@ namespace Snowflake.Data.Client
         
         public override int ExecuteNonQuery()
         {
+            logger.Debug($"ExecuteNonQuery, command: {CommandText}");
             SFBaseResultSet resultSet = ExecuteInternal();
             resultSet.Next();
             return resultSet.CalculateUpdateCount();
@@ -131,6 +136,7 @@ namespace Snowflake.Data.Client
 
         public override async Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
         {
+            logger.Debug($"ExecuteNonQueryAsync, command: {CommandText}");
             if (cancellationToken.IsCancellationRequested)
                 throw new TaskCanceledException();
 
@@ -141,6 +147,7 @@ namespace Snowflake.Data.Client
 
         public override object ExecuteScalar()
         {
+            logger.Debug($"ExecuteScalar, command: {CommandText}");
             SFBaseResultSet resultSet = ExecuteInternal();
             resultSet.Next();
             return resultSet.GetValue(0);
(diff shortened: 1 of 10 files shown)


Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -124,13 +128,15 @@ public override void Cancel()
         
         public override int ExecuteNonQuery()
         {
+            logger.Debug($"ExecuteNonQuery, command: {CommandText}");
```
smtakeda: can you explain what `$` prefix mean?

Snowflake.Data/Core/IRestRequest.cs
```diff
@@ -114,6 +114,15 @@ class AuthnRequestClientEnv
 
         [JsonProperty(PropertyName = "OS_VERSION")]
         internal String osVersion { get; set; }
+
+        [JsonProperty(PropertyName = "NET_RUNTIME")]
```
smtakeda: Is this a new property?

smtakeda: on top OS_VERSION, can you add `OS` to `Windows`? That's how ODBC set.

Snowflake.Data/Core/IRestRequest.cs
```diff
@@ -114,6 +114,15 @@ class AuthnRequestClientEnv
 
         [JsonProperty(PropertyName = "OS_VERSION")]
```
smtakeda: What is the OS version look like?

Snowflake.Data/Client/SnowflakeDbCommand.cs
smtakeda: looks like format string?

howryu: It is just .net syntax. Prefix `$` means content in curly bracket will be replaced by actual variable `CommandText`

Snowflake.Data/Core/IRestRequest.cs
howryu: This is how OS version looks like `ClientEnv: { APPLICATION: dotnet, OS_VERSION: Microsoft Windows NT 6.2.9200.0, NET_RUNTIME: .NET Core 4.6.26212.01 }`

Snowflake.Data/Core/SFSession.cs
```diff
@@ -72,7 +74,12 @@ static SFSession()
             AuthnRequestClientEnv clientEnv = new AuthnRequestClientEnv()
             {
                 application = System.Diagnostics.Process.GetCurrentProcess().ProcessName,
-                osVersion = System.Environment.OSVersion.VersionString
+                osVersion = System.Environment.OSVersion.VersionString,
+#if NET46
+                netRuntime = "CLR:" + Environment.Version.ToString()
+#else
+                netRuntime = System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription
+#endif
```
smtakeda: what is the value `netRuntime` look like? Can you give examples? Can it be combined with `OS_VERSION`?

In case of ODBC driver, OS_VERSION consists of the windows version followed by architecture.
```
"6.1-x86_64"
```

howryu: The initial idea of NET_RUNTIME is to distinguish .net core or .net framework. However, like I said, it is really difficult to get .net version in runtime. dotnet/corefx#17452

Snowflake.Data/Core/IRestRequest.cs
howryu: This is how JDBC client environment build:
```
      Map<String, Object> clientEnv = new HashMap<String, Object>();
      clientEnv.put("OS", System.getProperty("os.name"));
      clientEnv.put("OS_VERSION", System.getProperty("os.version"));
      clientEnv.put("JAVA_VERSION", System.getProperty("java.version"));
      clientEnv.put("JAVA_RUNTIME", System.getProperty("java.runtime.name"));
      clientEnv.put("JAVA_VM", System.getProperty("java.vm.name"));
```
JDBC also have some java specific attribute.

Snowflake.Data/Core/SFSession.cs
smtakeda: Ah I remember that. So we are providing two binaries to nuget.

howryu: Yeah, two binaries in one package.

PR #32, opened 2018-05-13
fix #19 by checking for `rows_loaded` column return in COPY statements

### LINKED ISSUE #19: ExecuteNonQuery() generate exception if COPY statement failed
Calling ExecuteNonQuery() on a COPY statement can generate an exception if the COPY statement fails and snowflake returns a single column instead of 3 that the code in ResultSetUtil.cs is expecting.

manigandham:


manigandham:
CSV of result set in both conditions:

When rows are loaded:
```
file,status,rows_parsed,rows_loaded,error_limit,errors_seen,first_error,first_error_line,first_error_character,first_error_column_name
s3://somefile.gz,LOADED,123,123,1,0,,,,
```

When nothing is loaded:
```
status
Copy executed with 0 files processed.
```

howryu:
@manigandham Thanks for your contribution. I will add a tests, and merge it afterwards.

howryu:
@manigandham I added tests in #36 for your change and it looks good.

PR #57, opened 2018-08-31
Implemented #8 Can you guys please add GetSchemaTable() into the library

abelykh0:


abelykh0:
Can you please consider merging this or exposing something non-standard (for example, see NpgsqlDataReader.GetColumnSchema) that would let me know what data types are in the result. I am not sure why checks have failed, the change seems to be harmless.

howryu:
Thanks I will review it today.

Snowflake.Data/Client/SnowflakeDbDataReader.cs
```diff
@@ -81,6 +81,49 @@ public override bool IsClosed
 
         public override int RecordsAffected => resultSet.CalculateUpdateCount();
 
+        public override DataTable GetSchemaTable()
+        {
+            if (this.FieldCount == 0)
+            {
+                // No result
+                return null;
+            }
```
howryu: We should probably return an empty dataTable instead of null. Although I don't think there is a possibility that no column is returned.

howryu:
I will add tests to it.

Snowflake.Data/Client/SnowflakeDbDataReader.cs
abelykh0: That's fine with me.

howryu:
Merged in #58

Issue #65, opened 2018-10-03
System.IndexOutOfRangeException on ExecuteScalar(Async) if query does not return a result


akrock:
### Issue description
If the query does not return a result ExecuteScalar and ExecuteScalarAsync will throw an IndexOutOfRangeException (I think this gets swallowed and returned as a generic error IIRC). 

The expected behavior is that ExecuteScalar would return DBNull in this case.

### Example code
```cs
var cmd = new SnowflakeDbCommand(_conn) {CommandText = "SELECT * FROM <TABLE> WHERE 1 !=1"};
var result = cmd.ExecuteScalar();
```
### Configuration
*Driver version:* Master, any framework.
*Server version:* 2.59.0

PR #66, opened 2018-10-03
PR Fix for #65

akrock:
Fix for issue #65

Check if Next/NextAsync has a positive return before attempting to call GetValue(0).
Return DBNull.Value if it does not.


howryu:
Merged in #117

Issue #67, opened 2018-10-03
Transient Errors during Http requests are not retried


akrock:
### Issue description
Potentially transient errors in the HttpUtil are not retried. I would expect that transient errors for the http connections would be retried automatically by the client.

### Configuration
*Driver version:* Master, any framework
*Server version:* 2.59.0

PR #68, opened 2018-10-03
Transient Errors during Http requests are not retried

akrock:
This came from our fork of the connector. I cannot actually remember now what the specific error was that was transient and retry-able. 

Some exploration would probably be good as to what errors could be transient. A max retries in this section also might be beneficial.


akrock:
Fix for #67

howryu:
Thanks seems reasonable for me. So this is the case that no http response at all right? So potentially something wrong in tcp layer or network, right?

akrock:
Yes, this would handle if something critical happened during SendAsync that produced an exception instead of an HTTP Response.

Issue #69, opened 2018-10-03
Statements mutating more than int.MaxValue rows cause a OverflowException


akrock:
### Issue description
If an insert or update statement causes more than int.MaxValue rows to be modified an OverflowException is thrown.

I would expect that the snowflake connector is able to handle this case w/o errors.

### Configuration
*Driver version:* Master, any framework
*Server version:* 2.59.0

PR #70, opened 2018-10-03
PR - Fix for #69

akrock:
I searched but I was not able to figure out the recommended driver behavior for this case.

I opted to return -1 if more than int.MaxValue rows were mutated and an OverflowException was generated. This may be non-standard behavior and you should probably confirm what the expected behavior for a .NET driver is in this case.

### DIFF
--- a/Snowflake.Data/Core/ResultSetUtil.cs
+++ b/Snowflake.Data/Core/ResultSetUtil.cs
@@ -17,32 +17,57 @@ namespace Snowflake.Data.Core
             SFResultSetMetaData metaData = resultSet.sfResultSetMetaData;
             SFStatementType statementType = metaData.statementType;
 
-            int updateCount = 0;
-            switch (statementType)
+            try
             {
-                case SFStatementType.INSERT:
-                case SFStatementType.UPDATE:
-                case SFStatementType.DELETE:
-                case SFStatementType.MERGE:
-                case SFStatementType.MULTI_INSERT:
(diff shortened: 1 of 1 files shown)


akrock:
This PR addresses #69

akrock:
Yeah you could also just request `GetValue<long>`

However you will still need to figure out what should be returned by the method in the case that it exceeds int.MaxValue since you are forced by the interface to return an int for `ExecuteNonQuery`

howryu:
Seems that should be the way. This actually happens after update/delete/insert succeed in the server side. Exception is not a good idea, which will lead to user thinking that their insertion failed. But I was wondering if there is a better to inform user. Like in JDBC, there is a class call SqlWarning. https://docs.oracle.com/javase/7/docs/api/java/sql/SQLWarning.html  I did not find anything similar in .net

akrock:
Rewrote this PR to use GetValue<long> instead of handling the overflow exceptions from GetValue<int>, still returns -1 in the case that more than int.MaxValue rows were modified.

ikudjoi:
Hi, any comment on this @howryu?

howryu:
@ikudjoi ThanksGiving now. I will fix it next Monday since I don't have access to my dev machine now. Sorry for that.

ikudjoi:
Ping

howryu:
Working on it now.

howryu:
Merged in #95

PR #71, opened 2018-10-03
Potential Fix for large result sets

akrock:
In looking at logs & the existing ChunkDownloader code I noticed that the BlockingCollection wasn't having the expected results.

Basically we were going full throttle on all the requests at once to S3 (which would then of course be limited by the number of concurrent http requests allowed from the runtime). I changed the implementation to use Lazy in the BlockingCollection and now spin up threads that will execute the DownloadChunkAsync method.

In my testing on a 8.2M result set ~80 chunks, it seems to save somewhere in the range of 5-10 seconds off the time to process the full result set. I also bumped the number of slots from 2 to 4.

In reality the longest portion of this time is simply waiting for S3 to respond w/ the data and download it. The JSON deserialization is not much of a factor and seems to consume less than 100ms per chunk.

### DIFF
--- a/Snowflake.Data/Core/SFChunkDownloader.cs
+++ b/Snowflake.Data/Core/SFChunkDownloader.cs
@@ -15,6 +15,7 @@ using System.Threading.Tasks;
 using System.Net.Http;
 using Newtonsoft.Json;
 using System.Diagnostics;
+using Newtonsoft.Json.Linq;
 using Newtonsoft.Json.Serialization;
 using Snowflake.Data.Log;
 
@@ -46,7 +47,8 @@ namespace Snowflake.Data.Core
             this.qrmk = qrmk;
             this.chunkHeaders = chunkHeaders;
             this.chunks = new List<SFResultChunk>();
-        
+            externalCancellationToken = cancellationToken;
+
             var idx = 0;
             foreach(ExecResponseChunk chunkInfo in chunkInfos)
             {
@@ -91,25 +93,27 @@ namespace Snowflake.Data.Core
                     chunkIndex = c.ChunkIndex,
                     qrmk = this.qrmk,
                     chunkHeaders = this.chunkHeaders,
-                    cancellationToken = this.externalCancellationToken
+                    cancellationToken = this.externalCancellationToken,
                 })));
             }
 
             _downloadTasks.CompleteAdding();
             _downloadQueue = new ConcurrentQueue<Lazy<Task<SFResultChunk>>>(_downloadTasks);
 
-            for (var i = 0; i < prefetchSlot; i++)
-            {
-                Thread t = new Thread(RunDownloads) { IsBackground = true};
-                t.Start();
-            }
+            for (var i = 0; i < prefetchSlot && i < chunks.Count; i++)
+                Task.Run(new Action(RunDownloads));
         }
         
         public Task<SFResultChunk> GetNextChunkAsync()
         {
+            if(_downloadTasks.IsCompleted)
+                logger.Info($"Total Parse Time: {TotalParseTime} ms");
+
             return _downloadTasks.IsCompleted ? Task.FromResult<SFResultChunk>(null) : _downloadTasks.Take().Value;
         }
-        
+
+        private long TotalParseTime = 0;
+
         private async Task<SFResultChunk> DownloadChunkAsync(DownloadContext downloadContext)
         {
             logger.Info($"Start downloading chunk #{downloadContext.chunkIndex+1}");
@@ -127,22 +131,26 @@ namespace Snowflake.Data.Core
                 chunkHeaders = downloadContext.chunkHeaders
             };
 
-            var httpResponse = await restRequest.GetAsync(downloadRequest, downloadContext.cancellationToken);
-            Stream stream = httpResponse.Content.ReadAsStreamAsync().Result;
-            IEnumerable<string> encoding;
-            //TODO this shouldn't be required.
-            if (httpResponse.Content.Headers.TryGetValues("Content-Encoding", out encoding))
+            var httpResponse = await restRequest.GetAsync(downloadRequest, downloadContext.cancellationToken).ConfigureAwait(false);
+            Stream stream = await httpResponse.Content.ReadAsStreamAsync().ConfigureAwait(false);
+
+            if (httpResponse.Content.Headers.TryGetValues("Content-Encoding", out var encoding))
             {
-                if (String.Compare(encoding.First(), "gzip", true) == 0)
+                if (string.Equals(encoding.First(), "gzip", StringComparison.OrdinalIgnoreCase))
                 {
                     stream = new GZipStream(stream, CompressionMode.Decompress);
                 }
             }
 
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
             parseStreamIntoChunk(stream, chunk);
+            sw.Stop();
+
+            Interlocked.Add(ref TotalParseTime, sw.ElapsedMilliseconds);
 
             chunk.downloadState = DownloadState.SUCCESS;
-            logger.Info($"Succeed downloading chunk #{downloadContext.chunkIndex+1}");
+            logger.Info($"Succeed downloading chunk #{downloadContext.chunkIndex+1} - {sw.ElapsedMilliseconds} parse - {chunk.rowCount}");
 
             return chunk;
         }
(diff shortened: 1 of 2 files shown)


manigandham:
Is it necessary to create new threads? That's very inefficient compared to using the Tasks, especially when they're such short-lived requests. We have many reporting queries running and this would likely cause scaling issues, especially with queries retrieving cached resultsets.

akrock:
I tested both Task.Run and Thread start for the `RunDownloads` method, the threads seemed to give a little more reliable performance, but changing these to tasks would still probably be fine. I did not test this at scale w/ concurrent query executions.

The meat of this change that is important is that we actually throttle and control the order of the requests that are made to S3 instead of firing everything at once.

manigandham:
Please use Tasks. Creating a new thread will exhaust the threadpool and they do not die quickly. It's very resource intensive to run a simple network-IO task like downloading over HTTP.

If you dont mind a new dependency, the [TPL Dataflow blocks](https://www.nuget.org/packages/Microsoft.Tpl.Dataflow/) make it easy to have a processing queue with a preset concurrency limit.

akrock:
Understood, I have switched to tasks over threads and have some testing of different approaches going on now. I should have an update today with the results.

akrock:
Just to update, I did switch it back to Tasks, and settled on 5 pre-fetch slots based on some local testing.

Additionally I switched the parsing of the result chunks to directly use the JsonTextReader instead of deserialize, which may see some speed improvements. It also allow us to pre-allocate the result matrix while the request downloads. 

For the tests I was downloading a 8.2M x 13 column table on a 200Mbps WIFI connection. Each row represents 20 timed runs. Baseline = the current master code.

|Test|AVG|MIN|MAX|
|--|--|--|--|
|baseline|44.3788|41.5140|48.9480|
|baseline|44.6101|41.8070|48.4670|
|baseline|44.7609|43.2060|47.5800|
|3 slots|31.2257|29.4190|33.8440|
|3 slots|31.4876|29.7100|35.5090|
|3 slots|31.8356|30.1010|33.4790|
|4 slots|27.0725|25.1240|29.5770|
|4 slots|27.2180|25.1190|29.8610|
|4 slots|27.9197|24.9170|38.2300|
|5 slots|25.2488|23.5080|28.8940|
|5 slots|24.5056|23.3100|25.6840|
|5 slots|25.6131|23.5500|27.9700|
|6 slots|24.1368|22.3310|28.9590|
|6 slots|24.6847|22.5050|37.3020|
|6 slots|23.8734|22.5290|27.4460|
|7 slots|23.4111|21.5380|25.5610|
|7 slots|22.9591|21.2560|24.7370|
|7 slots|24.3315|21.9390|29.1790|
|8 slots|22.4175|20.8210|27.5360|
|8 slots|22.8974|21.4960|26.5370|
|8 slots|22.1182|20.7590|25.4210|

howryu:
Thanks so much for your hardwork. I will review it at my earliest possible. My hands are tied to some other projects. 
Also, I am not surprised to see the improvement from switching to JsonTextReader. So correct me if I am wrong, JsonTextReader is like a streaming way of reading results, instead of allocating the whole chunk of data's memory, is that correct? In other drivers, I saw the improvement by switching from JsonMapper to a reading content token by token.

akrock:
https://www.newtonsoft.com/json/help/html/Performance.htm

> The absolute fastest way to read and write JSON is to use JsonTextReader/JsonTextWriter directly to manually serialize types. Using a reader or writer directly skips any of the overhead from a serializer, such as reflection.

The code was already using a Stream to deserialize the result chunk so it wasn't trying to process all the data into a string first. The change to read the Tokens directly from JsonTextReader does bypass some reflection overhead. Also since the read is forward-only there is no possible way that the `Deserialize` method could have allocated the full two dimension array up front, like we can now.

Snowflake.Data/Core/SFChunkDownloader.cs
```diff
@@ -60,38 +57,63 @@ class SFChunkDownloader
...
+
+                _downloadTasks.Add(t);
+            }
+
+            _downloadTasks.CompleteAdding();
+
+            _downloadQueue = new ConcurrentQueue<Lazy<Task<SFResultChunk>>>(_downloadTasks);
+
+            for (var i = 0; i < prefetchSlot && i < chunks.Count; i++)
+                Task.Run(new Action(RunDownloads));
```
howryu: It seems that prefetchSlot is used only when chunk downloader starts. However, I still think it make sense to prefetch when a chunk is taken from the BlockingCollections. That's why we use prefetchSlot for the upper-bound of blocking collections. i.e. always keeps 2 slots for prefetching.

Snowflake.Data/Core/SFChunkDownloader.cs
```diff
@@ -60,38 +57,63 @@ class SFChunkDownloader
...
+        }
+
+
+        private void FillDownloads()
+        {
+            _downloadTasks = new BlockingCollection<Lazy<Task<SFResultChunk>>>();
+
+            foreach (var c in chunks)
+            {
+                var t = new Lazy<Task<SFResultChunk>>(() => DownloadChunkAsync(new DownloadContext()
```
howryu: Does `Lazy` means that Task will be executed only when adding task to collections>?

howryu:
I will add an interface for chunk downloader and json text parser so that it is easier to add different implementation

howryu:
@akrock I added an interface to IChunkDownloader. I think we need some more discussion before merging your change, but I added an interface and we probably want a gradual rollout of your chunk downloader implementation. (I moved your version of chunk downloader to a new class called `SFChunkDownloaderV2`)

Snowflake.Data/Core/SFChunkDownloader.cs
akrock: The prefetch slot is used to spawn tasks that will persist until all of the data is downloaded.

Snowflake.Data/Core/SFChunkDownloader.cs
akrock: Lazy means that the Task won't be executed until the `Value` property is called.

Snowflake.Data/Core/SFChunkDownloader.cs
howryu: So what's the point in your code using BlockingCollection? It's not blocking anything. All `Lazy<Task<SFResultChunk>>` element will be added into BlockingCollection at the beginning of the chunk downloader right? Then data will only be downloaded when main thread has finished consuming previous chunk and call `GetNextChunkAsync`

howryu:
@akrock Can you elaborate what's wrong with current blockingCollections implementation? It seems that it behaves as expected. 

Our expectation for chunk downloader being as, foreground thread will consume and background thread will download. And prefetchSlot is used to control how many slots can exist in memory. Let's say we have 6 total chunks to download, and prefetchSlot is 2.  When `SFChunkDownloader` is initialized, background thread will download first 2 chunk data. And if main thread finished consuming the first chunk, background thread will go on downloading 3rd chunk. And it seems to me current code is doing what we expects.

akrock:
@howryu You are correct actually, I'm not sure where I got it in my head that we were letting them all go at once. However the new implementation does have some advantages still.

Since the "chunks" provided by snowflake are in no way uniform the current implementation can end up being blocked by a single download.

Example with 3 prefetch slots:
1 - Small (fast)
2 - Large (slow)
3 - Small (fast)
4 - Small (fast)
5 - Small (fast)
6 - Large (slow)
7 - Small (fast)

Assume small chunks complete in 3 seconds, while large chunks may take up to 45 seconds.

In our current code, 1, 2, and 3 will start download immediately (actually 4 will as well, but were going to ignore that for this argument).

Chunk 1 will be popped off as the user iterates the response, and 4 will begin downloading. Chunk 2 is then popped off however it has not finished downloading, so the user will not be able to immediately consume it. Chunk 5 starts because chunk 2 is popped off of the blocking collection. 

At this point, chunks 3,4,5 have completed download but we will not start the next slow chunk because the user has not been able to process chunk 2 yet.

Contrast that with the new implementation which will always ensure that the number of prefetch slot size downloads are **active** until the entire result set has been downloaded. With this implementation as soon as each chunk has downloaded, the next chunk will start. Therefore even though a large download is preventing the user from actively iterating chunk 2, the next large chunk will start as soon as an open slot is available. Overall this leads to a reduction of the wall-time required to retrieve the full result, as we are maximizing our parallel downloads.

Snowflake.Data/Core/SFChunkDownloader.cs
akrock: I kept the BlockingCollection for the `IsCompleted` behavior, we could use a simple Queue, but we would then need to implement our own checks to prevent a race condition where the consumer hits an empty Queue before it has been populated and assume that it has completed.

howryu:
@akrock  I am not convinced by your point. 

1. snowflake backend will ensure each chunk is around same size, although not super accurate.
2. In your example, you can always increase pretchSlot 4. The point here is that we don't want to let background thread go too faster than main consumer thread since potentially it can blow up application's memory. We already see such scenario in other driver (other driver use same strategy) that driver is only small part of the whole application but consume too much memory. 
3. I would like to expose a session parameter to allow use to control prefetchSlot, if you don't care memory usage, you can increase this number.

However, I still would like to thank you for your contribution to the stream parser. I think it make sense to use the JsonTextReader to parse token by token.

akrock:
@howryu I'm surprised to hear that the backend is attempting to make them uniform because from what I see they vary wildly.

In my test case that is 80 chunks I see the following:

MIN ROWS: 768
MIN SIZE: 27,045

MAX ROWS: 450,560
MAX SIZE: 16,874,671

I don't feel that simply allowing the adjustment of the prefetch slots is a 100% solution, is there any way we can allow the user to choose between a memory optimized approach vs a download speed approach as well?

howryu:
@akrock We actually have another parameter should be able to expose (currently not exposed), called client_memory_limit, which is a cap of maximum of memory that can be used for downloading chunks. However, this is not implemented yet. Basic idea is that user can increase prefetchSlot, however, if the memory usage hit the limit, then block the background thread from continuing downloading new chunks.

akrock:
I can work on integrating the fully parallel approach w/ the client_memory_limit concept if that will solve the concern of it possibly consuming too much memory if the result sets are not consumed as fast as they are downloaded.

howryu:
Thank you! Let me clean up this, and adding some code to deal with different config/sessionParameters to control different behavior. Also, I need to sync up with my team mate to have a consistent behavior with other drivers.

howryu:
I added #91 to allow use old/new chunk downloader.

PR #73, opened 2018-10-03
Set ConnectionState.Closed on Close()

### LINKED ISSUE #72: Close() does not update ConnectionState
### Issue description
Calling Close() on the connection does not actually update the connection state.

Expected: State to be ConnectionState.Closed after calling Close(). Currently ConnectionState.Open.

akrock:
Set  _connectionState = ConnectionState.Closed on Close()

Fixes #72


akrock:
@ikudjoi 
_connectionState will currently be Closed until SFSession is not null, so really the only time that SFSession would be null during a Close() call is if the connection was never opened. So I don't really think it matters... remaining inside the if block merely prevents a redundant assignment to a state that it already would be in.

howryu:
Merged in #88

Issue #81, opened 2018-10-17
Serialization & deserialization of numbers fails on clients using Culture with non-standard number formatting


ikudjoi:
### Issue description

1. Set your region to e.g. Finland. (On windows region & language settings). Comma will replace the dot as decimal separator.
2. Execute test SFDbDataReaderIT.testGetFloat.
3. Executing the test fails. Writing and reading decimal numbers should be possible on any locale.

### Example code
Provided already in the test.

### Error log
```
Failed   testGetFloat
Error Message:
 Snowflake.Data.Client.SnowflakeDbException : SQL compilation error:
Invalid bind value (1,23) for type (REAL).
Stack Trace:
   at Snowflake.Data.Core.SFStatement.BuildResultSet(QueryExecResponse response, CancellationToken cancellationToken) in C:\work\snowflake-connector-net\Snowflake.Data\Core\SFStatement.cs:line 121
   at Snowflake.Data.Core.SFStatement.Execute(Int32 timeout, String sql, Dictionary`2 bindings, Boolean describeOnly) in C:\work\snowflake-connector-net\Snowflake.Data\Core\SFStatement.cs:line 233
   at Snowflake.Data.Client.SnowflakeDbCommand.ExecuteInternal(Boolean describeOnly) in C:\work\snowflake-connector-net\Snowflake.Data\Client\SnowflakeDbCommand.cs:line 256
   at Snowflake.Data.Client.SnowflakeDbCommand.ExecuteNonQuery() in C:\work\snowflake-connector-net\Snowflake.Data\Client\SnowflakeDbCommand.cs:line 132
   at Snowflake.Data.Tests.SFDbDataReaderIT.testGetFloat() in C:\work\snowflake-connector-net\Snowflake.Data.Tests\SFDbDataReaderIT.cs:line 118
```

### Configuration
1.0.4

*Dotnet framework and version:*
.net standard 2.0

*Server version:*
3.0.5

*Client OS:*
Windows 10

howryu:
Merged in #87

marwesfur:
@howryu  In bf2bb58244b120d6a65c6fdd2160767d2f7e0cc5 setting the `Thread.CurrentThread.CurrentCulture` was moved to the  `SFDataConverter` ctor. Now, it doesn't work in a multi threaded environment anymore, where it might happen that instantiation of `SFDataConverter` and actual conversion take place in different threads. 

I had to switch to the non-async methods to make it work.

[...]

PR #82, opened 2018-10-17
Fix serialization and deserialization of numbers

ikudjoi:
Fix for #81.

Formatting with object.ToString() without specifying culture might be a bad idea anyway so I added it everywhere where applicable.


howryu:
Merged in #87

Issue #83, opened 2018-10-17
Add basic DbProviderFactory implementation to the project


ikudjoi:
### Issue description
To add support for Snowflake to Fluent Migrator project (https://github.com/fluentmigrator/fluentmigrator) the connector must implement _System.Data.Common.DbProviderFactory_, whose purpose is

> Represents a set of methods for creating instances of a provider's implementation of the data source classes.

In my experience very basic implementation is sufficient in the first place to allow using Fluent Migrator for Snowflake database migrations.

howryu:
Merged in #89

PR #84, opened 2018-10-17
#83 Add DbProviderFactory with inflicted changes

ikudjoi:
Add DbProviderFactory implementation SnowflakeDbFactory.
Had to add parameterless constructors to existing classes.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbCommand.cs
+++ b/Snowflake.Data/Client/SnowflakeDbCommand.cs
@@ -25,6 +25,14 @@ namespace Snowflake.Data.Client
 
         private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
 
+        public SnowflakeDbCommand()
+        {
+            logger.Debug("Constucting SnowflakeDbCommand class");
+            // by default, no query timeout
+            this.CommandTimeout = 0;
+            parameterCollection = new SnowflakeDbParameterCollection();
+        }
+
         public SnowflakeDbCommand(SnowflakeDbConnection connection)
         {
             logger.Debug("Constucting SnowflakeDbCommand class");
@@ -79,10 +87,7 @@ namespace Snowflake.Data.Client
 
         public override UpdateRowSource UpdatedRowSource
         {
-            get
-            {
-                return UpdateRowSource.None;
-            }
+            get => UpdateRowSource.None;
 
             set
             {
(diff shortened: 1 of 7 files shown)


ikudjoi:
I added implementation of SnowflakeDbCommandBuilder, although it seems not to be important for FluentMigrator. We urgently need these changes for this PR in FluentMigrator project.

https://github.com/fluentmigrator/fluentmigrator/pull/944

howryu:
I will try to merge it today. Thanks for your contribution!

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -25,6 +25,14 @@ public class SnowflakeDbCommand : DbCommand
 
         private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
 
+        public SnowflakeDbCommand()
```
howryu: So the reason why you add parameterless constructor is because .net driver allows application to set Connection object after DbCommand is created, right?

ikudjoi: Yes, that's right.

ikudjoi: Specially in the added Factory class this is needed. It's late already here, I won't probably answer further questions today.

howryu:
@ikudjoi  Thanks. I am doing some testing right now. It seems to me that I have to add this line to SnowflakeDbFactory.cs to make it work.
```
       public static readonly SnowflakeDbFactory Instance = new SnowflakeDbFactory();
```
Otherwise, it will complain
```
Error Message:
 System.InvalidOperationException : The requested .Net Framework Data Provider's implementation does not have an Instance field of a System.Data.Common.DbProviderFactory derived type.
Stack Trace:
   at System.Data.Common.DbProviderFactories.GetFactory(DataRow providerRow)
   at System.Data.Common.DbProviderFactories.GetFactory(String providerInvariantName)
   at Snowflake.Data.Tests.SFDbFactoryIT.TestSimpleDbFactory() in c:\Snowflake\git\snowflake-connector-net\snowflake-connector-net\Snowflake.Data.Tests\SFDbFactoryIT.cs:line 24
```
So I have to merge it tomorrow. Sorry for that.

ikudjoi:
Hi, sorry, I didn't write that test you are referring to but that change sounds cool to me. This is how I use the factory in FluentMigrator project
https://github.com/fluentmigrator/fluentmigrator/blob/238061cf4fef4e46cf5a08392dcecf34b960e21f/src/FluentMigrator.Runner.Snowflake/Processors/Snowflake/SnowflakeDbFactory.cs

So only full assembly name of the factory is given. The factory specifies the different classes of the Snowflake connector.

Btw, is there a reason that the command connection is immutable? I just tried to change as little as possible and made it throw an exception if the connection is changed.

howryu:
I guess command's connection can be changed. However, I am from JDBC. So in JDBC there is no way to change `Statement`'s `Connection` object. So I just mimic this behavior here.

howryu:
Also, a quick note. Our current SnowflakeDbAdaptor did not support delete or update or insert. Only select statement is supported. Not sure if this will influence your project.

howryu:
Merged in #89

ikudjoi:
Thanks for accepting the PR! You are going to release new NuGet package version sometime soon, right?
FluentMigrator is mostly about DDL statements, the integration tests were working just fine with my own builds of the connector.

howryu:
@ikudjoi I will release it today.

howryu:
FYI, 1.0.5 is out. https://www.nuget.org/packages/Snowflake.Data/

ikudjoi:
Cool, thanks!

PR #91, opened 2018-11-15
V2downloader

howryu:


howryu:
@akrock I added a switch in SFConfiguration class to allow user to specify which parser/chunkdownloader to use. For now, I am enabling the new json parser by default. For your chunk downloader I disabled it since it could be consuming lots of customer's memory.

PR #104, opened 2019-02-05
Fir fox #106

ikudjoi:
There was a rather severe bug in SFDataConverter method ExtractTimestamp. It takes a string as input which is actually a decimal number where integer part is seconds since Unix epoch and decimal part is the fraction part as-is. However, the method assumes that the fraction part is always nine decimals long, but in our setup it depends on the precision of the underlying column. If the decimal part is e.g. 123 with precision 3 (milliseconds) it gets converted into 123 nanoseconds, i.e. 0 milliseconds. It should get converted into 123000000 nanoseconds instead.

This pull requests provides a fix for the method.

While improving tests related to the fix I found that the connector, Snowflake server (3.12.0) and/or snowsql Windows client (v1.1.73) has trouble handling date(time)s that represent time before the unix epoch. Or dates closing the "magic" year 9999. I added test cases for them as well but commented them out as I was not able to provide solutions for them. I tried to find out how to file a snowsql bug could not find a better way than this:
https://snowflakecommunity.force.com/s/question/0D50Z00008qFUXwSAO/

Although my fix does not solve all problems related to date and time parsing I hope at least parsing the post unix epoch -> present dates with lesser accuracy are parsed right, pretty please fix it asap.

### DIFF
--- a/Snowflake.Data.Tests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/SFDataConverterTest.cs
@@ -2,6 +2,9 @@
  * Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
  */
 
+using System;
+using System.Data;
+
 namespace Snowflake.Data.Tests
 {
     using Snowflake.Data.Core;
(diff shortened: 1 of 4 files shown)


ikudjoi:
Hi, IMO this is a bug that should be fixed asap.

howryu:
I will merge it by this week.

howryu:
Running tests in #110

howryu:
Merged in #110

Issue #111, opened 2019-03-05
Inserting dates close to DateTime.Max (9999-12-31 23:59:59.9999999) to database and reading them does not work


ikudjoi:
### Issue description
Commented test cases in SFDbDataReaderIT show that there are problems related to "large" DateTime values. It seems that both inserting to database and reading from it fails because the Snowflake way of representing the time as nanoseconds from the unix epoch overflows the long max value when the represented time is close to DateTime.MaxValue. My quick study suggest that the problems start with dates after 2262-04-11 23:47:16.8547758.

### Example code
Please see e.g. SFDbDataReaderIT.TestConvertDatetime test cases that are commented out.

### Configuration
*Driver version:* git commit a2e27ec657650f7f747367d0b4f583e7dc7ab3a6 (newer than 1.0.6)

*Dotnet framework and version:* .net standard 2.0

*Server version:* 3.15.2

*Client OS:* Windows 10

Issue #103, opened 2019-02-04
DateTime is parsed wrong when querying from time or timestamp column with precision less than nanoseconds (9)


ikudjoi:
### Issue description
When querying DateTime or DateTimeOffset from Snowflake database columns, whose type is time or timestamp with precision less than 9, the result DateTime is parsed wrong, only seconds are parsed right. 

### Example code
I will push pull request tomorrow European time.

### Configuration
*Driver version:* 1.0.6
*Dotnet framework and version:* .net standard 2.0
*Server version:* E.g. 3.12.0
*Client OS:* Windows 10

ikudjoi:
You could close this after accepting my PR but I'm about to file another date parsing related bug in a moment.

howryu:
Close the issue since I have merged it in #110

PR #112, opened 2019-03-05
#111 Fix handling large DateTime values

ikudjoi:
When providing fix for #103 in PR #104 (also #110 related because of merge conflicts) I found that there are also other problems related to handling DateTime values. I was kind of hoping that you'd dig in to them and fix them but you didn't so far so here it goes.

I just filed in a separate issue #111 of problems related to large DateTime values and I am providing fix for the issue in this PR. Without these fixes both in inserting to database and reading from it will result in overflowing long value. These changes help avoiding them.

Also removed a test I earlier commented out after replacing it with another one.

### DIFF
--- a/Snowflake.Data/Core/SFDataConverter.cs
+++ b/Snowflake.Data/Core/SFDataConverter.cs
@@ -94,7 +94,8 @@ namespace Snowflake.Data.Core
                 case SFDataType.TIMESTAMP_NTZ:
 
                     Tuple<long, long> secAndNsec = ExtractTimestamp(srcVal);
-                    return UnixEpoch.AddTicks((long)(secAndNsec.Item1 * 1000 * 1000 * 1000 + secAndNsec.Item2) / 100);
+                    var tickDiff = secAndNsec.Item1 * 10000000L + secAndNsec.Item2 / 100L;
+                    return UnixEpoch.AddTicks(tickDiff);
 
                 default:
                     throw new SnowflakeDbException(SFError.INVALID_DATA_CONVERSION, srcVal, srcType, typeof(DateTime));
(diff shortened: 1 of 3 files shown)


Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -94,7 +94,8 @@ private static DateTime ConvertToDateTime(string srcVal, SFDataType srcType)
                 case SFDataType.TIMESTAMP_NTZ:
 
                     Tuple<long, long> secAndNsec = ExtractTimestamp(srcVal);
-                    return UnixEpoch.AddTicks((long)(secAndNsec.Item1 * 1000 * 1000 * 1000 + secAndNsec.Item2) / 100);
+                    var tickDiff = secAndNsec.Item1 * 10000000L + secAndNsec.Item2 / 100L;
```
ikudjoi: Here overflow is avoided by changing order of evaluation.

Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -235,8 +236,10 @@ private static DateTimeOffset ConvertToDateTimeOffset(string srcVal, SFDataType
                     }
                     else
                     {
-                        DateTime srcDt = ((DateTime)srcVal);
-                        destVal = ((long)(srcDt.Subtract(UnixEpoch).Ticks * 100)).ToString();
+                        DateTime srcDt = (DateTime)srcVal;
+                        var diff = srcDt.Subtract(UnixEpoch);
+                        var tickDiff = diff.Ticks;
+                        destVal = $"{tickDiff}00"; // Cannot multiple tickDiff by 100 because long might overflow.
```
ikudjoi: Here overflow is avoided by not multiplying by 100 but adding trailing zeroes instead.

howryu:
LGTM. I will run try running tests on appveyor. Thanks for your contribution.

howryu:
Seems failed on appveyor https://ci.appveyor.com/project/howryu/snowflake-connector-net/builds/22852486

ikudjoi:
My guess is that is time zone related problem. I'll look into that right away.

PR #115, opened 2019-03-06
#111 Fix handling large DateTime values

ikudjoi:
Replaced PR #112 with this new one after amend & rebase.
Copied from the original PR:

***

When providing fix for #103 in PR #104 (also #110 related because of merge conflicts) I found that there are also other problems related to handling DateTime values. I was kind of hoping that you'd dig in to them and fix them but you didn't so far so here it goes.

I just filed in a separate issue #111 of problems related to large DateTime values and I am providing fix for the issue in this PR. Without these fixes both in inserting to database and reading from it will result in overflowing long value. These changes help avoiding them.

Also removed a test I earlier commented out after replacing it with another one.

### DIFF
--- a/Snowflake.Data.Tests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/SFDataConverterTest.cs
@@ -3,7 +3,6 @@
  */
 
 using System;
-using System.Data;
 
 namespace Snowflake.Data.Tests
 {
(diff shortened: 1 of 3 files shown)


ikudjoi:
As I anticipated, the tests problem was timezone-related. I removed time parts of the dates in date test and everything works on my machine™.

howryu:
Merged in #116

PR #120, opened 2019-04-03
Add SERVICE_NAME support to .NET

ChTimTsubasa:
This PR contains two commits:

1. Refactor RestRequest with: 1) changing its name to RestRequester; 2) Remove all logics of RestRequest from the requester to request (inversion of control), since RestRequester should know nothing about RestRequest.

2. Add SERVICE_NAME support so that it can read SERVICE_NAME parameter from the payload and send out the next request with a service name in the header.

### DIFF
--- a/Snowflake.Data.Tests/Mock/MockServiceName.cs
+++ b/Snowflake.Data.Tests/Mock/MockServiceName.cs
@@ -16,31 +16,6 @@ namespace Snowflake.Data.Tests.Mock
         public const string INIT_SERVICE_NAME = "init";
         public Task<T> PostAsync<T>(IRestRequest request, CancellationToken cancellationToken)
         {
-            return PrepareResponse<T>(HttpMethod.Post, request, cancellationToken);
-        }
-
-        public T Post<T>(IRestRequest postRequest)
-        {
-            return Task.Run(async () => await PostAsync<T>(postRequest, CancellationToken.None)).Result;
-        }
-
-        public T Get<T>(IRestRequest request)
-        {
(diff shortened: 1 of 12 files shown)


ChTimTsubasa:
@howryu Somehow the secret passphrase is not included for my PR. Do I need to be added to some privilege group?

howryu:
Seems tests are running. https://ci.appveyor.com/project/howryu/snowflake-connector-net/builds/23557145

Snowflake.Data.Tests/Mock/MockServiceName.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        public T Get<T>(IRestRequest request)
+        {
+            return Task.Run(async () => await GetAsync<T>(request, CancellationToken.None)).Result;
+        }
+
+        public Task<T> GetAsync<T>(IRestRequest request, CancellationToken cancellationToken)
+        {
+            return PrepareResponse<T>(HttpMethod.Get, request, cancellationToken);
```
howryu: I don't think in your Mock you need this. Get to snowflake server only happened in long running query, where client is asking for query progress/result. In your mock, you always assume POST. Also, in line 44 in this class, you hard code HttpMethod to Post, so I would just return a null response here.

howryu:
LGTM

Snowflake.Data.Tests/Mock/MockServiceName.cs
ChTimTsubasa: Addressed

Issue #64, opened 2018-09-17
Doesn't auto-dispose connection & exception occurs when connection opened inside command


rshenton:
### Issue description
-using statement on connection should auto-dispose connection, but doesn't -> (should override Dispose & call Close)
-should allow connection to be opened in command?

### Example code
```cs
//THIS WORKS AND CLOSES CONNECTION
var con = new Snowflake.Data.Client.SnowflakeDbConnection();
con.ConnectionString = "CONNECTION_STRING_HERE";
try
{   
    con.Open();

    using (var cmd = new Snowflake.Data.Client.SnowflakeDbCommand(con))
    {
        //cmd.CommandType = CommandType.Text;
        cmd.CommandText = query;

        using (var da = new Snowflake.Data.Client.SnowflakeDbDataAdapter(cmd))
        {
            da.Fill(dt);
        }
    }
}
finally
{
    con.Close();
}

//THIS DOESN'T WORK AND FAILS TO CLOSE CONNECTION
using (var con = new Snowflake.Data.Client.SnowflakeDbConnection())
{
    con.ConnectionString = "CONNECTION_STRING_HERE";
                
    using (var cmd = new Snowflake.Data.Client.SnowflakeDbCommand(con))
    {
        cmd.CommandType = CommandType.Text;
        cmd.CommandText = query;

        con.Open();
        using (var da = new Snowflake.Data.Client.SnowflakeDbDataAdapter(cmd))
        {
            da.Fill(dt);
        }
    }

    //con.Close(); //This line doesn't matter since exception happens above and gets bypassed
}
```

### Configuration
Snowflake.Data(1.0.4)
netcoreapp2.0

howryu:
Thanks for your report. auto-dipose is a known issue, we have some other resources are not auto-dispose, #20.

For the second one, I will check why it is not working.

PR #121, opened 2019-04-04
Implement Dispose for Connection and Transaction.

### LINKED ISSUE #20: Dispose should rollback transaction if not committed
The SnowflakeDbTransaction does not properly Rollback a transaction upon dispose, I'm not sure where the official guidelines for extending DbTransaction are but the expected behavior is that the protected virtual Dispose method is overridden to Rollback the transaction upon Dispose if the transaction was not committed.

ChTimTsubasa:
This PR is to resolve #20 and #64 
There are four classed that implements IDisposable interface:
1. Connection
2. Transaction
3. Command
4. DataReader

I only implemented Dispose() function for connection and transaction. The dispose would call close() and rollback() respectably.

Our Command and DataReader does not hold any resources so there is not need to override the Dispose() on them.

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -217,5 +217,42 @@ namespace Snowflake.Data.Tests
                 conn.Close();
             }
         }
+
+        // Test that when a connection is disposed, a close would send out and unfinished transaction would be roll back.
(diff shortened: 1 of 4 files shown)


ChTimTsubasa:
@howryu Can you review this PR?

howryu:
LGTM

PR #128, opened 2019-05-17
Support Null valued Parameter

ChTimTsubasa:
In the current code base we just convert the null parameter to empty string.

### DIFF
--- a/Snowflake.Data.Tests/SFBindTestIT.cs
+++ b/Snowflake.Data.Tests/SFBindTestIT.cs
@@ -56,7 +56,39 @@ namespace Snowflake.Data.Tests
         [Test]
         public void testBindNullValue()
         {
-
+            using (SnowflakeDbConnection dbConnection = new SnowflakeDbConnection())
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/SFBindTestIT.cs
```diff
@@ -56,7 +56,51 @@ public void testArrayBind()
...
+                {
+                    using (IDbCommand command = dbConnection.CreateCommand())
+                    {
+                        command.CommandText = "create or replace table TEST_TBL (ID number);";
+                        command.ExecuteNonQuery();
+                    }
+                    
+                    using (IDbCommand command = dbConnection.CreateCommand())
+                    {
+                        command.CommandText = "insert into TEST_TBL values(:p0)";
```
howryu: So we already support name based binding?

ChTimTsubasa: Some how this works so yes.

Issue #28, opened 2018-04-18
Support authentication with Okta SSO


jsnplank:
This is the connection string I'm using: "scheme=https;host=snagajob.snowflakecomputing.com;port=443;account=snagajob;user=MYEMAIL@snagajob.com;password=MYPASSWORD;authenticator=https://snagajob.okta.com";

I know the authenticator key may not be currently supported by the .NET driver but it's what we use in different projects that utilize the JDBC driver.  

The error I'm getting when trying to connect is: "Incorrect username or password was specified"

I'm using the same username and password account I use for Okta and I login to the Snowflake web ui through Okta

howryu:
Right. This is not currently supported by .net driver. I will change the issue name and mark this as a feature.

howryu:
I talked this with our PM and we think we can target this at the end of May. Currently I have other projects in my pipeline.

celenaf:
Similar to Okta SSO, will there be support for ADFS coming soon?

RobertBest63:
So does anybody know if this got resolved?  Trying it now with the authenticator token and still getting the error.

howryu:
@RobertBest63 This is scheduled to be done on Q1 of 2019.

vinaykumar80:
@howryu @hkapre : Are we on track to have this feature integrated by end of Q1 0f 2019?

abelykh0:
What about Q2 0f 2019? Any chance?

ChTimTsubasa:
I will be working on this

hkapre:
End of July is the current target.

Targets are always tentative and we may need to push this out for a variety
of reasons. We are trying to pull this in earlier than July but July is a
safer bet for now.

Harsha

PR #132, opened 2019-05-30
Native Okta support

ChTimTsubasa:
The main purpose of this PR is to implement Native Okta support, requested from #28.
Manual e2e test and Mock Unit test for error cases is created.

Other than that, some refactor is done for RestRequest/RestRequester. Majorly:
1. Relocate some of the classes in RestRequester
2. Create a base abstract RestRequester to hold various timeouts.


Snowflake.Data/Core/Authenticator/IAuthenticator.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        /// Generate the authenticator given the session
+        /// </summary>
+        /// <param name="session">session that requires the authentication</param>
+        /// <returns>authenticator</returns>
+        /// <exception cref="SnowflakeDbException">when authenticator is unknown</exception>
+        internal static IAuthenticator GetAuthenticator(SFSession session)
+        {
+            string type = session.properties[SFSessionProperty.AUTHENTICATOR];
+            switch (type)
```
howryu: Maybe use if/elseif, switch seems not good here since you cannot switch for all cases.

Snowflake.Data/Core/Authenticator/IAuthenticator.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                case "snowflake":
+                    return new BasicAuthenticator(session);
+            }
+
+            if (type.EndsWith("okta.com") && type.StartsWith("https://"))
+            {
+                return new OktaAuthenticator(session, type);
```
howryu: Add comment here that type is actually okta url. Otherwise here it seems that type should be something like `okta`.

howryu:
One more thing, can you update the README.md with some mention about okta support? Maybe a simple example?

ChTimTsubasa:
@howryu Added in #135

PR #134, opened 2019-05-30
Snow 79080

smtakeda:


Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -59,7 +59,9 @@ public void SFTestSetup()
             if (testConfigs.TryGetValue(connectionKey, out testConnectionConfig))
             {
                 connectionString = String.Format(connectionStringFmt,
+                    testConnectionConfig.protocol,
```
howryu: Can you use https as default protocol if not specified? In the example in https://github.com/snowflakedb/snowflake-connector-net/blob/master/README.md, I did not provide protocol and existing customer who is developing against our production environment may break once they update.

smtakeda: yes, already set in the constructor of TestConfig

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -101,6 +106,15 @@ public class TestConfig
...
         [JsonProperty(PropertyName = "SNOWFLAKE_TEST_ROLE", NullValueHandling = NullValueHandling.Ignore)]
         internal string role { get; set; }
+
+        [JsonProperty(PropertyName = "SNOWFLAKE_TEST_PROTOCOL", NullValueHandling = NullValueHandling.Ignore)]
+        internal string protocol { get; set; }
+
+        public TestConfig()
+        {
+            this.protocol = "https";
+            this.port = "443";
```
smtakeda: @howryu these are the default values.

smtakeda:
ah, we need to check https://ci.appveyor.com/project/howryu/snowflake-connector-net/history

Issue #138, opened 2019-06-10
Deadlock when calling SnowflakeDbConnection.Open() from ASP.NET MVC


stegus64:
### Issue description
Deadlock when calling SnowflakeDbConnection.Open() from ASP.NET MVC

The problem is likely caused by the code in SFSession.Open() calling an async method without using ConfigureAwait(false)

See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html for more information.

To reproduce, create an ASP.NET MVC application for .NET Framework 4.6 using Visual Studio 2017. Add the code below to a controller.

### Example code
```cs
        [HttpGet]
        public IEnumerable<TestRow> Get()
        {
            var result = new List<TestRow>();

            using (var conn = new SnowflakeDbConnection())
            {
                conn.ConnectionString = connectionString;
                conn.Open(); // <<<------ Deadlock here

                IDbCommand cmd = conn.CreateCommand();
                cmd.CommandText = "SELECT TABLE_NAME, COMMENT FROM INFORMATION_SCHEMA.TABLES";
                IDataReader reader = cmd.ExecuteReader();

                while (reader.Read())
                {
                    var r = new TestRow() { table_name = reader.GetString(0), comment = reader.GetString(1) };
                    result.Add(r);
                }

                conn.Close();
            }

            return result;
        }
```

### Workaround
Declare the Controller method to be async and call OpenAsync instead, like this:

```cs
        [HttpGet]
        public async Task<IEnumerable<TestRow>> Get()
        {
            var result = new List<TestRow>();

            using (var conn = new SnowflakeDbConnection())
            {
                conn.ConnectionString = connectionString;
                await conn.OpenAsync(); 
}
}

```

### Configuration
*Driver version: 1.0.10*

*Dotnet framework and version: .net Framework 4.6, Visual Studio 2017*

*Client OS:*  Windows 10

PR #139, opened 2019-06-13
Fix Okta authentication Deadlock issue 

ChTimTsubasa:
This is a fix to address #138 
Have separate code routine to do sync authentication;
Using threadpool for execution.

### DIFF
--- a/Snowflake.Data.Tests/Mock/MockCloseSessionGone.cs
+++ b/Snowflake.Data.Tests/Mock/MockCloseSessionGone.cs
@@ -29,6 +29,11 @@ namespace Snowflake.Data.Tests.Mock
             return Task.FromResult<HttpResponseMessage>(null);
         }
 
+        public HttpResponseMessage Get(IRestRequest request)
+        {
+            return null;
+        }
+
         public T Post<T>(IRestRequest postRequest)
         {
             return Task.Run(async () => await PostAsync<T>(postRequest, CancellationToken.None)).Result;
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -94,7 +94,8 @@ namespace Snowflake.Data.Tests
                 }
                 catch (AggregateException e)
                 {
-                    Assert.AreEqual(270007, ((SnowflakeDbException)e.InnerException).ErrorCode);
+                    Assert.AreEqual(SFError.REQUEST_TIMEOUT.GetAttribute<SFErrorAttr>().errorCode, 
+                        ((SnowflakeDbException)e.InnerException).ErrorCode);
                 }
                 Assert.AreEqual(5, conn.ConnectionTimeout);
             }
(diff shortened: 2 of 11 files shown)


Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -92,9 +92,9 @@ public void TestLoginTimeout()
                     conn.Open();
                     Assert.Fail();
                 }
-                catch(SnowflakeDbException e)
+                catch (AggregateException e)
                 {
-                    Assert.AreEqual(SFError.REQUEST_TIMEOUT.GetAttribute<SFErrorAttr>().errorCode,e.ErrorCode);
+                    Assert.AreEqual(270007, ((SnowflakeDbException)e.InnerException).ErrorCode);
```
smtakeda: Use const variable. Don't we have a list of error code for .NET driver?

smtakeda:
Add test cases for OKTA auth.

howryu:
Can you elaborate a little bit more what cause the issue and what is the fix? It seems that you just add a synchronous `Authenticate` method? Why async does not work?

ChTimTsubasa:
@howryu Basically I was calling .Result, which will definitely result in deadlock. Based on some research (especially https://medium.com/rubrikkgroup/understanding-async-avoiding-deadlocks-e41f8f2c6f5d), As a library code, our best practice would be:
1. implementing both async and sync method, and separate code routine for best performance;
2. Implement our own syncContext and TaskScheduler to avoid deadlock in any circumstances.

This PR does not include 2 given that it is too much work and this PR should go fast to unblock users. But Task.Run() is used to mitigate deadlock issue. Another JIRA would be open to really finish 2.

ChTimTsubasa:
@smtakeda There is already a manual test called TestOktaConnection which would test against OKTA authentication on local reg. Maybe we can consider having automated native OKTA test on Prod server.

Snowflake.Data.Tests/SFConnectionIT.cs
ChTimTsubasa: Addressed

Snowflake.Data.Tests/Mock/MockCloseSessionGone.cs
```diff
@@ -29,6 +29,11 @@ public Task<HttpResponseMessage> GetAsync(IRestRequest request, CancellationToke
...
         }
 
+        public HttpResponseMessage Get(IRestRequest request)
+        {
+            return null;
+        }
+
         public T Post<T>(IRestRequest postRequest)
         {
             return Task.Run(async () => await PostAsync<T>(postRequest, CancellationToken.None)).Result;
```
howryu: So in theory this could cause deadlock as well right?

ChTimTsubasa: Yes

howryu:
LGTM. That's really good article to read about async coding in .NET.

Issue #140, opened 2019-06-17
More Deadlocks when calling from ASP.NET MVC


stegus64:
### Issue description
This is related to Issue #138 
There are a few more places in the code where you access the Result property of an Async task which can cause deadlock for example when calling from ASP.NET MVC

After the fix for #138, there is no longer a deadlock in SnowflakeDbConnection.Open(), but there is a deadlock in IDbDataReader.Read() if the resultset contains more than one chunk.

### Example code
```cs
        [HttpGet]
        public IEnumerable<TestRow> Get()
        {
            var result = new List<TestRow>();

            using (var conn = new SnowflakeDbConnection())
            {
                conn.ConnectionString = connectionString;
                conn.Open(); // <<<------ Deadlock here (fixed by #138)

                IDbCommand cmd = conn.CreateCommand();
                cmd.CommandText = "SELECT TABLE_NAME, COMMENT FROM INFORMATION_SCHEMA.TABLES";
                IDataReader reader = cmd.ExecuteReader();

                while (reader.Read()) //<------- Deadlock here if result contains multiple chunks
                {
                    var r = new TestRow() { table_name = reader.GetString(0), comment = reader.GetString(1) };
                    result.Add(r);
                }

                conn.Close();
            }

            return result;
        }
```
I searched the codebase for the regular expression Async\(.+Result to see if there are any more problems, and I found the following cases you should probably have a look at:

```cs
Find all "Async\(.+Result", Regular expressions, Subfolders, Keep modified files open, Find Results 1, "Entire Solution ( Including External Items )", "*.cs"
  C:\Work\snowflake-connector-net\Snowflake.Data\Core\Authenticator\OktaAuthenticator.cs(70):            var samlRawHtmlString = samlRawResponse.Content.ReadAsStringAsync().Result;
  C:\Work\snowflake-connector-net\Snowflake.Data\Core\SFBlockingChunkDownloader.cs(127):            Stream stream = httpResponse.Content.ReadAsStreamAsync().Result;
  C:\Work\snowflake-connector-net\Snowflake.Data\Core\SFResultSet.cs(110):            if ((nextChunk = _chunkDownloader?.GetNextChunkAsync().Result) != null)
```
The code in SFResultSet.cs line110 is the direct casue for the deadlock in .Read()

### Configuration
*Driver version: 1.0.10 prep-1.0.11*

*Dotnet framework and version: .net Framework 4.6, Visual Studio 2017*

[...]

PR #144, opened 2019-06-21
SNOW-82602 Improve deadlocks in the .NET driver

ChTimTsubasa:
Requested from #140

### DIFF
--- a/Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
+++ b/Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
@@ -67,7 +67,7 @@ namespace Snowflake.Data.Core.Authenticator
             logger.Debug("step 4: get SAML reponse from sso");
             var samlRestRequest = BuildSAMLRestRequest(ssoUrl, onetimeToken);
             var samlRawResponse = await session.restRequester.GetAsync(samlRestRequest, cancellationToken);
-            var samlRawHtmlString = samlRawResponse.Content.ReadAsStringAsync().Result;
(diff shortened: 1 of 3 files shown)


Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
```diff
@@ -67,7 +67,7 @@ async Task IAuthenticator.AuthenticateAsync(CancellationToken cancellationToken)
             logger.Debug("step 4: get SAML reponse from sso");
             var samlRestRequest = BuildSAMLRestRequest(ssoUrl, onetimeToken);
             var samlRawResponse = await session.restRequester.GetAsync(samlRestRequest, cancellationToken);
-            var samlRawHtmlString = samlRawResponse.Content.ReadAsStringAsync().Result;
+            var samlRawHtmlString = await samlRawResponse.Content.ReadAsStringAsync();
```
smtakeda: Let me understand about `await`. So this will spawn a thread to run ReadAsStringAsync() and the main thread will wait for it to finish in line 73?

ChTimTsubasa: Not necessary spawning a thread right here. The compiler would rewrite the code here to be a state machine, and decide whether it should proceed or return. This is the same pattern as asynchronous programming, but .NET allow you to yield the ugly work to compiler using async and await keywords.

For further reading: https://devblogs.microsoft.com/premier-developer/dissecting-the-async-methods-in-c/

PR #148, opened 2019-06-25
@SNOW-81851 add filter to ACCOUNT 

ChTimTsubasa:
It would remove the connection group and the region name.
I did not change any document because the scope for the statement "hostname field is going to be deprecated" in [JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-76706). In that JIRA SnowSQL is going to be deprecate hostname, but not mentioning other drivers.

@smtakeda @notkriswagner Do we want every driver to follow that SnowSQL and deprecate hostname, using only account to specify the account, region and connection group?

If yes then I will change the document accordingly, otherwise I would not modify the document since we already encourage the customers to use hostname rather than account to form URL in our docs.

### DIFF
--- a/Snowflake.Data/Core/SFSessionProperty.cs
+++ b/Snowflake.Data/Core/SFSessionProperty.cs
@@ -50,6 +50,35 @@ namespace Snowflake.Data.Core
     {
         static private SFLogger logger = SFLoggerFactory.GetLogger<SFSessionProperties>();
 
+        public override bool Equals(object obj)
+        {
+            if (obj == null) return false;
+            try
+            {
+                SFSessionProperties prop = (SFSessionProperties)obj;
+                foreach (SFSessionProperty sessionProperty in Enum.GetValues(typeof(SFSessionProperty)))
+                {
+                    if (this.ContainsKey(sessionProperty) ^ prop.ContainsKey(sessionProperty))
+                    {
+                        return false;
+                    }
+                    if (!this.ContainsKey(sessionProperty))
+                    {
+                        continue;
+                    }
+                    if (!this[sessionProperty].Equals(prop[sessionProperty]))
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            } catch (InvalidCastException)
+            {
+                logger.Warn("Invalid casting to SFSessionProperties");
+                return false;
+            }
+        }
+
         internal static SFSessionProperties parseConnectionString(String connectionString, SecureString password)
         {
             logger.Info("Start parsing connection string.");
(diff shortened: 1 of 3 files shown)


smtakeda:
`hostname` is not deprecated but `region`.

smtakeda:
And don't remove `region` unless we do preannouncement.

Snowflake.Data/Core/SFSessionProperty.cs
```diff
@@ -50,6 +50,35 @@ class SFSessionProperties : Dictionary<SFSessionProperty, String>
     {
         static private SFLogger logger = SFLoggerFactory.GetLogger<SFSessionProperties>();
 
+        public override bool Equals(object obj)
+        {
+            if (obj == null) return false;
+            try
+            {
+                SFSessionProperties prop = (SFSessionProperties)obj;
```
smtakeda: do you check the number of properties?

Snowflake.Data/Core/SFSessionProperty.cs
```diff
@@ -140,6 +171,26 @@ private static bool IsRequired(SFSessionProperty sessionProperty, SFSessionPrope
...
+            // filter the .<region> or .global
+            string parsedAccount = rawAccount;
+            int firstDot = parsedAccount.IndexOf('.');
+            if (firstDot != -1)
+            {
+                parsedAccount = parsedAccount.Substring(0, firstDot);
+            }
+
+            // filter the connection group
+            int firstDash = parsedAccount.IndexOf('-');
```
smtakeda: Today this is valid only of `global` is included in the subdomain?

notkriswagner: Yes, the filtering of the connection group should only be done if it contains `global`. I would also use `LastIndexOf` instead of `IndexOf`. Today we don't allow dashes in the account name, but if we did in the future, this may parse out too much information. The external ID will always appended to the end so `LastIndexOf` would always work.

Snowflake.Data/Core/SFSessionProperty.cs
ChTimTsubasa: line 61 would check every property should either both exist or both not exist.

Snowflake.Data/Core/SFSessionProperty.cs
ChTimTsubasa: https://github.com/snowflakedb/snowflake-connector-net/tree/prep-1.0.12#create-a-connection

ChTimTsubasa: We suggested users don't use account but rather using hostname.

notkriswagner:
LGTM, thanks

ChTimTsubasa:
@notkriswagner and I reach an consensus that we still encourage the customers to use hostname to fill in the global url information, and kept account being an account. That way we can reduce the complexity of connection fields construction in .NET and be consistent with the original principle: Account name should just be account name rather than a mix of everything.

Docs are updated to reflect this change.

Issue #153, opened 2019-07-15
IDataReader.Read() does not terminate


VantageKR:
### Issue description
When calling IDataReader's Read() method, when everything has been read, should return false.
This was the behaviour up until the recent package update - now it does not terminate and blocks on the .Read() call.

                 

### Example code
```csharp
IDataReader reader = cmd.ExecuteReader();
                DataTable data = new DataTable();
                data.Load(reader);  // does not terminate
```

or...

```csharp
IDataReader reader = cmd.ExecuteReader();
                DataTable data = new DataTable();
                //data.Load(reader);
                    DataTable schemaTable = reader.GetSchemaTable();
                    foreach (DataRow row in schemaTable.Rows)
                    {
                        string colName = row.Field<string>("ColumnName");
                        Type t = row.Field<Type>("DataType");
                        data.Columns.Add(colName, t);
                    }
                    while (reader.Read())
                    {
                        var newRow = data.Rows.Add();
                        foreach (DataColumn col in data.Columns)
                        {
                            newRow[col.ColumnName] = reader[col.ColumnName];
                        }
                    }                
```
### Error log
No actual errors, but the log terminates with this:
```
 INFO 32 Snowflake.Data.Core.SFResultSetMetaData - Found colun name IDENTITY_CYCLE under index 40

 INFO 32 Snowflake.Data.Core.SFResultSetMetaData - Found colun name COMMENT under index 41

 INFO 32 Snowflake.Data.Core.SFResultSet - Get next chunk from chunk downloader

 INFO 23 Snowflake.Data.Core.SFBlockingChunkDownloaderV3 - NextChunkToConsume: 0, NextChunkToDownload: 1


```

### Configuration
*Driver version:*
Version:  1.0.12
*Dotnet framework and version:* E.g. .net framework 4.5.2 or .net standard 2.0
.Net 4.6.2
*Server version:* E.g. 1.90.1
3.32.1 

*Client OS:* E.g. Debian 8.1 (Jessie), Windows 10
Windows 10

VantageKR:
This problem is not present in version 1.0.5 but this seems like a very old version compared to the latest?
What is the risk of running version 1.0.5 and not the latest stable please?
What changed from version 1.0.5 that might have an effect on this usage pattern, involving DataTable, DataRow and DataReader?

ChTimTsubasa:
[...]

Issue #154, opened 2019-07-31
Incorrect result when inserting into DATE column when client timezone is ahead of UTC


stegus64:
### Issue description
The test SFDbDataReader.TestGetDate() fails when running on a client that is ahead of UTC (Such as UTC+1, Stockholm)

The reason is probably the following code in SFDataConverter:208
```
                        long millis = (long)((DateTime)srcVal).ToUniversalTime().Subtract(
                            UnixEpoch).TotalMilliseconds;

```
ToUniversalTime() will change the date by subtracting from midnight when the client is ahead of UTC.
This means the wrong value is inserted into the database.

Everything works correctly when the client is after UTC such as in America.

The call to ToUniversalTime() should simply be removed.

### Configuration
*Driver version:* 1.0.12

*Dotnet framework and version:* .net standard 2.0

*Client OS:*  Windows 10

stegus64:
@ChTimTsubasa Please remember that this issue has not been fixed. pr #156 should fix this, but that pr has been closed without merging.

ChTimTsubasa:
Merged through #168

PR #155, opened 2019-08-02
Fix for #153

stegus64:
This fixes issue #153 by adding ConfigureAwait(false) to all await points in the library.
It also modifies the unit tests so any future problems like this will cause unit tests to fail.

### DIFF
--- a/Snowflake.Data.Tests/DedicatedThreadSynchronisationContext.cs
+++ b/Snowflake.Data.Tests/DedicatedThreadSynchronisationContext.cs
@@ -1,0 +1,128 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+
+namespace Snowflake.Data.Tests
+{
+    /*
+     * This class will not deadlock, but it will cause tests to fail if the Send or Post methods have been called during a test
+     */
+    public sealed class MockSynchronizationContext : SynchronizationContext
+    {
+        int callCount = 0;
+
+        public override void Post(SendOrPostCallback d, object state)
+        {
+            callCount++;
+            base.Post(d, state);
+        }
+
+        public override void Send(SendOrPostCallback d, object state)
+        {
+            callCount++;
+            base.Send(d, state);
+        }
+
+        public static void SetupContext()
+        {
+            SynchronizationContext.SetSynchronizationContext(new MockSynchronizationContext());
+        }
+
+        public static void Verify()
+        {
+            MockSynchronizationContext ctx = (MockSynchronizationContext)SynchronizationContext.Current;
+            Assert.Zero(ctx.callCount, "MockSynchronizationContext was called - this can cause deadlock. Make sure ConfigureAwait(false) is used in every await point in the library");
+            SynchronizationContext.SetSynchronizationContext(null);
+        }
+    }
+
+    /*
+     * This can be used to test what happens when a library metod is called from a SyncronizationContext.
+     * If there are any deadlocks in the code, this will trigger the deadlock.
+     * 
+     * Usage: 
+     *      DedicatedThreadSynchronisationContext.RunInContext(_ => TestSimpleLargeResultSet());
+     * 
+     */
+    public sealed class DedicatedThreadSynchronisationContext : SynchronizationContext, IDisposable
+    {
+        public DedicatedThreadSynchronisationContext()
+        {
+            m_thread = new Thread(ThreadWorkerDelegate);
+            m_thread.Start(this);
+        }
+
+        public void Dispose()
+        {
+            m_queue.CompleteAdding();
+        }
+
+        /// <summary>Dispatches an asynchronous message to the synchronization context.</summary>
+        /// <param name="d">The System.Threading.SendOrPostCallback delegate to call.</param>
+        /// <param name="state">The object passed to the delegate.</param>
+        public override void Post(SendOrPostCallback d, object state)
+        {
+            if (d == null) throw new ArgumentNullException("d");
+            m_queue.Add(new KeyValuePair<SendOrPostCallback, object>(d, state));
+        }
+
+        /// <summary> As 
+        public override void Send(SendOrPostCallback d, object state)
+        {
+            using (var handledEvent = new ManualResetEvent(false))
+            {
+                Post(SendOrPostCallback_BlockingWrapper, Tuple.Create(d, state, handledEvent));
+                handledEvent.WaitOne();
+            }
+        }
+
+        public int WorkerThreadId { get { return m_thread.ManagedThreadId; } }
+
+        // This will run the callback in a synchronizationContext that is equivalent to a GUI or ASP.Net program 
+        // If there are any async problems in the method, this code will provoke the deadlock.
+        public static void RunInContext(SendOrPostCallback d)
+        {
+            using (var ctx = new DedicatedThreadSynchronisationContext())
+            {
+                ctx.Send(d, null);
+            }
+        }
+
+        //=========================================================================================
+
+        private static void SendOrPostCallback_BlockingWrapper(object state)
+        {
+            var innerCallback = (state as Tuple<SendOrPostCallback, object, ManualResetEvent>);
+            try
+            {
+                innerCallback.Item1(innerCallback.Item2);
+            }
+            finally
+            {
+                innerCallback.Item3.Set();
+            }
+        }
+
+        /// <summary>The queue of work items.</summary>
+        private readonly BlockingCollection<KeyValuePair<SendOrPostCallback, object>> m_queue =
+            new BlockingCollection<KeyValuePair<SendOrPostCallback, object>>();
+
+        private readonly Thread m_thread = null;
+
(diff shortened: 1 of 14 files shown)


stegus64:
Added fix also for #154

Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -205,8 +205,7 @@ private static DateTimeOffset ConvertToDateTimeOffset(string srcVal, SFDataType
                     }
                     else
                     {
-                        long millis = (long)((DateTime)srcVal).ToUniversalTime().Subtract(
```
ChTimTsubasa: It's probably a good idea to change the behavior. We can change the test case to specify the input as UTC time with a Z.

Snowflake.Data.Tests/DedicatedThreadSynchronisationContext.cs
```diff
@@ -0,0 +1,128 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+
+namespace Snowflake.Data.Tests
```
ChTimTsubasa: Awesome mock case!

ChTimTsubasa:
Other than adding ConfigureAwait(false), we may also want to consider the method described in https://devblogs.microsoft.com/premier-developer//one-user-scenario-to-rule-them-all/ to have a custom awaiter. 2 benefits are:
1. We don't need to attach ConfigureAwait(false) to every line of await;
2. For 2nd and 3rd party code that we used but don't have control on, we can avoid having deadlock.

Snowflake.Data/Core/SFDataConverter.cs
stegus64: Not really sure what you mean by this. 

Please note that the problem is not in the test case. I dont think you should change the test case.
The issue is that if your client computer is in a timezone after UTC and you send in a normal datetime (with only a date and no timezone specified), the wrong date will be stored in snowflake.
As I said, this can be fixed simply by removing the call to ToUniversalTime()

It is not easy to write a unit test for this scenario since there really is no good way to change the timezone programmatically. The time zone setting in windows is system-wide - it can not be changed per thread or per process. But you can easily recreate the scenario by manually changing the timezone of your development computer to UTC+1 before running the unit tests.

stegus64:
> Other than adding ConfigureAwait(false), we may also want to consider the method described in https://devblogs.microsoft.com/premier-developer//one-user-scenario-to-rule-them-all/ to have a custom awaiter. 2 benefits are:
> 
> 1. We don't need to attach ConfigureAwait(false) to every line of await;
> 2. For 2nd and 3rd party code that we used but don't have control on, we can avoid having deadlock.

Interesting idea. I had not seen that article.

But note that you would have to add 
`await Awaiters.DetachCurrentSyncContext();`
To every async method in the library that could possibly be called directly or indirectly from a public method. This might be almost as much work as adding ConfigureAwait() at all await points in the library.

But I agree that it is very good that this will work even for 3rd party code.

One possible drawback of DetachCurrentSyncContext() is that it will use an extra threadpool thread when calling async library methods from a synchronization context.
I am not sure how much of a problem that is, but it seems that this method will make performance slightly worse, while adding ConfigureAwait(false) will make performance slightly better. This might make a difference in high-frequency scenarios like calling DataReader.ReadAsync() in a tight loop.

I have not tested this so I dont know how much difference it would make.

Snowflake.Data/Core/SFDataConverter.cs
stegus64: I suggest that you modify csharpTypeValToSfTypeVal to actually return a Date only, so regardless of the time part specified and the kind of Datetime you should always return a date only. 

This can be done with the following change:

`                case DbType.Date:
                    destType = SFDataType.DATE.ToString();
                    if (srcVal.GetType() != typeof(DateTime))
                    {
                        throw new SnowflakeDbException(SFError.INVALID_DATA_CONVERSION, srcVal, 
                            srcVal.GetType().ToString(), DbType.Date.ToString());
                    }
                    else
                    {
                        DateTime dt = ((DateTime)srcVal).Date;
                        var ts = dt.Subtract(UnixEpoch);
                        long millis = (long)(ts.TotalMilliseconds);
                        destVal = millis.ToString();
                    }
                    break;
`

I also suggest adding the following test case to make sure it works regardless of the kind of datetime used and the timezone the computer is in:

`        [Test]
        [TestCase("2100-12-31 23:59:59.9999999", DateTimeKind.Utc)]
        [TestCase("2100-12-31 23:59:59.9999999", DateTimeKind.Local)]
        [TestCase("2100-12-31 23:59:59.9999999", DateTimeKind.Unspecified)]
        [TestCase("2200-01-01 00:00:00.0000000", DateTimeKind.Utc)]
        [TestCase("2200-01-01 00:00:00.0000000", DateTimeKind.Local)]
        [TestCase("2200-01-01 00:00:00.0000000", DateTimeKind.Unspecified)]
        [TestCase("1960-01-01 00:00:00.0000000", DateTimeKind.Unspecified)]
        [TestCase("9999-12-31 23:59:59.9999999", DateTimeKind.Unspecified)]
        [TestCase("1982-01-18 16:20:00.6666666", DateTimeKind.Unspecified)]
        [TestCase("1982-01-18 23:59:59.0000000", DateTimeKind.Unspecified)]
        [TestCase(null, DateTimeKind.Unspecified)]
        public void TestConvertDate(string inputTimeStr, object kind=null)
        {
            if (kind == null)
                kind = 0;
            DateTime inputTime;
            if (inputTimeStr == null)
            {
                inputTime = DateTime.Now;
            }
            else
            {
                inputTime = DateTime.ParseExact(inputTimeStr, "yyyy-MM-dd HH:mm:ss.fffffff", CultureInfo.InvariantCulture);
            }
            var dtExpected = inputTime.Date;
            internalTestConvertDate(dtExpected, DateTime.SpecifyKind(inputTime, (DateTimeKind)kind));
        }

        private void internalTestConvertDate(DateTime dtExpected, DateTime testValue)
        {
            var result = SFDataConverter.csharpTypeValToSfTypeVal(System.Data.DbType.Date, testValue);
            // Convert result to DateTime for easier interpretation
            var unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
            DateTime dtResult = unixEpoch.AddMilliseconds(Int64.Parse(result.Item2));
            Assert.AreEqual(dtExpected, dtResult);
        }
`

ChTimTsubasa:
My understanding is that we only need to call `await Awaiters.DetachCurrentSyncContext();` at the start of each public method or the starting point where we call async methods inside a sync method. Once the customer awaiter swaps in, the following awaits in the same methods or in the called methods won't have deadlock anymore. Otherwise the 3rd party code case won't work.

stegus64:
> My understanding is that we only need to call `await Awaiters.DetachCurrentSyncContext();` at the start of each public method or the starting point where we call async methods inside a sync method. Once the customer awaiter swaps in, the following awaits in the same methods or in the called methods won't have deadlock anymore. Otherwise the 3rd party code case won't work.

Yes, this was what i meant.

But it is not always so easy to find which async methods are called by sync methods. It is much easier to just make sure that all async methods have this line at the beginning. Just like it is much easier to add ConfigureAwait() to all await points rather than trying to figure out where it is actually needed.

ChTimTsubasa:
> > My understanding is that we only need to call `await Awaiters.DetachCurrentSyncContext();` at the start of each public method or the starting point where we call async methods inside a sync method. Once the customer awaiter swaps in, the following awaits in the same methods or in the called methods won't have deadlock anymore. Otherwise the 3rd party code case won't work.
> 
> Yes, this was what i meant.
> 
> But it is not always so easy to find which async methods are called by sync methods. It is much easier to just make sure that all async methods have this line at the beginning. Just like it is much easier to add ConfigureAwait() to all await points rather than trying to figure out where it is actually needed.

I see what you meant. Would the testcase you added failed if there is a 3rd party code that we used does not follow the ConfigureAwait(false) style? If yes then I would agree with the ConfigureAwait(false) solution

stegus64:
> I see what you meant. Would the testcase you added failed if there is a 3rd party code that we used does not follow the ConfigureAwait(false) style? If yes then I would agree with the ConfigureAwait(false) solution

Yes, the testcase should fail even if the problem is in 3rd party code. I think it would fail for any situation that would cause an async deadlock.

ChTimTsubasa:
@stegus64 Can you remove the fix to #154 in this PR and open another PR for that? We need to make an announcement before hand since that would be an behavior change and let's don't delay this one.

I will submit a testcase PR of your fix for #153 and if that passes we can check in your change.

stegus64:
This is actually the first time I use a fork and pull requests with a public github project.
I originally tried to create a separate pull request, but I did something wrong so it ended up on the same pr.

Do you have any tips for how I should do this?

stegus64:
I have now tried to separate the fix for #154 into its own pr. I hope it is correct.

ChTimTsubasa:
@stegus64 Committed to prep-1.0.13 and will be release through 1.0.13 next week. Thanks for your contribution!

Issue #180, opened 2019-12-12
NullReferenceError with DbCommand created before opening Connection 


ThijsKoot:
### Issue description
When a DbCommand is created from a connection before the connection has been opened a first time, executing the DbCommand results in a NullReferenceException. This is because the SnowflakeDbCommand-constructor assigns the value of the sfSession-property of SnowflakeDbConnection to itself. The problem here is that sfSession is set when opening the connection, meaning that DbCommand assigns a null value if the connection has never been opened. I'm creating a pull request with a fix for this. 

### Example code
```cs
using (var conn = new SnowflakeDbConnection { ConnectionString = connstr })
{
	var cmd = conn.CreateCommand();
	cmd.CommandText = "select current_version()";
	
	conn.Open();
	
	var result = cmd.ExecuteScalar();
	
	conn.Close();
}
```

### Error log
Not required, problem is in the connector itself.

### Configuration
*Driver version:* 1.1.0

*Dotnet framework and version:* .net standard 2.0

ChTimTsubasa:
Thank you for reporting this issue and submitting a PR.

PR #181, opened 2019-12-12
Create SfStatement when executing DbCommand #180

ThijsKoot:
I delayed the creation of SfStatement until it's needed: when executing a command. Behavior does not change with this implementation. All tests pass and I added a new test to test for the issue this fix addresses.


ThijsKoot:
Build failed because pipeline couldn't decrypt parameters.json.enc. Not sure if there's anything I can do about that? _Works on my machine™_. Jokes aside, all tests pass and building release config works as well.

PR #185, opened 2020-02-04
SNOW-118679 Add GCP Add GCS appveyor tests

ChTimTsubasa:


ChTimTsubasa:
I need to skip SimpleLargeResultTest as it failed consistently against GCP on Appveyor. Local test cannot recreate the issue. Will open another JIRA to track it.

PR #188, opened 2020-02-20
Remove raw message logging

ChTimTsubasa:


snowstakeda:
@ankit-bhatnagar167 , this would be an emergency release.

ChTimTsubasa:
@ankit-bhatnagar167 I will directly check in to master since there is no prep branch and the new version number is not clear to me.

PR #200, opened 2020-04-29
Fix sequential okta auth failure

vese:
There is an issue when using OktaAuthentificator.
Request for idp (step 3)  leads to 403 Forbidden if HttpClient has cookie sid, which applies in step 4.
So that second connection to sf through okta fails.


vese:
#258 
#272

Issue #56, opened 2018-08-29
JsonReaderException on Cancel()


Castlejoe:
Hi,

When snowflake is very busy, you risk to cancel (timeout kicked in) an async command before it actually was run. In this case, the cancel post returns a json that looks like this:

```javascript
{
    "data": {
        "internalError": false,
        "errorCode": "000605",
        "age": 0,
        "sqlState": "01000",
        "queryId": null
    },
    "message": "Identified SQL statement is not currently executing.",
    "code": "000605",
    "success": false
}
```

This is then tried to be deserialized into `NullDataResponse` - which fails since `data` is expected to be string:

 `"errorType": "JsonReaderException",
"errorMessage": "Unexpected character encountered while parsing value: {. Path 'data', line 2, position 12.",
"stackTrace": [
"at Newtonsoft.Json.JsonTextReader.ReadStringValue(ReadType readType)",
"at Newtonsoft.Json.JsonTextReader.ReadAsString()",
"at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)",
"at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)",
"at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)",
"at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)",
"at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)",
"at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)",
"at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)",
"at Snowflake.Data.Core.RestRequestImpl.PostAsync[T](SFRestRequest postRequest, CancellationToken cancellationToken)",
"at Snowflake.Data.Core.RestRequestImpl.<>c__DisplayClass11_0'1.<<Post>b__0>d.MoveNext()"
`

howryu:
[...]

PR #201, opened 2020-05-01
Exceptions from CancellationToken callbacks are hard to catch, so don't

tdg5:
The issue I'm running into and proposing a solution for seems like a sequel to https://github.com/snowflakedb/snowflake-connector-net/issues/56 and is described below after my proposed solution.

I propose not throwing exceptions in the `CancellationToken` callbacks because these errors can be hard to catch (I've not been able to catch one :-/) and these exceptions cause the runtime/app to shutdown.

I like this guys perspective of not raising exceptions in `CancellationToken` callbacks: https://bengribaudo.com/blog/2018/02/08/4360/understanding-cancellation-callbacks

Here's also some stack overflow highlighting the difficulty of catching exceptions raised in `CancellationToken` callbacks: https://stackoverflow.com/questions/23618634/how-do-you-catch-cancellationtoken-register-callback-exceptions

Here's the problem I'm trying to solve in issue form in case someone wants to propose an alternative strategy.

### Issue description
The issue I'm running into seems very related to https://github.com/snowflakedb/snowflake-connector-net/issues/56.

Maybe it is something related to the particulars of my async setup, but I'm running into `Identified SQL statement is not currently executing` and/or `Session no longer exists.  New login required to access the service.` exceptions that I'm struggling to `try/catch` that keep killing my app. Having recreated the situation with a strawman not related to snowflake or my app, I'm inclined to think there's a bad pattern going on here.

### Example code
Here's my generic sample of how exceptions raised in CancellationToken callbacks are hard to catch and should be avoided. No matter how I sliced it, I couldn't figure out how to catch an exception when the timeout occurred. This may just be a reflection of my C# naivety.
```cs
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Hosting;
using System.Threading.Tasks;
using System.Threading;
using System;

namespace DukeKaboom
{
    public class Program
    {
        public static void Main(string[] args)
        {
          try {
              Task.Run(() => {
                  try {
                      using var normalCancelSource = new CancellationTokenSource();
                      using var timeoutCancelSource = new CancellationTokenSource(1000);
                      using var linkedCancelSource = CancellationTokenSource.CreateLinkedTokenSource(timeoutCancelSource.Token, normalCancelSource.Token);
                      linkedCancelSource.Token.Register(() => {
                          throw new Exception("boom");
                      }, true);
                      Thread.Sleep(TimeSpan.FromSeconds(3));
                      Console.WriteLine("Leaving the try");
                  }
                  catch (Exception e) {
                    Console.WriteLine("caught it");
                  }
              });
          }
          catch (Exception e) {
              Console.WriteLine("caught a thing");
          }
          Thread.Sleep(TimeSpan.FromSeconds(5));
          Console.WriteLine("yay");
        }
    }
}
```
I wasn't able to get this app to the `yay` or any of the `caught` `WriteLine`s.

### Error log
Depending on the particulars of why the cancellation failed, the errors could look like those below:
```
Session no longer exists.  New login required to access the service.
   at Snowflake.Data.Core.SFStatement.Cancel()
   at Snowflake.Data.Core.SFStatement.<registerQueryCancellationCallback>b__24_0()
   at System.Threading.CancellationToken.<>c.<.cctor>b__26_0(Object obj)
   at System.Threading.CancellationTokenSource.CallbackNode.<>c.<ExecuteCallback>b__9_0(Object s)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) 

Identified SQL statement is not currently executing.                                                                                                                                                                 
   at Snowflake.Data.Core.SFStatement.Cancel()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)    
```
I think there's a hint in the stack traces here that it's going to be hard to catch these exceptions.

### Configuration
*Driver version:* 1.1.2

*Dotnet framework and version:* .net standard 2.0

*Server version:* 4.14.4

*Client OS:* Ubuntu 18.04.4

Please let me know of any questions I can answer or context I can provide.

Thanks in advance!

### DIFF
--- a/Snowflake.Data/Core/SFStatement.cs
+++ b/Snowflake.Data/Core/SFStatement.cs
@@ -147,7 +147,7 @@ namespace Snowflake.Data.Core
                 externalCancellationToken);
             if (!_linkedCancellationTokenSouce.IsCancellationRequested)
             {
-                _linkedCancellationTokenSouce.Token.Register(() => Cancel());
+                _linkedCancellationTokenSouce.Token.Register(() => Cancel(false));
             }
         }
 
@@ -289,24 +289,37 @@ namespace Snowflake.Data.Core
             }
         }
         
-        internal void Cancel()
+        internal void Cancel(bool throwOnFailure = true)
         {
             SFRestRequest request = BuildCancelQueryRequest();
             if (request == null)
                 return;
 
-            var response = _restRequester.Post<NullDataResponse>(request);
+            try {
+                var response = _restRequester.Post<NullDataResponse>(request);
 
-            if (response.success)
-            {
-                logger.Info("Query cancellation succeed");
+                if (response.success)
+                {
+                    logger.Info("Query cancellation succeed");
+                }
+                else
+                {
+                    SnowflakeDbException e = new SnowflakeDbException(
+                        "", response.code, response.message, "");
+                    logger.Error("Query cancellation failed.", e);
+                    if (throwOnFailure)
+                    {
+                        throw e;
+                    }
+                }
             }
-            else
+            catch (Exception e)
             {
-                SnowflakeDbException e = new SnowflakeDbException(
-                    "", response.code, response.message, "");
                 logger.Error("Query cancellation failed.", e);
-                throw e;
+                if (throwOnFailure)
+                {
+                    throw e;
+                }
             }
         }
         


alexrosenfeld10:
@sfc-gh-stakeda tagging you here too for visibility. It would be nice to get this reviewed

alexrosenfeld10:
@tdg5 seems the same as https://github.com/snowflakedb/snowflake-connector-net/pull/276 . I guess your PR was missing a salesforce case 😉

tdg5:
This has been handled in another way.

PR #206, opened 2020-05-11
logic error

sfc-gh-nsfard:
Apologies, there was a logic flaw in the last script!


sfc-gh-stakeda:
This was fixed in #207

PR #207, opened 2020-05-16
updated the wss name

sfc-gh-stakeda:
Updated WSS product and project names based on the agreement with @sfc-gh-jfan


wss.bat
```diff
@@ -1,41 +1,54 @@
...
 java -jar wss-unified-agent.jar -apiKey %WHITESOURCE_API_KEY%^
     -c %WSS_CONFIG%^
-    -project %PROJECT_NAME%^
     -product %PRODUCT_NAME%^
+    -project %PROJECT_NAME%^
     -projectVersion baseline^
     -requestFiles whitesource\update-request.txt^
     -wss.url https://saas.whitesourcesoftware.com/agent
 
 IF %ERRORLEVEL% NEQ 0 (
```
sfc-gh-stakeda, line 42: @sfc-gh-nsfard , I don't think this check works. The above scan with baseline will be success at all times, so the date based versioning won't work. Can you please confirm?

If so, my latest update won't work too, because it always sends the update twice for both baseline and the date version.

Issue #213, opened 2020-06-08
SNOW-165479: Blocking thread on retry handler


pcchannz:
https://github.com/snowflakedb/snowflake-connector-net/blob/3adccf292e8e582a9c56cbbfb039fba7a826807c/Snowflake.Data/Core/HttpUtil.cs#L207

Replace with asynchronous wait:
await Task.Delay(backOffInSec * 1000);

I observed that during heavy load, the waiting time on retry handler could add up.

Prior updating the line, 35% of my application threads were spent on synchronous sleep.
![image](https://user-images.githubusercontent.com/40587526/83988607-d829df00-a997-11ea-8970-e786cf70a426.png)

After updating the line, it was around 8.5%
![image](https://user-images.githubusercontent.com/40587526/83988659-027b9c80-a998-11ea-94ab-768d6d6b7f7c.png)

Thank you!

sfc-gh-jtang:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of April 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link Snowflake Community is the fastest way to get a response.

PR #228, opened 2020-08-05
SNOW-165479 - Use Task.Delay instead of Thread.Sleep in the Async method.

akrock:
Use Task.Delay instead of Thread.Sleep in the Async method.

Addresses #213 SNOW-165479: Blocking thread on retry handler


extremeandy:
LGTM 👍

icambron:
This really needs to be merged. You can't block async threads like this.

extremeandy:
It's because of things like this that we end up just maintaining our own fork 🤷‍♂️

akrock:
@extremeandy I know right, we have never been on the official snowflake build over the last 3-4 years, but recently It's gotten so bad I have literally thought about making our fork an "unofficial" one and actually putting it up in Nugets.

This is the branch we are actively running in production (that includes this fix and some others).
https://github.com/akrock/snowflake-connector-net/tree/pc-master-2020
NOTE: I haven't synced it up in about a year, but it is stable for our usage (probably about 1million daily queries). 

I will look to picking up some more of the active PRs and other changes when I can work it into a sprint in the next few weeks. If there is community interest in an "unofficial" fork at that point I might be able to invest a little extra time to get  CI & a package setup.

extremeandy:
I've considered switching over to using this package but I'm not sure how mature it is, and the lack of support for chunks could be an issue:

https://github.com/fixer-m/snowflake-db-net-client/issues

akrock:
Nice I wasn't aware of that one, I will keep an eye on it. Unfortunately not having support for chunks will be a deal breaker us.

icambron:
We also keep a fork and sync it up every once in a while. It's frustrating. At some point one of us will have to publish or we're all doing the same work independently

stegus64:
@SimbaGithub This seems like something you should really merge as soon as possible.

SimbaGithub:
@akrock Do you mind resolving conflicts with master so that I can approve this and merge?
Otherwise, I can just create new PR if that's easier for you.

akrock:
@SimbaGithub it's updated

SimbaGithub:
Awesome, thank you @akrock! 
Will merge right now.

SimbaGithub (approved):

PR #239, opened 2020-09-24
Updated third parties with CVEs and removed unused packages.config files

SimbaGithub:
Third parties are updated to account for the CVEs. And the unused "packages.config" files are removed since it caused some CVEs to show up even though the packages listed in those files are not installed or used by the project


Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -19,7 +19,8 @@
   <ItemGroup>
     <PackageReference Include="AngleSharp" Version="0.12.1" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <PackageReference Include="log4net" Version="2.0.8" />
+    <PackageReference Include="log4net" Version="2.0.11" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
sfc-gh-hyu, line 23: Just curious, why this is package reference is needed?

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -29,6 +30,12 @@
     <Reference Include="System.Web" />
   </ItemGroup>
   
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
+    <PackageReference Include="Microsoft.PowerShell.Commands.Utility">
```
sfc-gh-hyu, line 34: What is powershell commands utitlity used for?

Snowflake.Data/Snowflake.Data.csproj
SimbaGithub: Hi, System.Text.RegularExpressions is a dependency of log4net
And it has a vulnerability which version 4.3.1 fixes: https://www.whitesourcesoftware.com/vulnerability-database/CVE-2019-0820

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
SimbaGithub: Powershell is not actually used by the driver. But .NET Core 2.0 has a vulnerability which upgrading Powershell to 6.1.0 is part of the fix: https://www.whitesourcesoftware.com/vulnerability-database/CVE-2018-8292

sfc-gh-hyu (approved):
Thanks for the explanation. LGTM.

PR #243, opened 2020-10-21
Plug and play

SimbaGithub:
Documentation for plug-and-play logging design: 
[Plug-&-Play Logging Design.pdf](https://github.com/snowflakedb/snowflake-connector-net/files/5412946/Plug-.-Play.Logging.Design.pdf)

Example of plug-and-play using Serilog:
[PlugAndPlayEx.txt](https://github.com/snowflakedb/snowflake-connector-net/files/5412947/PlugAndPlayEx.txt)

### DIFF
--- a/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
+++ b/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
@@ -95,8 +95,7 @@ namespace Snowflake.Data.Core
 
         public Task<IResultChunk> GetNextChunkAsync()
         {
-            logger.InfoFmt("NextChunkToConsume: {0}, NextChunkToDownload: {1}",
-                nextChunkToConsumeIndex, nextChunkToDownloadIndex);
+            logger.Info($"NextChunkToConsume: {nextChunkToConsumeIndex}, NextChunkToDownload: {nextChunkToDownloadIndex}");
             if (nextChunkToConsumeIndex < chunkInfos.Count)
             {
                 Task<IResultChunk> chunk = taskQueues[nextChunkToConsumeIndex % prefetchSlot];
@@ -147,7 +146,7 @@ namespace Snowflake.Data.Core
             {
                 ParseStreamIntoChunk(stream, chunk);
             }
-            logger.InfoFmt("Succeed downloading chunk #{0}", chunk.chunkIndexToDownload);
+            logger.Info($"Succeed downloading chunk #{chunk.chunkIndexToDownload}");
             return chunk;
         }
 
--- a/Snowflake.Data/Logger/SFLogger.cs
+++ b/Snowflake.Data/Logger/SFLogger.cs
@@ -20,25 +20,15 @@ namespace Snowflake.Data.Log
 
         bool IsFatalEnabled();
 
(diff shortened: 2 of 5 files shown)


Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
```diff
@@ -95,8 +95,7 @@ private int GetPrefetchThreads(SFBaseResultSet resultSet)
 
         public Task<IResultChunk> GetNextChunkAsync()
         {
-            logger.InfoFmt("NextChunkToConsume: {0}, NextChunkToDownload: {1}",
-                nextChunkToConsumeIndex, nextChunkToDownloadIndex);
+            logger.Info($"NextChunkToConsume: {nextChunkToConsumeIndex}, NextChunkToDownload: {nextChunkToDownloadIndex}");
```
SimbaGithub, line 98: Replaced InfoFmt() with equivalent Info() code since InfoFmt() and other Fmt() codes are removed to simplify the interface

Snowflake.Data/Logger/SFLogger.cs
```diff
@@ -20,29 +20,19 @@ interface SFLogger
...
-
-        void FatalFmt(string fmt, params object[] args);
     }
 
     enum LoggingEvent
     {
         DEBUG, INFO, WARN, ERROR, FATAL
     }
-}
+}
```
sfc-gh-hyu, line 38: new line here.

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -8,10 +8,52 @@ namespace Snowflake.Data.Log
...
     {
+        private static bool isLoggerEnabled = true;
+
+        private static SFLogger logger = null;
+
+        private SFLoggerFactory()
+        {
+        }
+
+        public static void disableLogger()
```
sfc-gh-hyu, line 19: I am not an expert on .net scoping. Have you verified that if this method is accessible from a different project?

sfc-gh-hyu:
Just to be clear, this method still requires log4net as a dependency right? If the customer is using a different version of log4net, there will still be a conflict?

Snowflake.Data/Logger/SFLoggerFactory.cs
SimbaGithub: Yes, confirmed that the static methods are accessible when called from a different project.
The attached "PlugAndPlayEx" file can be placed in the Snowflake.Data.Tests project for testing as well

Snowflake.Data/Logger/SFLogger.cs
SimbaGithub: Will add newline

SimbaGithub:
> Just to be clear, this method still requires log4net as a dependency right? If the customer is using a different version of log4net, there will still be a conflict?

Yes, log4net is still a dependency. But with this change, the customer can implement the SFLogger interface with their version of log4net to remove the conflict. I tested this using the oldest version of log4net (1.2.10) and latest (2.0.11)

sfc-gh-tchen (approved):
LGTM

sfc-gh-sshankar (approved):
Approval

PR #250, opened 2020-10-28
SNOW 213810 Update parameters.json for logging into aws and azure

sfc-gh-hchaturvedi:
This PR updates the parameters.json encrypted file for aws and azure. The goal is to be able update the connection params so that the connector  can login to Snowflake successfully. This is to prevent tests from failing due to schema not found errors.


sfc-gh-igarish:
@sfc-gh-hchaturvedi do we need this PR? Looks very old. Could you please close, if it's no more required?

PR #258, opened 2020-12-15
Use AuthenticatorFactory instead of returning BasicAuthenticator by d…

SimbaGithub:
…efault


SimbaGithub:
Integrated as part of the jwt and oauth authenticators work.

PR #246, opened 2020-10-23
Get rid of CurrentThread.CurrentCulture dependency

BillSytin:
Small changes to address an issue
https://github.com/snowflakedb/snowflake-connector-net/issues/219

Issue #219, opened 2020-06-25
Snowflake driver changes current culture to Invariant


BillSytin:
### Issue description
Snowflake driver changes current culture to Invariant.
Applications often depend on the current thread culture, and it could be harmful and unexpected.

### Example code
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/SFDataConverter.cs
```cs
internal SFDataConverter()
        {
            Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
        }
```

clegendre:
What about the status of this issue and the respective PR ?
This is blocking us for deploying our application.

sfc-gh-igarish:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of April 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

PR #264, opened 2021-01-31
SNOW-344502 - Improved performance when downloading large amounts of data.

stegus64:
* Uses handwritten parsers for integers and decimals
* Keeps data in UTF8 format until we absolutely need to convert it to strings

In a testcase that downloads 266 million rows we had the following runtime before this change:

.Net Framework 00:13:16
Core           00:05:16

After this change the same testcase has the following runtime:

.Net Framework 00:02:18
Core           00:02:29

### DIFF
--- a/Snowflake.Data.Tests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/SFDataConverterTest.cs
@@ -51,11 +51,10 @@ namespace Snowflake.Data.Tests
                 inputTime = DateTime.ParseExact(inputTimeStr, "yyyy-MM-dd HH:mm:ss.fffffff", CultureInfo.InvariantCulture);
             }
 
-            var dataConverter = new SFDataConverter();
             var unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
             var tickDiff = inputTime.Ticks - unixEpoch.Ticks;
             var inputStringAsItWasFromDatabase = (tickDiff / 10000000.0m).ToString(CultureInfo.InvariantCulture);
-            var result = dataConverter.ConvertToCSharpVal(inputStringAsItWasFromDatabase, SFDataType.TIMESTAMP_NTZ, typeof(DateTime));
+            var result = SFDataConverter.ConvertToCSharpVal(inputStringAsItWasFromDatabase, SFDataType.TIMESTAMP_NTZ, typeof(DateTime));
             Assert.AreEqual(inputTime, result);
         }
 
@@ -97,5 +96,29 @@ namespace Snowflake.Data.Tests
             Assert.AreEqual(dtExpected, dtResult);
         }
 
+        [Test]
+        [TestCase("9223372036854775807")]
+        [TestCase("-9223372036854775808")]
+        [TestCase("-1")]
+        [TestCase("999999999999999999")]
+        public void TestConvertToInt64(string s)
+        {
+            Int64 actual = (Int64)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int64));
+            Int64 expected = Convert.ToInt64(s);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        [TestCase("9223372036854775807.9223372036854775807")]
+        [TestCase("-9223372036854775807.1234567890")]
+        [TestCase("-1.300")]
+        [TestCase("999999999999999999.000000000000100000000000")]
+        [TestCase("4294967295.4294967296")]
+        public void TestConvertToDecimal(string s)
+        {
+            decimal actual = (decimal)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(decimal));
+            decimal expected = Convert.ToDecimal(s, System.Globalization.CultureInfo.InvariantCulture);
+            Assert.AreEqual(expected, actual);
+        }
     }
 }
--- a/Snowflake.Data.Tests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/SFDataConverterTest.cs
@@ -117,7 +117,8 @@ namespace Snowflake.Data.Tests
         public void TestConvertToDecimal(string s)
         {
             decimal actual = (decimal)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(decimal));
-            decimal expected = Convert.ToDecimal(s, System.Globalization.CultureInfo.InvariantCulture);
+            decimal expected = Convert.ToDecimal(s, CultureInfo.InvariantCulture);
+
             Assert.AreEqual(expected, actual);
         }
     }
--- a/Snowflake.Data.Tests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/SFDataConverterTest.cs
@@ -109,6 +109,72 @@ namespace Snowflake.Data.Tests
         }
 
         [Test]
+        [TestCase("2147483647")]
+        [TestCase("-2147483648")]
+        [TestCase("-1")]
+        [TestCase("0")]
+        public void TestConvertToInt32(string s)
+        {
+            Int32 actual = (Int32)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int32));
+            Int32 expected = Convert.ToInt32(s);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        [TestCase("32767")]
+        [TestCase("-32768")]
+        [TestCase("-1")]
+        [TestCase("0")]
+        public void TestConvertToInt16(string s)
+        {
+            Int16 actual = (Int16)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int16));
+            Int16 expected = Convert.ToInt16(s);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        [TestCase("255")]
+        [TestCase("0")]
+        public void TestConvertToByte(string s)
+        {
+            byte actual = (byte)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(byte));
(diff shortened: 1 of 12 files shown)


stegus64:
@sfc-gh-abhatnagar Have you seen this PR? 

I am not sure about the proper process to suggest improvements, but this PR will improve performance in certain important situations by a factor of 5 or more.

Are you not interested?

SimbaGithub:
This will definitely improve conversion of numeric and date-time data!

Snowflake.Data/Core/FastParser.cs
```diff
@@ -0,0 +1,115 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Snowflake.Data.Core
+{
+    public class FastParser
+    {
+
+        public static Int64 FastParseInt64(byte[] s, int offset, int len)
```
SimbaGithub, line 10: What about adding _private static Int64 FastParseInt(byte[] s, int offset, int len, UInt64 maxValue)_, with the same code/logic than _FastParseInt64_ and _FastParseInt32_ and have them call FastParseInt with the correct max value? FastParseInt16 and FastParseByte could also be added easily and catch overflow ahead of FastParseInt32.

Snowflake.Data.Tests/SFDataConverterTest.cs
```diff
@@ -97,5 +96,29 @@ private void internalTestConvertDate(DateTime dtExpected, DateTime testValue)
...
+        [Test]
+        [TestCase("9223372036854775807.9223372036854775807")]
+        [TestCase("-9223372036854775807.1234567890")]
+        [TestCase("-1.300")]
+        [TestCase("999999999999999999.000000000000100000000000")]
+        [TestCase("4294967295.4294967296")]
+        public void TestConvertToDecimal(string s)
+        {
+            decimal actual = (decimal)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(decimal));
+            decimal expected = Convert.ToDecimal(s, System.Globalization.CultureInfo.InvariantCulture);
```
SimbaGithub, line 120:
```suggestion
            decimal expected = Convert.ToDecimal(s, CultureInfo.InvariantCulture);

```

Snowflake.Data/Core/FastParser.cs
stegus64: There is a good reason for having separate implementations of FastParseInt32 and FastParseInt64. 
FastParseInt32 is faster because it only works with 32-bit integers. When the end user calls DataReader.GetInt(i) we know he only expects a 32-bit value so we can call the faster version FastParseInt32 directly.
I don't see a good reason to define a FastParseInt method.

stegus64:
This PR also solves #273  #246  #219

Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -20,74 +20,100 @@ public enum SFDataType
...
+                    return srcVal.Buffer[0] == '1' || srcVal.Buffer[0] == 't' || srcVal.Buffer[0] == 'T';
+                }
+                else if (destType == typeof(byte[]))
+                {
+                    return srcType == SFDataType.BINARY ?
+                        HexToBytes(srcVal.ToString()) : srcVal.GetBytes();
+                }
+                else if (destType == typeof(Int16))
+                {
+                    return (Int16)FastParser.FastParseInt32(srcVal.Buffer, srcVal.offset, srcVal.length);
```
SimbaGithub, line 82: This will not catch overflows and return erroneous data to the user instead of an error.

Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -20,74 +20,100 @@ public enum SFDataType
...
+                }
+                else if (destType == typeof(byte))
+                {
+                    return (byte)FastParser.FastParseInt32(srcVal.Buffer, srcVal.offset, srcVal.length);
```
SimbaGithub, line 86: Here as well, it will also not catch overflows

Snowflake.Data/Core/FastParser.cs
SimbaGithub: I am not sure how much this would affect the performance, but if this is the reason behind those almost 2 identical functions then it's fine.

stegus64: image

This is a benchmark of different parsing methods. As you can see FastParseInt32 is about 30-40% faster than FastParseInt64. I think it is well worth the effort to enable that speedup.

TypeConverter was the old code before this PR. As you can see, the old code was 57 times slower than the new version on .Net Framework

Snowflake.Data/Core/SFDataConverter.cs
stegus64: Good catch! I thought a narrowing cast generated overflow, but you are correct.
Fixed

Snowflake.Data/Core/SFDataConverter.cs
stegus64: Fixed

SimbaGithub:
LGTM :)

@sfc-gh-hyu  : Can you take a look and see if you are Ok to green light this as well?

sfc-gh-hyu (commented):
@SimbaGithub I understands the big picture but did not look into the detail. I think the change make sense to me. My only concern is that do we want to add some kind of parameter/config to fall back in case of we found some bugs in the new code?

SimbaGithub:
@sfc-gh-hyu : We can always revert if needed, I think it's probably better than leaving both alive in the connector. Now I agree that this touches data and is crucial to the connector. I would expect that if all the tests pass then it should be Ok, but I don't have an extensive knowledge of the existing tests. Do you know of missing edge case tests which need to be added? Do you have something specific in mind?

I'll take another look for edge cases as well.

SimbaGithub:
Here is the plan : When CI is back to normal, let's review the test cases, add more if missing ones are identified and then we should be good to go with integrating this in next month release.

stegus64:
Sounds good

sfc-gh-hyu (approved):

Snowflake.Data.Tests/SFDataConverterTest.cs
```diff
@@ -97,5 +96,96 @@ private void internalTestConvertDate(DateTime dtExpected, DateTime testValue)
...
+        [TestCase("-1.300")]
+        [TestCase("999999999999999999.000000000000100000000000")]
+        [TestCase("4294967295.4294967296")]
+        public void TestConvertToDecimal(string s)
+        {
+            decimal actual = (decimal)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(decimal));
+            decimal expected = Convert.ToDecimal(s, CultureInfo.InvariantCulture);
+
+            Assert.AreEqual(expected, actual);
+        }
```
SimbaGithub, line 189: Can we had a few extra tests unsupported conversions and a couple of special cases?
```     }

        [Test]
        [TestCase("9223372036854775807.9223372036854775807")]
        [TestCase("-9223372036854775807.1234567890")]
        [TestCase("-1.300")]
        [TestCase("999999999999999999.000000000000100000000000")]
        [TestCase("4294967295.4294967296")]
        [TestCase("1.5e-36")]
        [TestCase("1.5e+38")]
        //[TestCase("inf")] -- TODO - Not supported yet
        //[TestCase("-inf")] -- TODO - Not supported yet
        [TestCase("NaN")]
        public void TestConvertToFloat(string s)
        {
            double actualDouble = (double)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(double));
            double expectedDoulbe = Convert.ToDouble(s, CultureInfo.InvariantCulture);

            Assert.AreEqual(actualDouble, expectedDoulbe);

            float actualFloat = (float)SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(float));
            float expectedFloat = Convert.ToSingle(s, CultureInfo.InvariantCulture);

            Assert.AreEqual(expectedFloat, actualFloat);
        }

        [Test]
        [TestCase("thisIsNotAValidValue")]
        [TestCase("-1.300")]
        [TestCase("425.426")]
        [TestCase("1.5e-36")]
        [TestCase("1.5e+38")]
        [TestCase("inf")]
        [TestCase("-inf")]
        [TestCase("NaN")]
        public void TestInvalidConversionInvalidInt(string s)
        {
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int32)));
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int64)));
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(Int16)));
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(byte)));
        }

        [Test]
        [TestCase("thisIsNotAValidValue")]
        public void TestInvalidConversionInvalidFloat(string s)
        {
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(float)));
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(double)));
        }

        [Test]
        [TestCase("thisIsNotAValidValue")]
        [TestCase("1.5e-36")]
        [TestCase("1.5e+38")]
        [TestCase("inf")]
        [TestCase("-inf")]
        [TestCase("NaN")]
        public void TestInvalidConversionInvalidDecimal(string s)
        {
            Assert.Throws<FormatException>(() => SFDataConverter.ConvertToCSharpVal(s, SFDataType.FIXED, typeof(decimal)));
        }

```

stegus64: Added these tests to the PR

SimbaGithub:
Thank you Stefan for the work on this! Going to merge it now. It will be in the next release.

PR #276, opened 2021-03-11
Log query cancellation failure as a warning

SimbaGithub:
Calling cancel on a query should not generate an exception when it fails: https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.cancel?view=net-5.0

This fixes the issue by logging the failed cancellation as a warning instead of throwing an exception.
Note: a different exception is still thrown since the command failed to execute. And is bubbled up to SnowflakeDbCommand for visibility


SimbaGithub:
@howryu This is related to salesforce case: 00297660. Please review when available, thanks.

sfc-gh-hyu (approved):

Issue #233, opened 2020-09-19
SNOW-197712: Why are we setting global settings inside each httpclient creation?  


mhunting:
https://github.com/snowflakedb/snowflake-connector-net/blob/21f2d01a2add057ed5d19130d9a485b7289f5ff8/Snowflake.Data/Core/HttpUtil.cs#L43

I believe this is a global setting and affects more than just your library behavior.
https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.defaultconnectionlimit?view=netcore-3.1

shawnjones253:
same with the other `ServicePointManager` stuff https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/HttpUtil.cs#L59

sfc-gh-igarish:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of April 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

PR #277, opened 2021-03-12
Avoid setting ServicePoint options that are global to the entire program

stegus64:
Move the option setting to RetryHandler.SendAsync() so they only have a local effect

Also adds p.Expect100Continue = false which avoids about 100 ms overhead for each web request

Note that the these settings have been removed since they can only be set globally:

            // ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
            // ServicePointManager.CheckCertificateRevocationList = true;

The default value for these works fine and if an end user needs to make a change he can use a config file

Solves issue #233

@SimbaGithub @sfc-gh-hyu  Please review

### DIFF
--- a/Snowflake.Data/Core/HttpUtil.cs
+++ b/Snowflake.Data/Core/HttpUtil.cs
@@ -55,13 +55,14 @@ namespace Snowflake.Data.Core
 
         static private void initHttpClient()
         {
-            // enforce tls v1.2
-            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
-            ServicePointManager.UseNagleAlgorithm = false;
-            ServicePointManager.CheckCertificateRevocationList = true;
-
-            // Control how many simultaneous connections to each host are allowed from this client
-            ServicePointManager.DefaultConnectionLimit = 20;
+            // Note that these options can only be set globally
+            // We do not want to do this. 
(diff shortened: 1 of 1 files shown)


sfc-gh-hyu:
@stegus64 Did you use .net framework or .net core? If .net framework, what's the version are you using? The reason I am asking is that we are trying to enforce tls version and certificationRevocationList check in a local settings. So far we have found following: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.sslprotocols?view=net-5.0 and https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.checkcertificaterevocationlist?view=net-5.0
However, both settings required .net core or .net framework 4.7.2 +

stegus64:
I have been using both .net framework and .net core. You are right that the functionality to change SSL and CRL locally is only available in very recent versions. That is why I suggest to remove these settings from the library.
I think that for most users the default values will be sufficient, and if not these settings can be changed for a certain application by the end user with a config file.

In any case, in my opinion it is not a good idea for a library to try to control settings that are global to an entire program.

By changing these settings globally like in the original code you make it impossible for end users or application developers to control these options at all. There might be valid reasons (performance) to avoid the CRL check for example, especially since these options affects all request by the entire program - not only to snowflake.

stegus64:
Fixed by #298. Closing this PR

PR #280, opened 2021-03-17
Fix incorrect instructions in README

benjanderson:
Account and region is required when creating a connection string
account=testaccount;user=testuser;password=XXXXX;db=testdb;schema=testschema


sfc-gh-igarish:
I am not finding associated issue. Please create an issue and then raise the PR.

Issue #285, opened 2021-03-26
SNOW-310449: SFDataConverter.csharpValToSfVal throws Exception for everything but SFDataType.TIMESTAMP_LTZ


gojimmypi:
### Issue description
When [binding parameters](https://github.com/snowflakedb/snowflake-connector-net/blob/6bad661afcc9c1d9bcbd2a7ec0a1096963b53aca/Snowflake.Data/Client/SnowflakeDbCommand.cs#L230) via `convertToBindList` in a `SnowflakeDbCommand`, types other than `SFDataType.TIMESTAMP_LTZ` should be allowed.

### Example code
See [this line](https://github.com/snowflakedb/snowflake-connector-net/blob/6bad661afcc9c1d9bcbd2a7ec0a1096963b53aca/Snowflake.Data/Client/SnowflakeDbCommand.cs#L283). A call to `SFDataConverter.csharpValToSfVal` with a `parameter.Value` set to anything other than  `SFDataType.TIMESTAMP_LTZ`  will [throw an exception](https://github.com/snowflakedb/snowflake-connector-net/blob/6bad661afcc9c1d9bcbd2a7ec0a1096963b53aca/Snowflake.Data/Core/SFDataConverter.cs#L316).

### Error log
There's a [not implemented exception thrown](https://github.com/snowflakedb/snowflake-connector-net/blob/6bad661afcc9c1d9bcbd2a7ec0a1096963b53aca/Snowflake.Data/Core/SFDataConverter.cs#L316) for any types other than `SFDataType.TIMESTAMP_LTZ`

### Configuration
Client 1.2.0

*Server version:* 
N/A

*Client OS:* 
Windows 10, Visual Studio 2019 Enterprise

sfc-gh-igarish:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of April 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

PR #286, opened 2021-03-26
csharpValToSfVal should allow at least SFDataType.TEXT

gojimmypi:
All values other than `SFDataType.TIMESTAMP_LTZ` thow an exception,

See https://github.com/snowflakedb/snowflake-connector-net/issues/285


sfc-gh-pbulawa:
Thank you for your contribution. The issue has been resolved within #315.

PR #290, opened 2021-04-08
Performance investigation

SimbaGithub:


Snowflake.Data/Core/SFResultSet.cs
```diff
@@ -99,23 +154,64 @@ internal override bool Next()
...
+            stopwatch.Stop();
+            Logger.Debug("All data retrieved. Last row returned.");
+            Logger.Debug($"Fetch process took {stopwatch.Elapsed.TotalMilliseconds} ms");
+            if (Logger.IsDebugEnabled())
+            {
+                if (timings.TryGetValue("nextChunk", out double nextChunkTotalTime))
+                {
+                    Logger.Debug("---- Time spent getting next chunk ----");
+                    Logger.Debug($"Total time : {nextChunkTotalTime} ms");
+                    Logger.Debug($"Average time : {nextChunkTotalTime / (float)_totalChunkCount } ms\n");
```
howryu, line 209: Can you change those logging to debug level? This way potentially we can get those data points only if we enabled the info level logging, which potentially have a much less log volume.

Snowflake.Data/Core/SFBaseResultSet.cs
```diff
@@ -43,28 +117,88 @@ internal string GetString(int columnIndex)
...
+                        return getObjectInternal(columnIndex);
+                    });
             }
         }
 
-        internal object GetValue(int columnIndex)
+        internal object GetValue(int columnIndex, out double time)
         {
-            string val = getObjectInternal(columnIndex);
+            string val = Measure(@"getObjectInternal", out double getObjectInternalTime, () =>
```
howryu, line 147: Do you know the overhead of this Measure function?

howryu (commented):
LGTM in general. However, I do have a concern on how much overhead this will bring in the normal case,  i.e. those customers who did not want to enabled the measurement. Do you think if it is worth to have two version of the method. The default function call still goes to the old method. If the function will only be measured if some kind of config indicates that the function should be measured.

Snowflake.Data/Core/SFResultSet.cs
SimbaGithub: The logging is already in Debug level here. The ones at info level are not printing much, only a line at the beginning of the read, another one every 100000 rows and a last one when retrieval is done.

SimbaGithub: The logging is already in Debug level here. The ones at info level are not printing much, only a line at the beginning of the read, another one every 100000 rows and a last one when retrieval is done.

The timing processing is done if log level is Debug only.

This is a sample for select * from SNOWFLAKE_SAMPLE_DATA.TPCH_SF100.LINEITEM limit 350000:
- ERROR - No log : Fetch time = 9157
- INFO - Will log start/end fetch, every 100000 rows and total fetch time : Fetch time = 9797
- DEBUG - Same as INFO plus timing for the methods GetValue, ConvertToCSharpVal, getObjectInternal and GetNextChunkAsync : Fetch time = 15407

Snowflake.Data/Core/SFBaseResultSet.cs
SimbaGithub: It is quite big. The run time increased from 10 to 15sec in the sample query "select * from SNOWFLAKE_SAMPLE_DATA.TPCH_SF100.LINEITEM limit 350000".

stegus64: This seems like a very bad idea!

SFBaseResultSet.GetValue() is a very performance-sensitive function. It can be called millions of times. You should absolutely not add any extra overhead to this code path.

I have used the following code to test:

```
            using (IDbCommand command = conn.CreateCommand())
            {
                command.CommandText = "select * from SNOWFLAKE_SAMPLE_DATA.TPCH_SF100.LINEITEM limit 1500000";
                var dr = command.ExecuteReader();
                sw = System.Diagnostics.Stopwatch.StartNew();
                while (dr.Read())
                {
                    rowcount++;
                    for (int i = 0; i < dr.FieldCount; i++)
                    {
                        dr.GetValue(i);
                    }
                }
                sw.Stop();
                Console.WriteLine("rowcount = {0} Time = {1}", rowcount, sw.Elapsed);
            }

```
I have tested this from a console program running NetFramework with all logging turned OFF

After this change: 42.1 s
Before this change: 18.0 s

This change makes the code more than twice as slow in an important usecase even without turning on logging at all.

If we do the same test with logLevel=DEBUG we get 61 s instead.

Please do NOT merge this PR.

SimbaGithub: Hi Stefan, 
Don't worry, we are not going to ever merge this!
We will be using this to collect more detailed metric on the different types conversions and only timing the retrieval execution as you are doing here does not help with this. Yes, when data conversion is improved it shows up in the overall time, but it doesn't highlight if a specific conversion (JSON to C#) need the most improvement or if the raw data transfer is a bottle neck.
I know you worked on improving the conversion to numeric types and this is pretty awesome, but there is most likely other things that can be improved.

stegus64: Good to hear that you are not planning to merge this!

I am not really sure what you are trying to accomplish here, but this kind of micro-benchmarking can be very misleading. 

You try to measure each individual function call and then sum all the measurements. This does not work properly for these very fast functions.

On my computer, Stopwatch has a resolution of 100 ns - each tick is 100ns

dr.GetValue(0) for an int64 column (with my improvements) takes about 80 ns (measured by calling it 10000000 times in a tight loop).

It is impossible to measure something that only takes 80ns with a clock that has 100ns resolution. You are only measuring your own overhead.

If you want to measure the performance of GetValue() for different kinds of data, I recommend that you call GetValue() in a tight loop without calling dr.Next().

That way you are not loading any data from snowflake while measuring performance.

You might also want to have a look at this: https://github.com/dotnet/BenchmarkDotNet

SimbaGithub: Thanks Stefan! BenchmarkDotNet seems very good for benchmarking specific methods like the conversion method. I didn't see how it could be applied into gathering metrics from a customer application, but it is definitely useful.

PR #295, opened 2021-04-19
Rotated credentials for dotnet

sfc-gh-hchaturvedi:
Credentials rotation for dotnet in response to code cov leak


sfc-gh-mkeller (approved):
🚢 🚢

PR #202, opened 2020-05-06
Prevent SnowflakeDbConnection.AsyncOpen from blocking Close request

tdg5:
Similar to the code that already exists in the synchronous version of `Open`, add some error handling to make sure the DbConnection doesn't end up in a weird state where it will block forever when `Close`  or `Dispose` is called.

PR #296, opened 2021-04-20
Catch exceptions in SnowflakeDbConnection.Dispose

dnawnOlo:
Ultimately the issue I was running into was a TaskCanceledException that was being thrown within `SnowflakeDbConnection.Close()`. When this was called within `SnowflakeDbConnection.Dispose()` and that exception was thrown, it would crash the process. I'm just looking to wrap `this.Close()` within `SnowflakeDbConnection.Dispose()` in a try/catch to prevent an Exception being thrown within Dispose.

**Original Comment**

Any exceptions that occur in the method being called in `Task.Run()` end up as unhandled exceptions that cause the process to terminate. 

The case I ran into was that we had a network issue that prevented us from connecting to Snowflake. The `SnowflakeDbConnection.OpenAsync()` call set the ConnectionState to `ConnectionState.Open`. The fact it sets the ConnectionState to Open when the call failed would be addressed by https://github.com/snowflakedb/snowflake-connector-net/pull/202. 

When SnowflakeDbConnection is disposed of and `SfSession.close()` is called, ultimately `RestRequester.Post()` is called and the PostAsync call made in the Task.Run throws an AggregateException and crashes our w3wp process. These changes are to prevent those unhandled exceptions scenarios, log the exceptions and then throw up to the caller of the method that can further catch those exceptions and respond appropriately.

### DIFF
--- a/Snowflake.Data/Core/RestRequester.cs
+++ b/Snowflake.Data/Core/RestRequester.cs
@@ -48,7 +48,19 @@ namespace Snowflake.Data.Core
         public T Post<T>(IRestRequest request)
         {
             //Run synchronous in a new thread-pool task.
-            return Task.Run(async () => await PostAsync<T>(request, CancellationToken.None)).Result;
+            var task = Task.Run(async () => await PostAsync<T>(request, CancellationToken.None));
+            try
+            {
+                return task.Result;
+            }
+            catch (AggregateException ae)
+            {
+                foreach (var ex in ae.InnerExceptions)
+                {
+                    logger.Error("Rest request failure", ex);
+                }
+                throw ae.GetBaseException();
+            }
         }
 
         public async Task<T> PostAsync<T>(IRestRequest request, CancellationToken cancellationToken)
@@ -63,7 +75,19 @@ namespace Snowflake.Data.Core
         public T Get<T>(IRestRequest request)
         {
             //Run synchronous in a new thread-pool task.
-            return Task.Run(async () => await GetAsync<T>(request, CancellationToken.None)).Result;
+            var task = Task.Run(async () => await GetAsync<T>(request, CancellationToken.None));
+            try
+            {
+                return task.Result;
+            }
+            catch (AggregateException ae)
+            {
+                foreach (var ex in ae.InnerExceptions)
+                {
+                    logger.Error("Rest request failure", ex);
+                }
+                throw ae.GetBaseException();
+            }
         }
 
         public async Task<T> GetAsync<T>(IRestRequest request, CancellationToken cancellationToken)
@@ -85,7 +109,19 @@ namespace Snowflake.Data.Core
             HttpRequestMessage message = request.ToRequestMessage(HttpMethod.Get);
 
             //Run synchronous in a new thread-pool task.
-            return Task.Run(async () => await GetAsync(request, CancellationToken.None)).Result;
+            var task = Task.Run(async () => await GetAsync(request, CancellationToken.None));
+            try
+            {
+                return task.Result;
+            }
+            catch (AggregateException ae)
+            {
+                foreach (var ex in ae.InnerExceptions)
+                {
+                    logger.Error("Rest request failure", ex);
+                }
+                throw ae.GetBaseException();
+            }
         }
         
         private async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, 
--- a/Snowflake.Data/Core/RestRequester.cs
+++ b/Snowflake.Data/Core/RestRequester.cs
@@ -48,19 +48,7 @@ namespace Snowflake.Data.Core
         public T Post<T>(IRestRequest request)
         {
             //Run synchronous in a new thread-pool task.
-            var task = Task.Run(async () => await PostAsync<T>(request, CancellationToken.None));
-            try
-            {
-                return task.Result;
-            }
-            catch (AggregateException ae)
-            {
-                foreach (var ex in ae.InnerExceptions)
-                {
-                    logger.Error("Rest request failure", ex);
-                }
-                throw ae.GetBaseException();
-            }
+            return Task.Run(async () => await PostAsync<T>(request, CancellationToken.None)).Result;
         }
 
         public async Task<T> PostAsync<T>(IRestRequest request, CancellationToken cancellationToken)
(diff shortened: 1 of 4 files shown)


Snowflake.Data/Core/RestRequester.cs
```diff
@@ -48,7 +48,19 @@ static internal RestRequester Instance
         public T Post<T>(IRestRequest request)
         {
             //Run synchronous in a new thread-pool task.
-            return Task.Run(async () => await PostAsync<T>(request, CancellationToken.None)).Result;
+            var task = Task.Run(async () => await PostAsync<T>(request, CancellationToken.None));
```
akrock, line 51: These changes mainly just add logging around GetAsync/PostAsync exceptions.  If the goal was enhanced logging during failures it might be more explictly to simply add the logging in these methods (which would also benefit the ASYNC code paths -- not just the sync ones).

Snowflake.Data/Core/RestRequester.cs
```diff
@@ -48,7 +48,19 @@ static internal RestRequester Instance
...
+            {
+                return task.Result;
+            }
+            catch (AggregateException ae)
+            {
+                foreach (var ex in ae.InnerExceptions)
+                {
+                    logger.Error("Rest request failure", ex);
+                }
+                throw ae.GetBaseException();
```
akrock, line 62: I would have to write a test case, but I'm pretty sure `ae.GetBaseException()` is likely to just be the `ae` instance, in which case this doesn't give us any benefit over simply just using `throw` to rethrow the exception and it will clobber the stack trace on the way out...

Snowflake.Data/Core/SFSession.cs
```diff
@@ -160,10 +160,17 @@ internal void close()
...
-            if (!response.success)
+            try
+            {
+                var response = restRequester.Post<CloseResponse>(closeSessionRequest);
+                if (!response.success)
+                {
+                    logger.Debug($"Failed to delete session, error ignored. Code: {response.code} Message: {response.message}");
+                }
+            }
+            catch (Exception ex)
```
akrock, line 171: Swallowing errors on `Close()` could be considered a breaking change if anyone was relying on this behavior.

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -295,15 +295,21 @@ internal void Cancel()
...
+                if (response.success)
+                {
+                    logger.Info("Query cancellation succeed");
+                }
+                else
+                {
+                    logger.Warn("Query cancellation failed.");
+                }
+            }            
+            catch (Exception ex)
```
akrock, line 310: Swallowing errors on `Cancel()` could be considered a breaking change if anyone was relying on this behavior.

akrock:
> Any exceptions that occur in the method being called in Task.Run() end up as unhandled exceptions that cause the process to terminate.

Have you tested this assertion in an isolated environment?

:one: Exceptions thrown during a Task execution will be "observed" when calling `.Result` at this point they are then thrown on the thread that calls `.Result`.

> Note that, if an exception occurred during the operation of the task, or if the task has been cancelled, the Result property does not return a value. Instead, attempting to access the property value throws an AggregateException exception.
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.result?view=net-5.0

:two: In the event that as an exception is thrown in a task execution and that exception is not "observed", your process will not crash, **in most cases**. Anything after .NET 4.5 (.netcore and net5) will not crash due to "un-observed" exceptions see: https://devblogs.microsoft.com/pfxteam/task-exception-handling-in-net-4-5/.

To me it seems like the culprit in your crash might have been misdiagnosed, it's likely not the exceptions that were thrown during the Task executions but rather the exceptions thrown (validly IMOP) during close/cancel which your application did not handle gracefully.

dnawnOlo:
@akrock Thanks for the feedback. Much appreciated. Sorry, I was wrong with my `Any exceptions that occur in the method being called in Task.Run() end up as unhandled exceptions that cause the process to terminate.` statement. The issue I'm running into is a task being cancelled. I cloned this repo and brought it into my solution as a project reference so I could step through and see what was happening. What I observed was a Task Canceled Exception causing an AggregateException on SnowflakeDbConnection.Dispose() which results in this error:

```
An unhandled exception occurred and the process was terminated.

Application ID: /LM/W3SVC/1/ROOT

Process ID: 21276

Exception: System.AggregateException

Message: One or more errors occurred.

StackTrace:    at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Snowflake.Data.Core.SFSession.close()
   at Snowflake.Data.Client.SnowflakeDbConnection.Close()
   at Snowflake.Data.Client.SnowflakeDbConnection.Dispose(Boolean disposing)
   at Snowflake.Data.Client.SnowflakeDbConnection.Finalize()

InnerException: System.Threading.Tasks.TaskCanceledException

Message: A task was canceled.
```

The only reason this is occurring is because I do not have network connectivity to Snowflake which is to simulate an issue we ran into in our Production environment. I believe any calls the RestRequester makes in this case will end up as Task Canceled Exceptions unless there is a timeout specified on the request or a cancellation token with a timeout is used.

I moved away from a `using` block to explicitly disposing of the connection as follows:

```
private async Task<T> ExecuteQuery<T>(string methodName, Func<IDbConnection, Task<T>> getTask)
        {
            var conn = new SnowflakeDbConnection() { ConnectionString = _connectionString };

            try
            {
                var cancellationToken = new CancellationTokenSource(10000);
                await conn.OpenAsync(cancellationToken.Token);

                var task = getTask(conn);
                return await task;
            }
            catch (AggregateException ae)
            {
                foreach (var e in ae.InnerExceptions)
                {
                    Logger.Error(e, $"{methodName}: An error occurred trying to execute query");
                }
                throw;
            }
            catch (Exception ex)
            {
                Logger.Error(ex, $"{methodName}: An error occurred trying to execute query");
                throw;
            }
            finally
            {
                DisposeConnection(conn);
            }
        }

        private void DisposeConnection(SnowflakeDbConnection conn)
        {
            try
            {
                conn?.Dispose();
            }
            catch (AggregateException ae)
            {
                foreach (var e in ae.InnerExceptions)
                {
                    Logger.Error(e, "An error occurred trying to dispose of connection");
                }
                throw;
            }
            catch (Exception ex)
            {
                Logger.Error(ex, "An error occurred trying to dispose of connection");
                throw;
            }
        }
    }
```

I can catch the exception when trying to explicitly dispose, but since an exception is thrown within https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Client/SnowflakeDbConnection.cs#L164, SnowflakeDbConnection isn't disposed of. When the garbage collector comes around and calls the Dispose method on SnowflakeDbConnection, it throws the TaskCanceledException, but I have no way of catching in that scenario.

akrock:
@dnawnOlo That makes a lot of sense, thanks for the clarifications.

I think in this case we could probably get away w/ simply wrapping the call to `Close()` in the dispose method w/ a try/catch log. It would keep the behaviors where `Close()` & `Cancel()` will still throw exceptions, yet prevent this potential issue where the connector might blow an exception on a finalizer (which is definitely bad).

The snowflake committers might have their own opinions on this as well.

dnawnOlo:
@akrock Yeah, that approach makes a lot more sense. After doing that, I verified that that ultimately takes care of my issue. The TaskCanceledException is caught in the catch and logged, but Dispose no longer blows up and the process does not crash.

akrock:
Looks good to me.

SimbaGithub (approved):

sfc-gh-hyu (approved):

PR #298, opened 2021-04-23
SNOW-348801 - Local http connection settings with upgrade to net framework472

SimbaGithub:


README.md
```diff
@@ -377,3 +377,8 @@ This CVE has been reported in systems.text.regularexpressions.dll which is used
 * Delete any logs collected thus far and make sure that all copies are deleted. 
 * If you cannot upgrade for any reason, please ensure all debugging is disabled
 * If you are concerned about a potential compromise, contact Snowflake Customer Support for assistance with invalidating all active sessions/tokens. 
+
+3. Global http connection settings -
```
sfc-gh-dyoshinaga, line 381: Can you change "http" to "HTTP"?

README.md
```diff
@@ -377,3 +377,8 @@ This CVE has been reported in systems.text.regularexpressions.dll which is used
 * Delete any logs collected thus far and make sure that all copies are deleted. 
 * If you cannot upgrade for any reason, please ensure all debugging is disabled
 * If you are concerned about a potential compromise, contact Snowflake Customer Support for assistance with invalidating all active sessions/tokens. 
+
+3. Global http connection settings - 
+	Snowflake has identified an issue where the connector is globally enforcing TLS 1.2 and certificate revocation checks with connectors up to v1.2.1.
```
sfc-gh-dyoshinaga, line 382: Can you change "connector" to "driver"?

sfc-gh-dyoshinaga, line 382: Can you change "connectors up to v1.2.1" to "the .NET Driver v1.2.1 and earlier versions"?

README.md
```diff
@@ -377,3 +377,8 @@ This CVE has been reported in systems.text.regularexpressions.dll which is used
...
+	Starting with v2.0.0, the connector will set these locally. However note that connector is now targeting .Net framework 4.7.2.
```
sfc-gh-dyoshinaga, line 383: Can you add a blank line before this so that this is a separate paragraph?

Currently, all of this is being treated as a single paragraph. ( https://github.com/snowflakedb/snowflake-connector-net/blob/04ce70ae0b67911c9c56ae693f0b84ceabfa2419/README.md#notice )

sfc-gh-dyoshinaga, line 383: Maybe move the "However" to the next paragraph and add a blank line before it. e.g.

```
Starting with ...

Note that v2.0.0 of the driver requires version 4.7.2 (or a more recent version) of the .NET Framework. In addition, when upgrading to v2.0.0, you might need to run "Update-Package -reinstall" to update the dependencies.
```

sfc-gh-dyoshinaga (commented):
Thank you for updating the docs!

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -173,6 +177,11 @@ internal RetryHandler(HttpMessageHandler innerHandler) : base(innerHandler)
                 HttpResponseMessage response = null;
                 int backOffInSec = 1;
 
+                ServicePoint p = ServicePointManager.FindServicePoint(requestMessage.RequestUri);
+                p.Expect100Continue = false; // Saves about 100 ms per request
+                p.UseNagleAlgorithm = false; // Saves about 200 ms per request
+                p.ConnectionLimit = 20;      // Default value is 2, we need more connections for performing multiple parallel queries
```
sfc-gh-hyu, line 183: Sigh. this looks ugly, but I also don't find a good way to enforce that. My only questions is this code will be executed for every request right? Not just the retried requests right?

sfc-gh-hyu (approved):

sfc-gh-hyu:
cc @stegus64

Snowflake.Data/Core/HttpUtil.cs
SimbaGithub: Yes, it will since it's in the SendAsync function. But I suspect that something similar is done under the hood using the global parameter anyway.

README.md
SimbaGithub: @sfc-gh-dyoshinaga : I think I addressed all the comments, please double-check.

sfc-gh-dyoshinaga: Thank you -- these look resolved

sfc-gh-dyoshinaga (approved):
LGTM for README.md

PR #299, opened 2021-04-23
SNOW-346204: Correct an incorrect INSERT statement.

sfc-gh-mgilkey:
Hi Shige,

I am not sure which reviewer to assign this to. Please feel free to re-assign this if appropriate. Thank you.

--mgilkey

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,11 @@ using (IDbConnection conn = new SnowflakeDbConnection())
     conn.Open();
 
     IDbCommand cmd = conn.CreateCommand();
+    cmd.CommandText = "create or replace table T(cola int)";
+    int count = cmd.ExecuteNonQuery();
+    Assert.AreEqual(0, count);
+
+    IDbCommand cmd = conn.CreateCommand();
     cmd.CommandText = "insert into t values (?), (?), (?)";
 
     var p1 = cmd.CreateParameter();
@@ -323,6 +328,10 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 
     var count = cmd.ExecuteNonQuery();
     Assert.AreEqual(3, count);
+
+    cmd.CommandText = "drop table if exists T";
+    count = cmd.ExecuteNonQuery();
+    Assert.AreEqual(0, count);
 
     conn.Close();
 }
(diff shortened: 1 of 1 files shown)


sfc-gh-stakeda:
@sfc-gh-sshankar , please assign this review.

SimbaGithub:
The original insert statement is correct. 
It shows how to insert 3 rows in a table with one column and how bound parameters are converted from C# to Snowflake type (here, let's say the column is int, it shows that you can bind int32 or int16 C# data).
The change made here assumes inserting one row in a table with 3 columns and the last assert "Assert.AreEqual(3, count);" will fail because only one row will be added.

sfc-gh-mgilkey:
I greatly appreciate the correction and the explanation. I've restored the original code, and added the explanation so that future readers don't get confused.

SimbaGithub:
This looks good, thanks for adding the description !
If you want, the following example includes the table creation and drops it at the end, this way it is clear what is used. I would also keep the description you added, because it helps.

```
using (IDbConnection conn = new SnowflakeDbConnection())


            {
                conn.ConnectionString = ConnectionString;
                conn.Open();
                
                IDbCommand cmd = conn.CreateCommand();
                cmd.CommandText = "create or replace table T(cola int)";
                int count = cmd.ExecuteNonQuery();
                Assert.AreEqual(0, count);

                cmd = conn.CreateCommand();
                cmd.CommandText = "insert into T values (?),(?),(?)";
                var p1 = cmd.CreateParameter();
                p1.ParameterName = "1";
                p1.Value = 10;
                p1.DbType = DbType.Int32;
                cmd.Parameters.Add(p1);
                var p2 = cmd.CreateParameter();
                p2.ParameterName = "2";
                p2.Value = 10000L;
                p2.DbType = DbType.Int32;
                cmd.Parameters.Add(p2);
                var p3 = cmd.CreateParameter();
                p3.ParameterName = "3";
                p3.Value = (short)1;
                p3.DbType = DbType.Int16;
                cmd.Parameters.Add(p3);
                count = cmd.ExecuteNonQuery();
                Assert.AreEqual(3, count);

                cmd.CommandText = "drop table if exists T";
                count = cmd.ExecuteNonQuery();
                Assert.AreEqual(0, count);

                conn.Close();
            }
```

SimbaGithub (approved):

PR #312, opened 2021-05-11
SNOW-348919  - Remove hidden httpclient timeout, changed connection timeout and Rest request default timeout to 15 min

SimbaGithub:


Snowflake.Data/Core/SFSession.cs
```diff
@@ -149,21 +172,25 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
...
-            SFRestRequest closeSessionRequest = new SFRestRequest
-            {
-                Url = BuildUri(RestPath.SF_SESSION_PATH, queryParams),
-                authorizationToken = string.Format(SF_AUTHORIZATION_SNOWFLAKE_FMT, sessionToken)
-            };
-          
-            var response = restRequester.Post<CloseResponse>(closeSessionRequest);
-            if (!response.success)
+            // Only close the session if it is opened
+            if (null != sessionToken)
```
sfc-gh-hyu, line 176: Maybe change this to 
```
if (sessionToken == null) return;

....
```
This way, you don't need to include majority of the code in this if block.

Snowflake.Data/Core/SFSession.cs
```diff
@@ -92,15 +92,38 @@ internal Uri BuildLoginUrl()
...
+                }
+
+                connectionTimeout = timeoutInSec > 0 ? TimeSpan.FromSeconds(timeoutInSec) : Timeout.InfiniteTimeSpan;
 
-            int timeoutInSec = int.Parse(properties[SFSessionProperty.CONNECTION_TIMEOUT]);
-            connectionTimeout = timeoutInSec > 0 ? TimeSpan.FromSeconds(timeoutInSec) : Timeout.InfiniteTimeSpan;
+            }
+            catch (Exception e)
+            {
+                logger.Error(e.Message);
```
sfc-gh-hyu, line 122: I don't understand this try-catch block. What kind of error could this block throw?

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -176,20 +175,26 @@ internal RetryHandler(HttpMessageHandler innerHandler) : base(innerHandler)
...
                 ServicePoint p = ServicePointManager.FindServicePoint(requestMessage.RequestUri);
                 p.Expect100Continue = false; // Saves about 100 ms per request
                 p.UseNagleAlgorithm = false; // Saves about 200 ms per request
                 p.ConnectionLimit = 20;      // Default value is 2, we need more connections for performing multiple parallel queries
 
-                TimeSpan httpTimeout = (TimeSpan)requestMessage.Properties["TIMEOUT_PER_HTTP_REQUEST"];
+                TimeSpan httpTimeout = (TimeSpan)requestMessage.Properties[SFRestRequest.HTTP_REQUEST_TIMEOUT_KEY];
+                logger.Debug("Http request timeout : " + httpTimeout);
+                TimeSpan restTimeout = (TimeSpan)requestMessage.Properties[SFRestRequest.REST_REQUEST_TIMEOUT_KEY];
+                logger.Debug("Rest request timeout : " + restTimeout);
```
sfc-gh-hyu, line 189: Use string format?

sfc-gh-hyu (changes requested):

Snowflake.Data/Core/HttpUtil.cs
SimbaGithub: Concatenation is supposed to be faster for .Net.
But I think it raises a point about the time we'll spend logging.
I added these log lines in debug, just in case with need to troubleshoot more issues, to make sure that the correct timeouts are propagated to the httpclient.

I did a micro benchmark and I will update the code shortly with the optimized version which will check if IsDebugEnabled.

With log level Debug
|              Method |        Mean |     Error |     StdDev |      
|-------------------- |------------:|----------:|-----------:|
|             WithLog |   856.41 ns | 28.928 ns |  84.840 ns |
|      CheckBeforeLog |    15.32 ns |  0.431 ns |   1.250 ns | 
| WithStringFormatLog | 1,440.98 ns | 88.273 ns | 258.889 ns | 

With log level in Error
|              Method |        Mean |     Error |    StdDev |
|-------------------- |------------:|----------:|----------:|
|             WithLog |   791.04 ns | 15.815 ns | 18.212 ns |
|      CheckBeforeLog |    12.00 ns |  0.244 ns |  0.342 ns |
| WithStringFormatLog | 1,166.14 ns | 22.869 ns | 32.798 ns |

Snowflake.Data/Core/SFSession.cs
sfc-gh-hyu: Then I would prefer a smaller try catch block. What do you think?

sfc-gh-hyu (approved):

Snowflake.Data/Core/SFSession.cs
SimbaGithub: Yes, I think we can have it around the Boolean.Parse and int.Parse. There should not be any NPE or other things that would throw in the code after that.

sfc-gh-hyu (approved):
LGTM

sfc-gh-hyu (approved):

PR #315, opened 2021-05-12
SNOW-197209 - Add support for DB.Null for all supported DBType and SFDataType values

SimbaGithub:


Snowflake.Data/Core/SFSession.cs
```diff
@@ -222,6 +222,7 @@ internal void UpdateSessionParameterMap(List<NameValueParameter> parameterList)
             logger.Debug("Update parameter map");
             foreach (NameValueParameter parameter in parameterList)
             {
+                logger.Debug(parameter.name + " : " + parameter.value);
```
sfc-gh-hyu, line 225: I am not sure how expensive is it in .net or c#, but in java, string concat in logging is probably not a good idea? I should probably use DebugFmt or String format?

sfc-gh-hyu (approved):

Snowflake.Data/Core/SFSession.cs
SimbaGithub: Oups, this log line was not supposed to make it in the PR. It was there for me only. I'll just remove it completely.

PR #311, opened 2021-05-07
Prevent memory leaks due to expired certs

alexrosenfeld10:
See https://github.com/dotnet/core/issues/6229 and https://github.com/newrelic/newrelic-dotnet-agent/issues/544

@SimbaGithub This should probably be released ASAP, it has a pretty devastating impact and is really hard to debug.

PR #317, opened 2021-05-14
SNOW-357759 - Fix critical bug in sf data converter for converting booleans

alexrosenfeld10:
Credit to @tdg5 for the source code fix.

### DIFF
--- a/Snowflake.Data/Core/SFDataConverter.cs
+++ b/Snowflake.Data/Core/SFDataConverter.cs
@@ -70,7 +70,8 @@ namespace Snowflake.Data.Core
                 }
                 else if (destType == typeof(Boolean))
                 {
-                    return srcVal.Buffer[0] == '1' || srcVal.Buffer[0] == 't' || srcVal.Buffer[0] == 'T';
+                    var val = srcVal.Buffer[srcVal.offset];
+                    return val == '1' || val == 't' || val == 'T';
                 }
                 else if (destType == typeof(byte[]))
                 {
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -70,7 +70,8 @@ internal static object ConvertToCSharpVal(UTF8Buffer srcVal, SFDataType srcType,
                 }
                 else if (destType == typeof(Boolean))
                 {
-                    return srcVal.Buffer[0] == '1' || srcVal.Buffer[0] == 't' || srcVal.Buffer[0] == 'T';
+                    var val = srcVal.Buffer[srcVal.offset];
```
alexrosenfeld10, line 73: You'll find that the unit tests will _fail_ if you remove this source code change. This is super critical as it can mutate data and return `false` for things that are actually `true`.

alexrosenfeld10:
@SimbaGithub @sfc-gh-hyu this caused a sev0 for us. I have a support ticket open with Snowflake about it.

sfc-gh-hyu (approved):
Thanks for this. LGTM @SimbaGithub Can you take a look as well?

alexrosenfeld10:
@sfc-gh-hyu Appreciate the quick review. Please check #311 while you're here - it's equally as critical.

Not to be rude, but are there plans to ramp up maintenance and support in this repo? There is a lot here that could be improved, and my company has invested _hours_ of my and other engineers time into a private fork due to these kind of critical bugs.

sfc-gh-hyu:
@alexrosenfeld10 I understood your position and I am sorry for that. I already comment in the other jira, but I would rather let my security folks comment on that.

alexrosenfeld10:
@sfc-gh-hyu Thanks, I'm glad to at least have some kind of response on the pulls.

PR #326, opened 2021-05-26
SNOW-361754 - Add proxy support

SimbaGithub:
Add proxy support via connection setting.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -199,7 +199,16 @@ namespace Snowflake.Data.Client
             if (disposed)
                 return;
 
-            this.Close();
+            try
+            {
+                this.Close();
+            } 
+            catch (Exception ex)
+            {
+                // Prevent an exception from being thrown when disposing of this object
+                logger.Error("Unable to close connection", ex);
+            }
+            
             disposed = true;
 
             base.Dispose(disposing);
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -158,6 +158,7 @@ namespace Snowflake.Data.Client
                     }
                     else
                     {
+                        logger.Debug("All good");
                         // Only continue if the session was opened successfully
                         OnSessionEstablished();
                     }
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -116,9 +116,12 @@ namespace Snowflake.Data.Client
                 logger.Error("Unable to connect", e);
                 if (!(e.GetType() == typeof(SnowflakeDbException)))
                 {
-                    throw new SnowflakeDbException(e.InnerException,
-                                SFError.INTERNAL_ERROR,
-                                "Unable to connect");
+                    throw 
+                       new SnowflakeDbException(
+                           e,
+                           SnowflakeDbException.CONNECTION_FAILURE_SSTATE,
+                           SFError.INTERNAL_ERROR,
+                           "Unable to connect. " + e.Message);
                 }
                 else
                 {
@@ -142,9 +145,11 @@ namespace Snowflake.Data.Client
                     Exception sfSessionEx = previousTask.Exception;
                         _connectionState = ConnectionState.Closed;
                         logger.Error("Unable to connect", sfSessionEx.InnerException);
-                        throw new SnowflakeDbException(sfSessionEx.InnerException,
-                            SFError.INTERNAL_ERROR,
-                            "Unable to connect");
+                        throw new SnowflakeDbException(
+                           sfSessionEx,
+                           SnowflakeDbException.CONNECTION_FAILURE_SSTATE,
+                           SFError.INTERNAL_ERROR,
+                           "Unable to connect");
                     }
                     else if (previousTask.IsCanceled)
                     {
@@ -153,9 +158,8 @@ namespace Snowflake.Data.Client
                     }
                     else
                     {
-                        logger.Debug("All good");
-                    // Only continue if the session was opened successfully
-                    OnSessionEstablished();
+                        // Only continue if the session was opened successfully
+                        OnSessionEstablished();
                     }
                 },
                 cancellationToken);
--- a/Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
+++ b/Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
@@ -73,7 +73,7 @@ namespace Snowflake.Data.Core.Authenticator
             VerifyPostbackUrl();
 
             logger.Debug("step 6: send SAML reponse to snowflake to login");
-            await base.LoginAsync(cancellationToken);  
+            await base.LoginAsync(cancellationToken).ConfigureAwait(false);  
         }
 
         void IAuthenticator.Authenticate()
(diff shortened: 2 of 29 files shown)


Snowflake.Data/Core/ChunkDownloaderFactory.cs
```diff
@@ -31,7 +31,8 @@ class ChunkDownloaderFactory
                         responseData.chunks,
                         responseData.qrmk,
                         responseData.chunkHeaders,
-                        cancellationToken);
+                        cancellationToken,
+                        resultSet.sfStatement.SfSession.restRequester);
```
sfc-gh-hyu, line 35: Do we need to update other chunk downloaders as well?

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -64,12 +87,17 @@ static private void initHttpClient()
                 // Enforce tls v1.2
                 SslProtocols = SslProtocols.Tls12,
                 AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
-            };
+                CookieContainer = cookieContainer = new CookieContainer()
```
sfc-gh-hyu, line 90: Why do we need this?

Snowflake.Data/Core/RestRequester.cs
```diff
@@ -34,17 +36,87 @@ internal class RestRequester : IRestRequester
...
+        /// <param name="noProxyList">The list of urls to by-pass the proxy.</param>
+        /// <returns>The RestRequester for this proxy.</returns>
+        /// <exception cref="System.FormatException">The port is not a valid int</exception>
+        /// <exception cref="System.OverflowException">The port value is too large</exception>
+        internal static IRestRequester GetRestRequester(
+            string proxyHost,
+            string proxyPort,
+            string proxyUser,
+            string proxyPassword,
+            string noProxyList)
```
sfc-gh-hyu, line 69: Can we wrap those string into a single class/struct, maybe name it `ProxyInfo`? It might be easier since you don't need to pass 5 arguments around.

Snowflake.Data/Core/RestRequester.cs
```diff
@@ -34,17 +36,87 @@ internal class RestRequester : IRestRequester
     {
         private static SFLogger logger = SFLoggerFactory.GetLogger<RestRequester>();
 
-        private static readonly RestRequester instance = new RestRequester();
+        private static Dictionary<string, IRestRequester> RestRequestersByProxy = 
+            new Dictionary<string, IRestRequester>();
```
sfc-gh-hyu, line 40: It seems a little weird to use string as the key here? Can we use IWebProxy as the key here as well? Then we can move the initialization of webproxy inside SFSession?

Snowflake.Data/Core/RestRequester.cs
sfc-gh-hyu: I think we should pass in IWebProxy here?

SimbaGithub: That was my original idea too, but it isn't really going to work.
The pool of requesters is global to the driver.
The RestRequester is the one who has the responsibility to create a new IWebProxy based on the connection settings if needed, because it "knowns" about the existing ones.
Where would you create the IWebProxy otherwise? It's an object, to make sure a new one is not created when there is already one for the same settings, you would have to go through the list of existing IWebProxy and check the Address, Credentials and byPassList properties. This is why I use a map of settings/IWebProxy as a fast way to know if there was an existing IWebProxy ready to use.

SimbaGithub: For the first comment, I am not sure that a class/struct is really needed.
It would mean creating a new instance, store the info inside, pass the object instead of the strings themselves and right away extract the strings from the class/struct. 
There is only one call to GetRestRequester in SFSession to get an requester for the Session. This is more or less similar to calling a constructor. The difference here is that it creates a new instance only if needed. I would not create a class/struct for a constructor, would you?
I also like to make it clear, by listing them, that the RestRequester instance returned is based on these 5 settings. If you use a class/struct it hides this a little bit more.

Snowflake.Data/Core/RestRequester.cs
SimbaGithub: No, this is not really possible. The RestRequester has the responsibility to create the IWebProxy instance. Using a IWebProxy has a key means that it has to be the same instance. If SFSession creates an IWebProxy instance based on the connection settings, even if the connection settings for 2 SnowflakeDbConnection are the same, the IWebProxy of each SFSession will be a different instance, and therefor a different key.

Snowflake.Data/Core/HttpUtil.cs
SimbaGithub: umh, so I looked back into the file history. It seems that I inadvertently removed it last time and none of use caught it. This is necessary for subsequent Okta connection to work (https://github.com/snowflakedb/snowflake-connector-net/pull/272).

Snowflake.Data/Core/ChunkDownloaderFactory.cs
SimbaGithub: Yes, I did update them too. 
The other chunk downloaders have the resultset object and are getting the restRequester from there.
See changes in SFBlockingChunkDownloader.cs and SFBlockingChunkDownloaderV3.cs.

sfc-gh-hyu (approved):

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -153,9 +158,8 @@ public override Task OpenAsync(CancellationToken cancellationToken)
                     }
                     else
                     {
-                        logger.Debug("All good");
```
sfc-gh-hyu, line 156: We should probably not removing this log?

Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
```diff
@@ -46,11 +46,6 @@ async Task IAuthenticator.AuthenticateAsync(CancellationToken cancellationToken)
         {
             logger.Info("Okta Authentication");
 
-            // Clear cookies before authenticating because when a cookie is present in the request,
-            // Okta will assume it is coming from a browser and perform a CSRF check.
-            // This will ensure that we are NOT including the ‘sid’ cookie with the request.
-            HttpUtil.ClearCookies(oktaUrl);
```
sfc-gh-hyu, line 52: So we don't need to clear cookies any more?

sfc-gh-hyu (approved):

Snowflake.Data/Client/SnowflakeDbConnection.cs
SimbaGithub: I added this mainly for myself, about 2 commits ago, so that I could do a quick grep on the log. But, you are right, it won't hurt. I'll add it back.

Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
SimbaGithub: I disabled the cookies in the httpHandler (UseCookies = false // Disable cookies) so it should not be necessary anymore.
I'm pretty sure it tested that it was working fine, but I'll run the Okta test again to be 100% certain.

Snowflake.Data/Client/SnowflakeDbConnection.cs
SimbaGithub: Added back the line.

Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
SimbaGithub: Retested okta, external browser and JWT, it works fine. Turning off cookies in the http handler means that we don't need to clean-up.

sfc-gh-hyu (approved):

PR #335, opened 2021-06-17
Add SecretDetector and unit tests

SimbaGithub:


SimbaGithub:
@sfc-gh-hchaturvedi Similar to the nodejs connector, it currently does not use telemetry but let me know if there are areas where the Secret Detector should be used

Snowflake.Data/Logger/SecretDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+                                    MaskAWSTokens(
+                                        MaskSASTokens(
+                                            MaskAWSKeys(text)))))));
+
+                if (result.maskedText != text)
+                {
+                    result.isMasked = true;
+                }
+            }
+            catch (Exception ex)
```
sfc-gh-hchaturvedi, line 101: Can we add a log line in here to communicate that an exception occurred during secret masking?

sfc-gh-hchaturvedi:
The main logic seems consistent with the secret detector of other drivers. Have we thought of any other secrets that .NET deals with or secret patterns that are exclusive to .NET that might not be common with other drivers? Like how do we configure the passwords in .NET do we call them pwd or Passwords or something else?

sfc-gh-hchaturvedi:
> @sfc-gh-hchaturvedi Similar to the nodejs connector, it currently does not use telemetry but let me know if there are areas where the Secret Detector should be used

The secret detector should be used for parsing logs primarily. Any logs that we generated should pass through this before being dumped to a file or elsewhere. Not sure if .NET / NodeJS has telemetry, but if they do the telemetry messages should go through the detector as well.

Snowflake.Data/Logger/SecretDetector.cs
SimbaGithub: I tried this but there's an issue when both secret detector and logger call each other:
- Logger references SecretDetector to mask messages
- SecretDetector references Logger to log the exception

The logger could be changed from static to non-static but the logger was made non-static so that users can pass their own logger.

The exception thrown is recorded in `result.maskedText = ex.Message;` so that it shows on the log but let me know if you have suggestions

SimbaGithub:
> The main logic seems consistent with the secret detector of other drivers. Have we thought of any other secrets that .NET deals with or secret patterns that are exclusive to .NET that might not be common with other drivers? Like how do we configure the passwords in .NET do we call them pwd or Passwords or something else?

I didn't see any secrets exclusive to .NET, it uses the same set of authenticators which the existing patterns cover. The other logs are Http-related but don't contain sensitive info. 
Based on SFSessionProperty.cs, passwords are called "PASSWORD" which is included in the patterns

SimbaGithub:
@sfc-gh-hchaturvedi The logger now uses the SecretDetector to mask messages. For custom patterns, since the SecretDetector is static, it uses the following functions to set and clear custom patterns.
- Set custom patterns:
```
string[] regex = new string[2] 
{ 
    @"(testCustomPattern\s*:\s*""([a-z]{8,})"")", 
    @"(testCustomPattern\s*:\s*""([0-9]{8,})"")" 
};
string[] masks = new string[2] { "maskCustomPattern1", "maskCustomPattern2" };

SecretDetector.SetCustomPatterns(regex, masks);
```


- Clear custom patterns:
```
SecretDetector.ClearCustomPatterns();
```

Snowflake.Data/Logger/SecretDetector.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                if (result != text)
+                {
+                    return result;
+                }
+            }
+            return text;
+        }
+
+        private static readonly string AWS_KEY_PATTERN = @"(aws_key_id|aws_secret_key|access_key_id|secret_access_key)\s*=\s*'([^']+)'";
```
sfc-gh-hchaturvedi, line 76: Can we update this to include 

`'AWS_SECRET_KEY': 'bajbpafjb;a;adjpboajbvjavbjadj'`

I think the current pattern only accounts for `=` signs and not `:`, also quotes around the key

sfc-gh-hchaturvedi (approved):
LGTM! Just the one comment about the AWS secret key should be addressed before merging.

Snowflake.Data/Logger/SecretDetector.cs
SimbaGithub: Ok

PR #336, opened 2021-06-24
Better dispose of CancellationTokenSources

tdg5:
`CancellationTokenSource` implements `IDisposable` so should be disposed. My experience suggests that these do get GC'd eventually as it currently stands, but it seems like explicit handling is more desirable.

### DIFF
--- a/Snowflake.Data/Core/HttpUtil.cs
+++ b/Snowflake.Data/Core/HttpUtil.cs
@@ -245,12 +245,17 @@ namespace Snowflake.Data.Core
                         }
                     }
 
+                    if (childCts != null)
+                    {
+                        childCts.Dispose();
+                    }
+
                     if (response != null)
                     {
                         if (response.IsSuccessStatusCode) {
                             return response;
                         }
-                        else 
+                        else
                         {
                             logger.Debug($"Failed Response: {response.ToString()}");
                             bool isRetryable = isRetryableHTTPCode((int)response.StatusCode);
@@ -261,7 +266,7 @@ namespace Snowflake.Data.Core
                             }
                         }
                     }
-                    else 
+                    else
                     {
                         logger.Info("Response returned was null.");
                     }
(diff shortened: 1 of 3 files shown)


tdg5:
Here's some sample output from running `dotnet-dump collect` + `dotnet-dump analyze -c dumpheap` showing there are more `CancellationTokenSource` things lying around than have any right to be after a few minutes running an app making 20-50 queries per minute:
```
Statistics:
              MT    Count    TotalSize Class Name
00007fa773a993a0     3429       219456 System.Threading.CancellationTokenSource
00007fa778a6bcd8     3440       275200 System.Threading.CancellationTokenSource+Linked1CancellationTokenSource
00007fa775e7ada8     3456       331776 System.Threading.TimerQueueTimer
00007fa775e7aef8     6921       387576 System.Threading.CancellationTokenSource+CallbackPartition
00007fa775e7af68     6872       604736 System.Threading.CancellationTokenSource+CallbackPartition[]
00007fa775e7b628    10300       824000 System.Threading.CancellationTokenSource+CallbackNode
```

alexrosenfeld10 (approved):

alexrosenfeld10:
cc

@sfc-gh-kterada @sfc-gh-hkapre

SimbaGithub:
This looks like a good change, but I think we should also add clean-up for the other Cancellation tokens because they will be leaking too, isn't it?
I found restRequestTimeout and linkedCts in RestRequester and childCts in HttpUtil. 
It can likely be done in a finally block for these ones.

tdg5:
@SimbaGithub I added handling for the other CancellationTokenSource instances you mentioned, but you owe me one 😜 . We are now better able to confirm that the `CancellationTokenSource` instances in SFStatement were causing the memory growth issue we were seeing.
memory-spikes-better-yay

SimbaGithub (approved):

PR #340, opened 2021-06-30
Add getTimeSpan to SnowflakeDbDataReader

SimbaGithub:
Add getTimeSpan to SnowflakeDbDataReader for retrieving TIME data as Timespan

### DIFF
--- a/Snowflake.Data/Core/SFDataConverter.cs
+++ b/Snowflake.Data/Core/SFDataConverter.cs
@@ -219,7 +219,7 @@ namespace Snowflake.Data.Core
                 }
             }
 
-            return (intPart * 1000 * 1000 * 1000 + decimalPart) / 100L;
+            return intPart * 10000000L + decimalPart / 100L;
 
         }
 
--- a/Snowflake.Data/Core/SFDataConverter.cs
+++ b/Snowflake.Data/Core/SFDataConverter.cs
@@ -64,6 +64,10 @@ namespace Snowflake.Data.Core
                 {
                     return ConvertToDateTime(srcVal, srcType);
                 }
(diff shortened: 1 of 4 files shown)


Snowflake.Data/Core/SFDataConverter.cs
```diff
@@ -184,30 +196,33 @@ private static DateTimeOffset ConvertToDateTimeOffset(UTF8Buffer srcVal, SFDataT
             100000000 
         };
 
-        private static Tuple<long, long> ExtractTimestamp(UTF8Buffer srcVal)
+        private static long GetTicksFromSecondAndNanosecond(UTF8Buffer srcVal)
```
sfc-gh-hyu, line 199: Maybe add a comment on when this is method is used. I guess this is used where caller only needs precision of microseconds ? since it looks like you divide 100 at the end of this function?

sfc-gh-hyu (approved):
LGTM

Snowflake.Data/Core/SFDataConverter.cs
SimbaGithub: .Net DateTime and DateTimeOffset don't support nanoseconds precision. The best you can get are Ticks.
A single tick represents one hundred nanoseconds. This is why it's divided by 100 at the end. If you look at the original code, you'll see that this was done for each data in ConvertToDateTime before. I simply moved it to a common place to avoid duplication and confusion.
This method does return the number of ticks for the given seconds and nanoseconds value. I tried to get a meaningful name for it, looks like I failed :D I'll add a comment and submit the PR.

sfc-gh-hyu (approved):

PR #348, opened 2021-07-22
SNOW-399633 Document support for proxies and two new methods

sfc-gh-dyoshinaga:
Description
Document support for proxies and two new methods (SnowflakeDbDataReader.getTimeSpan and SnowflakeDbConnection.CloseAsync).

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,12 +124,12 @@ The following table lists all valid connection properties:
 | PRIVATE_KEY                |Depends   | The private key to use for key-pair authentication. Must be used in combination with AUTHENTICATOR=snowflake_jwt. <br /> If the private key value includes any equal signs (=), make sure to replace each equal sign with two signs (==) to ensure that the connection string is parsed correctly.|
 | TOKEN                      |Depends   | The OAuth token to use for OAuth authentication. Must be used in combination with AUTHENTICATOR=oauth.|
 | INSECUREMODE               |No   	| Set to true to disable the certificate revocation list check. Default is false.|
-| USEPROXY                   | No       | Set to true if you need to use a proxy server. The default value is false. |
-| PROXYHOST                  | No       | The hostname of the proxy server. |
-| PROXYPORT                  | No       | The port number of the proxy server. |
-| PROXYUSER                  | No       | The username for authenticating to the proxy server. |
-| PROXYPASSWORD              | No       | The password for authenticating to the proxy server. |
-| NONPROXYHOSTS              | No       | The list of hosts that the driver should connect to directly, bypassing the proxy server. Separate the hostnames with a URL-escaped pipe symbol (`%7C`). |
+| USEPROXY                   | No       | Set to true if you need to use a proxy server. The default value is false. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYHOST                  | No       | The hostname of the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYPORT                  | No       | The port number of the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYUSER                  | No       | The username for authenticating to the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYPASSWORD              | No       | The password for authenticating to the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| NONPROXYHOSTS              | No       | The list of hosts that the driver should connect to directly, bypassing the proxy server. Separate the hostnames with a pipe symbol (`|`). You can also use an asterisk (`*`) as a wildcard. <br/> <br/> This parameter was introduced in v2.0.4. |
 
 
 <br />
@@ -270,8 +270,9 @@ If you are using a different method for authentication, see the examples below:
   * `{okta_url_endpoint}` is the URL for the endpoint for your Okta account (e.g. `https://<okta_account_name>.okta.com`).
   * `{login_name_for_IdP}` is your login name for your IdP.
 
-The following example configures the driver to connect through the proxy server `myproxyserver` on port `8888`. The driver
-authenticates to the proxy server as the user `test` with the password `test`:
+In v2.0.4 and later releases, you can configure the driver to connect through a proxy server. The following example configures the
+driver to connect through the proxy server `myproxyserver` on port `8888`. The driver authenticates to the proxy server as the
+user `test` with the password `test`:
 
 ```cs
 using (IDbConnection conn = new SnowflakeDbConnection())
@@ -307,7 +308,14 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 ```
 
 Note that for a `TIME` column, the reader returns a `System.DateTime` value. If you need a `System.TimeSpan` column, call the
-`getTimeSpan` method in `SnowflakeDbDataReader`.
+`getTimeSpan` method in `SnowflakeDbDataReader`. This method was introduced in the v2.0.4 release.
+
+Note that because this method is not available in the generic `IDataReader` interface, you must cast the object as
+`SnowflakeDbDataReader` before calling the method. For example:
+
+```cs
+TimeSpan timeSpanTime = ((SnowflakeDbDataReader)reader).GetTimeSpan(13);
+```
 
 Bind Parameter
 --------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,7 +123,13 @@ The following table lists all valid connection properties:
...
 | TOKEN                      |Depends   | The OAuth token to use for OAuth authentication. Must be used in combination with AUTHENTICATOR=oauth.|
-| INSECUREMODE               |No   	    | Set to true to disable the certificate revocation list check. Default is false.|
+| INSECUREMODE               |No   	| Set to true to disable the certificate revocation list check. Default is false.|
+| USEPROXY                   | No       | Set to true if you need to use a proxy server. The default value is false. |
+| PROXYHOST                  | No       | The hostname of the proxy server. |
+| PROXYPORT                  | No       | The port number of the proxy server. |
+| PROXYUSER                  | No       | The username for authenticating to the proxy server. |
+| PROXYPASSWORD              | No       | The password for authenticating to the proxy server. |
+| NONPROXYHOSTS              | No       | The list of hosts that the driver should connect to directly, bypassing the proxy server. Separate the hostnames with a URL-escaped pipe symbol (`%7C`). |
```
sfc-gh-dyoshinaga, line 132: Is it correct that the pipe symbol should be URL-escaped?

sfc-gh-dyoshinaga: Does this support using an asterisk (`*`) as a wildcard?

README.md
```diff
@@ -264,6 +270,19 @@ If you are using a different method for authentication, see the examples below:
   * `{okta_url_endpoint}` is the URL for the endpoint for your Okta account (e.g. `https://<okta_account_name>.okta.com`).
   * `{login_name_for_IdP}` is your login name for your IdP.
 
+The following example configures the driver to connect through the proxy server `myproxyserver` on port `8888`. The driver
+authenticates to the proxy server as the user `test` with the password `test`:
+
+```cs
+using (IDbConnection conn = new SnowflakeDbConnection())
+{
+    conn.ConnectionString = "account=testaccount;user=testuser;password=XXXXX;db=testdb;schema=testschema;useProxy=true;proxyHost=myproxyserver;proxyPort=8888;proxyUser=test;proxyPassword=test";
```
sfc-gh-dyoshinaga, line 279: Is this example correct?

README.md
```diff
@@ -287,6 +306,9 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+Note that for a `TIME` column, the reader returns a `System.DateTime` value. If you need a `System.TimeSpan` column, call the
+`getTimeSpan` method in `SnowflakeDbDataReader`.
+
```
sfc-gh-dyoshinaga, line 311: Is there an example of this that I can add here?

README.md
```diff
@@ -339,6 +361,13 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+Close the Connection
+--------------------
+
+To close the connection, call the `Close` method of `SnowflakeDbConnection`.
+
+If you want to avoid blocking threads while the connection is closing, call the `CloseAsync` method instead.
+
```
sfc-gh-dyoshinaga, line 370: Is there an example of this that I can add here?

README.md
```diff
@@ -339,6 +361,13 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+Close the Connection
+--------------------
+
+To close the connection, call the `Close` method of `SnowflakeDbConnection`.
+
+If you want to avoid blocking threads while the connection is closing, call the `CloseAsync` method instead.
```
sfc-gh-dyoshinaga, line 369: Do we need to explain the CancellationToken argument?
Is it this?
https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-5.0

README.md
```diff
@@ -123,7 +123,13 @@ The following table lists all valid connection properties:
 | PRIVATE_KEY_PWD            |No        | The passphrase to use for decrypting the private key, if the key is encrypted.|
 | PRIVATE_KEY                |Depends   | The private key to use for key-pair authentication. Must be used in combination with AUTHENTICATOR=snowflake_jwt. <br /> If the private key value includes any equal signs (=), make sure to replace each equal sign with two signs (==) to ensure that the connection string is parsed correctly.|
 | TOKEN                      |Depends   | The OAuth token to use for OAuth authentication. Must be used in combination with AUTHENTICATOR=oauth.|
-| INSECUREMODE               |No   	    | Set to true to disable the certificate revocation list check. Default is false.|
+| INSECUREMODE               |No   	| Set to true to disable the certificate revocation list check. Default is false.|
+| USEPROXY                   | No       | Set to true if you need to use a proxy server. The default value is false. |
+| PROXYHOST                  | No       | The hostname of the proxy server. |
```
sfc-gh-dyoshinaga, line 128: Do these settings and the new methods apply to the 2.x version of the driver?

README.md
SimbaGithub: So, for '%7c', I did this based on the implementation for JDBC. But it is not necessary, if the user passes a '|' character, I've tested it and it's Ok (same for JDBC as well).

Yes, the wildcard * is supported

README.md
SimbaGithub: yes, it's correct

README.md
SimbaGithub: Yes, here is the example : 

`TimeSpan timeSpanTime = ((SnowflakeDbDataReader)reader).GetTimeSpan(13);`

Can you also add in that this is not a generic IDataReader method, which means that the reader must be a SnowflakeDbDataReader to be able to use this new method. This is shown in the example above where I cast the reader.

README.md
SimbaGithub: CancellationToken is a well know async parameter. It give the user a token which can be used to cancel an asynchronous method if needed. I think it's Ok and it doesn't need an extra explanation.

Now, since this is another method which is not readily available in IDbConnection, the user need to make sure that he is using a SnowflakeDbConnection.

README.md
SimbaGithub: Here is an example

```
CancellationTokenSource cancellationTokenSource  = new CancellationTokenSource();
// Close the connection
((SnowflakeDbConnection)conne).CloseAsync(cancellationTokenSource.Token);
```

README.md
sfc-gh-dyoshinaga: Thanks! I changed this to use a `|` symbol and mentioned the asterisk as well.

README.md
sfc-gh-dyoshinaga: Thanks! I added the example and mentioned that you need to cast before calling the method.

README.md
sfc-gh-dyoshinaga: Thanks! I added this

README.md
sfc-gh-dyoshinaga: Thanks! I added a note to tell readers to cast before calling the method

README.md
sfc-gh-dyoshinaga: @sfc-gh-abhatnagar Is this upcoming release number going to be 2.0.4?

README.md
```diff
@@ -123,7 +123,13 @@ The following table lists all valid connection properties:
 | PRIVATE_KEY_PWD            |No        | The passphrase to use for decrypting the private key, if the key is encrypted.|
 | PRIVATE_KEY                |Depends   | The private key to use for key-pair authentication. Must be used in combination with AUTHENTICATOR=snowflake_jwt. <br /> If the private key value includes any equal signs (=), make sure to replace each equal sign with two signs (==) to ensure that the connection string is parsed correctly.|
 | TOKEN                      |Depends   | The OAuth token to use for OAuth authentication. Must be used in combination with AUTHENTICATOR=oauth.|
-| INSECUREMODE               |No   	    | Set to true to disable the certificate revocation list check. Default is false.|
+| INSECUREMODE               |No   	| Set to true to disable the certificate revocation list check. Default is false.|
+| USEPROXY                   | No       | Set to true if you need to use a proxy server. The default value is false. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYHOST                  | No       | The hostname of the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
```
SimbaGithub, line 128: @sfc-gh-dyoshinaga I think you might want to change required 'No' to 'Depends' for ProxyHost and ProxyPort and explain that this is required if USEPROXY is true.

README.md
```diff
@@ -123,7 +123,13 @@ The following table lists all valid connection properties:
...
+| PROXYPORT                  | No       | The port number of the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYUSER                  | No       | The username for authenticating to the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
+| PROXYPASSWORD              | No       | The password for authenticating to the proxy server. <br/> <br/> This parameter was introduced in v2.0.4. |
```
SimbaGithub, line 131: Here also, this is required if USEPROXY is true and PROXYUSER is provided

README.md
sfc-gh-dyoshinaga: Thanks! Done.

README.md
sfc-gh-dyoshinaga: Thanks! Done

SimbaGithub (approved):

PR #357, opened 2021-08-04
Logging and SecretDetector revisions

SimbaGithub:


Snowflake.Data.Tests/SecretDetectorTest.cs
```diff
@@ -87,51 +87,91 @@ public void TestAWSKeys()
 
             // aws_key_id with double quotes on key
             BasicMasking(@"""aws_key_id"":'aaaaaaaa'", @"""aws_key_id"":'****'");
+
+            //If attribute is enclose in simple or double quote
```
sfc-gh-hchaturvedi, line 91: nit: typo: `single`

Snowflake.Data.Tests/SecretDetectorTest.cs
```diff
@@ -87,51 +87,91 @@ public void TestAWSKeys()
...
+            BasicMasking(@"'aws_key_id'='aaaaaaaa'", @"'aws_key_id'='****'");
+            BasicMasking(@"""aws_key_id""='aaaaaaaa'", @"""aws_key_id""='****'");
```
sfc-gh-hchaturvedi, line 93: so the key is in double quotes but the value is in single quotes?

sfc-gh-hchaturvedi (approved):
Minor comments .. Rest LGTM.

sfc-gh-hyu (approved):

Snowflake.Data.Tests/SecretDetectorTest.cs
SimbaGithub: The test is based on the pattern for AWS key :         private static readonly string AWS_KEY_PATTERN = @"(aws_key_id|aws_secret_key|access_key_id|secret_access_key)('|"")?(\s*[:=]\s*)'([^']+)'";

I think Louie found a case when the key was in " and that's why he added it. We might need to add " for the value. I'll see if I can find the log line which was related.

sfc-gh-hchaturvedi: I think this is related to the request I made while he was fixing the nodejs driver. It's fine I remember now. There was a leak scare a while ago where the AWS Secret KEY showed up in the logs. The request was in response to that.

PR #360, opened 2021-08-12
SNOW-172987 - COPY INTO error

SimbaGithub:
Copy into advance the cursor to get the row update information but does not rewind the cursor breaking result set retrieval afterwards.

### DIFF
--- a/Snowflake.Data.Tests/SFDbDataReaderIT.cs
+++ b/Snowflake.Data.Tests/SFDbDataReaderIT.cs
@@ -1178,6 +1178,54 @@ namespace Snowflake.Data.Tests
         }
 
         [Test]
+        public void TestCopyCmdResultSet()
+        {
+            using (IDbConnection conn = new SnowflakeDbConnection())
+            {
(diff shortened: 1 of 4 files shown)


sfc-gh-hyu (commented):
Can you add a test?

alexrosenfeld10:
as a user, I'm curious what the impact of this bug in existing state is? Please inform.

SimbaGithub:
Good point @sfc-gh-hyu, I have added a test for executing COPY INTO as query and reading the first row. The existing test was only executing COPY INTO as DML.

@alexrosenfeld10 : The issue here is that when trying to read data from a COPY INTO result set (ExecuteReader instead of ExecuteNonQuery), it would fail because the cursor was moved after the first and only row to get the data in the "rows_loaded" cell and use it for the update count.

alexrosenfeld10:
Ah, ok. So the actual copying of the data is bug-free. Great, thanks

sfc-gh-hyu (approved):

PR #387, opened 2021-10-26
Rethrow top level Exception instead of InnerException

Coder3333:
I am having trouble debugging with the original code, because it trims out the top level Exception, and instead only re-throws or logs the InnerException.  I have modified the code to retain the original top level Exception for re-throws and logging.


Coder3333:
These changes have been addressed in other, more recent PR's, so I am cancelling this one.

PR #399, opened 2021-11-30
SF DB Command for query Id

abhirob85:


sfc-gh-igarish:
I am not finding issue# associated with PR. Please create an issue and then raise the PR. as such we are planning similar change in our upcoming release.

PR #430, opened 2022-03-15
snow-380 00236853: .NET connection pooling request

SimbaGithub:
I created a static Concurrent Dictionary for the connection pool. And updated those open and close functions for the connection pool feature. Therefore the old test case I have to change the connection status from close to open, after connection close, because that connection remains at the connection pool.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -26,7 +26,7 @@ namespace Snowflake.Data.Client
 
         internal int _connectionTimeout;
 
-        internal bool isActive = false;
+        internal int _refCount = 0;
 
         private bool disposed = false;
 
@@ -70,7 +70,7 @@ namespace Snowflake.Data.Client
             this.SfSession = conn.SfSession;
             this._connectionState = conn._connectionState;
             this._connectionTimeout = conn._connectionTimeout;
-            this.isActive = conn.isActive;
+            this._refCount = conn._refCount;
             this.disposed = conn.disposed;
             this.ConnectionString = conn.ConnectionString;
             this.Password = conn.Password;
@@ -146,7 +146,7 @@ namespace Snowflake.Data.Client
             SnowflakeDbConnection conn = SnowflakeDbConnectionPool.getConnection(this.ConnectionString, this.Password);
             if (conn != null)
             {
-                conn.isActive = false;
+                conn._refCount--;
             }
             else
             {
@@ -159,7 +159,7 @@ namespace Snowflake.Data.Client
             SnowflakeDbConnection conn = SnowflakeDbConnectionPool.getConnection(this.ConnectionString, this.Password);
             if (conn != null)
             {
-                conn.isActive = false;
+                conn._refCount--;
                 return Task.CompletedTask;
             }
             else
@@ -218,7 +218,7 @@ namespace Snowflake.Data.Client
             SnowflakeDbConnection conn = SnowflakeDbConnectionPool.getConnection(this.ConnectionString, this.Password);
             if(conn != null)
             {
-                conn.isActive = true;
+                conn._refCount++;
                 this.copy(conn);
             }
             else
@@ -234,6 +234,7 @@ namespace Snowflake.Data.Client
             SetSession();
             try
             {
+                _refCount++;
                 SfSession.Open();
             }
             catch (Exception e)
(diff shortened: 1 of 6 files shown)


Grauenwolf:
While we certainly do need connection pooling, don't like this design. 

With ADO.NET connection pooling, you don't pool the DbConnection object itself. Instead you pool the underlying connect that the DbConnection object wraps.

Calling SnowflakeDbConnection.Dispose should...

1. Set SnowflakeDbConnection.State to closed.
2. Return the SFSession to the pool (assuming it isn't in a faulted state)

When SnowflakeDbConnection.Open is called

1. Fetch a SFSession from the pool
2. Verify the SFSession is still viable
3. If 1 or 2 fails, create a new SFSession

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -24,8 +26,155 @@ public class SnowflakeDbConnection : DbConnection
...
         internal int _connectionTimeout;
 
+        private bool isActive = false;
+
         private bool disposed = false;
 
+        /// <summary>
+        ///     Connection pool 
+        /// </summary>
+        private static ConcurrentDictionary<Tuple<string, SecureString>, SnowflakeDbConnection> connectionPool;
```
sfc-gh-cshi, line 36: Do not feel it is a good idea to have connection pool info and logic inside SnowflakeDbConnection. There should be a ConnectionPool to wrap and hide the pool related logic. Such as https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Core/ConnectionPool.cs

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -24,8 +26,155 @@ public class SnowflakeDbConnection : DbConnection
...
+                    conn.CloseConnection();
+                    conn.Dispose(false);
+                    curSize--;
+                }
+            }
+        }
+
+        private static SnowflakeDbConnection getConnection(string connStr, SecureString pw)
+        {
+            var connKey = Tuple.Create(connStr, pw);
```
sfc-gh-cshi, line 79: The password might not need to be part of the key.

Snowflake.Data/Client/SnowflakeDbConnection.cs
Grauenwolf: As long as the connection pool is hidden from me, the user of the DbConnection class, I couldn't care less where the logic lives.

I just need to know that pooling is happening somewhere so that we don't leak http clients.

Snowflake.Data/Client/SnowflakeDbConnection.cs
Grauenwolf: I have seen systems where multiple users could have the same username, but with different passwords. 

Snowflake better not be doing that.

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -100,9 +141,36 @@ public override void Close()
             _connectionState = ConnectionState.Closed;
         }
 
+        public override void Close()
```
sfc-gh-cshi, line 144: is Close() multi thread safe?

Grauenwolf: DBConnection subclasses aren't expected to be threadsafe.

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -100,9 +141,36 @@ public override void Close()
             _connectionState = ConnectionState.Closed;
         }
 
+        public override void Close()
+        {
+            SnowflakeDbConnection conn = SnowflakeDbConnectionPool.getConnection(this.ConnectionString, this.Password);
+            if (conn != null)
+            {
+                conn._refCount--;
```
sfc-gh-cshi, line 149: I am afraid the update of _refCount needs to be protected, like Interlocked.increment/decrement?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -100,9 +141,36 @@ public override void Close()
...
+            SnowflakeDbConnection conn = SnowflakeDbConnectionPool.getConnection(this.ConnectionString, this.Password);
+            if (conn != null)
+            {
+                conn._refCount--;
+            }
+            else
+            {
+                CloseConnection();
+            }
+        }
```
sfc-gh-cshi, line 155: This Close() is not actually doing anything turning off this SnowflakeDbConnection. What if users accidentally run multiple time close() on the same connection object?

SimbaGithub:
Please go to the new PR here https://github.com/snowflakedb/snowflake-connector-net/pull/455

PR #431, opened 2022-03-15
snow-502 00374258 DEL character issue

SimbaGithub:
bug fix at ReusableChunkParser.cs and I created a test case for this.

### DIFF
--- a/Snowflake.Data.Tests/SFDelCharT.cs
+++ b/Snowflake.Data.Tests/SFDelCharT.cs
@@ -1,0 +1,157 @@
+﻿/*
+ * Copyright (c) 2012-2019 Snowflake Computing Inc. All rights reserved.
+ */
+
+using System;
+using System.Linq;
+using System.Data.Common;
+using System.Data;
+using System.Globalization;
+using System.Text;
+namespace Snowflake.Data.Tests
+{
+    using Snowflake.Data.Client;
+    using Snowflake.Data.Core;
+    using NUnit.Framework;
+
+    [TestFixture]
+    class SFDelCharT : SFBaseTest
+    {
+
+        [Test]
+        public void testDelChar()
+        {
+            using (IDbConnection conn = new SnowflakeDbConnection())
+            {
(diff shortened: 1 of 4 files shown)


Snowflake.Data.Tests/SFDelCharT.cs
```diff
@@ -0,0 +1,157 @@
...
+                    Assert.AreEqual(1, insertReader.RecordsAffected);
+                }
+                */
+                string selectCommand = "select * from deltest1";
+                cmd.CommandText = selectCommand;
+                cmd.CommandType = System.Data.CommandType.Text;
+
+                var sb = new StringBuilder();
+                var count = 0;
+                using (var sw = new System.IO.StreamWriter(@"D:\Snowflake\Log\test.txt", true, Encoding.UTF8, 4096))
```
sfc-gh-cshi, line 117: absolute path?

Snowflake.Data/Core/ReusableChunkParser.cs
```diff
@@ -123,6 +123,9 @@ public async Task ParseChunk(IResultChunk chunk)
                                     break;
                                 case -1:
                                     throw new SnowflakeDbException(SFError.INTERNAL_ERROR, $"Unexpected end of stream in escape sequence");
+                                default:
```
sfc-gh-cshi, line 126: Do we need to handle other escape case as well? Make sure we are good with all these cases: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#string-escape-sequences

SimbaGithub: it already cover all escape cases.

Snowflake.Data.Tests/SFDelCharT.cs
SimbaGithub: I updated

sfc-gh-cshi (approved):
LGTM. comments on test range 35000 and 4000 should be helpful

PR #438, opened 2022-04-06
Implement GET Command

SimbaGithub:
Compression option tests are currently not included in this PR

### DIFF
--- a/Snowflake.Data.Tests/SFPutGetTest.cs
+++ b/Snowflake.Data.Tests/SFPutGetTest.cs
@@ -34,6 +34,9 @@ namespace Snowflake.Data.Tests
             const string TABLE_STAGE = "TABLE_STAGE";
             const string NAMED_STAGE = "NAMED_STAGE";
 
+            const string FALSE_COMPRESS = "FALSE";
+            const string TRUE_COMPRESS = "TRUE";
+
             const string UPLOADED = "UPLOADED";
             const string DOWNLOADED = "DOWNLOADED";
 
@@ -56,7 +59,6 @@ namespace Snowflake.Data.Tests
 
             string copyIntoTable = $"COPY INTO {TEST_TEMP_TABLE_NAME}";
             string copyIntoStage = $"COPY INTO {TEST_TEMP_TABLE_NAME} FROM @{DATABASE_NAME}.{SCHEMA_NAME}.{TEST_TEMP_STAGE_NAME}";
-            string copyIntoUser = $"COPY INTO {TEST_TEMP_TABLE_NAME} FROM @~/";
 
             string removeFile = $"REMOVE @{DATABASE_NAME}.{SCHEMA_NAME}.%{TEST_TEMP_TABLE_NAME}";
             string removeFileUser = $"REMOVE @~/";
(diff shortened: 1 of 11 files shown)


Snowflake.Data.Tests/SFPutGetTest.cs
```diff
@@ -62,13 +61,19 @@ public void TestPutCommand(string compressionType)
                 // Write row data to temp file
                 File.WriteAllText(filePath, ROW_DATA);
 
-                string putQuery = $"PUT file://${filePath} @{DATABASE_NAME}.{SCHEMA_NAME}.%{TEST_TEMP_TABLE_NAME}";
+                string tempDirectory = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
sfc-gh-cshi, line 64: PUT and GET test should cover PUT/GET to stage other than table. Also compression option tests are missing. If these are not addressed in this PR, PR summary should be updated accordingly.

SimbaGithub: I've added tests for the other staging types and updated the summary

Snowflake.Data.Tests/SFPutGetTest.cs
```diff
@@ -47,97 +52,157 @@ public void TestPutCommand(string compressionType)
             string createTable = $"create or replace table {TEST_TEMP_TABLE_NAME} ({COL1} STRING," +
             $"{COL2} STRING," +
             $"{COL3} STRING)";
-            string copyInto = $"COPY INTO {TEST_TEMP_TABLE_NAME}";
+            string createStage = $"create or replace stage {TEST_TEMP_STAGE_NAME}";
```
sfc-gh-cshi, line 55: For stage creating, please refer to https://docs.snowflake.com/en/sql-reference/sql/create-stage.html, include tests cover both options: ENCRYPTION = (TYPE = 'SNOWFLAKE_FULL' | TYPE = 'SNOWFLAKE_SSE') . Since sometime dot net code might have to work on already created stages.

Snowflake.Data.Tests/SFPutGetTest.cs
```diff
@@ -47,97 +52,157 @@ public void TestPutCommand(string compressionType)
...
+                    else if (stageType == TABLE_STAGE)
                     {
-                        Assert.AreEqual(reader.GetString(0), COL1_DATA);
-                        Assert.AreEqual(reader.GetString(1), COL2_DATA);
-                        Assert.AreEqual(reader.GetString(2), COL3_DATA);
+                        putQuery = $"PUT file://{filePath} @{DATABASE_NAME}.{SCHEMA_NAME}.%{TEST_TEMP_TABLE_NAME}";
                     }
+                    else if (stageType == NAMED_STAGE)
+                    {
+                        putQuery = $"PUT file://{filePath} @{DATABASE_NAME}.{SCHEMA_NAME}.{TEST_TEMP_STAGE_NAME}";
```
sfc-gh-cshi, line 94: For PUT/GET, should test if GET can work with both AUTO_COMPRESS = TRUE | FALSE following https://docs.snowflake.com/en/sql-reference/sql/put.html

Snowflake.Data/Core/FileTransfer/EncryptionProvider.cs
```diff
@@ -129,6 +129,95 @@ static private byte[] encryptFileKey(byte[] masterKey, byte[] unencryptedFileKey
...
+            Aes aes = Aes.Create();
+            aes.Key = key;
+            aes.Mode = CipherMode.CBC;
+            aes.Padding = PaddingMode.PKCS7;
+            aes.IV = iv;
+
+            MemoryStream targetStream = new MemoryStream();
+            CryptoStream cryptoStream = new CryptoStream(targetStream, aes.CreateDecryptor(), CryptoStreamMode.Write);
+
+            byte[] inFileBytes = File.ReadAllBytes(inFile);
```
sfc-gh-cshi, line 216: ReadAllBytes() at once might not be not a good idea when it is a big file. Maybe using FileStream and ReadAsync when it is large (e.g. hundreds MB to dozens GB) could help for memory consumption and performance, please test with large files to make sure the performance is acceptable and comparable with snowsql.

sfc-gh-cshi: The same comment apply to CreateEncryptedBytes. I am ok if we handle large file and performance improvement later. But I would like to have some numbers to help PM and us to understand the performance handling large files and how it compare with with snowsql's PUT and GET. So we could estimate the priority of the performance improvement.

sfc-gh-cshi (approved):
LGTM. the block size for big files could be configurable.

PR #439, opened 2022-04-08
Exception best practices

AmmonRoberts:
Changed exception throwing to better follow best practices to preserve stack trace.

### DIFF
--- a/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
+++ b/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
@@ -732,7 +732,7 @@ namespace Snowflake.Data.Core
             }
             catch (Exception ex)
             {
-                Console.WriteLine("Unhandled exception while uploading file.", ex);
+                Logger.Debug("Unhandled exception while uploading file.", ex);
                 throw;
             }
             finally
--- a/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
+++ b/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
@@ -90,7 +90,7 @@ namespace Snowflake.Data.Core
         /// The type of transfer either UPLOAD or DOWNLOAD.
         /// </summary>
         private readonly CommandTypes CommandType;
-        
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -732,7 +732,8 @@ private ISFRemoteStorageClient renewExpiredClient()
             }
             catch (Exception ex)
             {
-                throw ex;
+                Console.WriteLine("Unhandled exception while uploading file.", ex);
```
Grauenwolf, line 735: Shouldn't this be writing to `Logger`, not `Console`?

AmmonRoberts: It most definitely should! Let me fix that real fast, thanks!

Grauenwolf:
Let's see if they actually pick this up. Given how old some rather obvious PRs are, I'm tempted to to give up and create my own fork of the connector.

alexrosenfeld10:
@Grauenwolf I have had private fork of this for quite some time. This repo doesn't get much attention. I've found the only / best way to get changes through if you really need them is to open a snowflake support case...

Coder3333:
I just submitted a PR without realizing you had one along the same lines.  I took it further by fixing the ToString method of SnowflakeDbException, which currently hides the stack trace.  I really wish someone would take these PR's seriously; otherwise, why is this thing open source?

https://github.com/snowflakedb/snowflake-connector-net/pull/452

Grauenwolf:
@alexrosenfeld10 @Coder3333 I created a public fork at https://github.com/TortugaResearch/Tortuga.Data.Snowflake/ 

It's a major refactor job, borderline a rewrite. But I think it puts the connector into a state where we can actually build upon it with confidence.

When you aren't busy, take a look at let me know what you think.

sfc-gh-igarish:
@AmmonRoberts Changes look good. Do you mind to create an issue and then raise the PR? Sorry for late response. We are going through all PR  right now.

sfc-gh-igarish (approved):
LGTM

sfc-gh-igarish:
You can merge it. Please create an issue for a PR. So we can track it.

PR #447, opened 2022-04-26
snow-561 add option to disable retry

SimbaGithub:
use "disableretry=true" on connection string to disable retry


sfc-gh-cshi (commented):
LGTM. Would like to have a test case to cover this option.

sfc-gh-cshi (approved):
Approve for now. Test coverage is still needed.

PR #450, opened 2022-04-27
allow case insensitive fallback ordinal search SFResultSetMetaData

AlbrechtCatface:
allow case insensitive fallback ordinal search


sfc-gh-igarish:
Not finding associated issue#. Please create an issue# and then raise the PR.

PR #452, opened 2022-04-29
Correct Exception Usage

Coder3333:
The SnowflakeDbException.ToString method was hiding the full details that one would expect from an Exception, including the stack trace.  This has been refactored to include the details, while retaining the custom information previously included in the Message.

Additionally, cleaned up some code that was miss using Exceptions in a way that lost exception details.


sfc-gh-sghosh:
Hi Team,

Please merge the code after approval, its pending from long, customer asking for it.

Thanks,
Sujan

Coder3333:
It looks like one of the changes was missed when this PR was merged into Master.  I have updated this PR from Master, and you can see the change in KeyPairAuthenticator.cs that also needs to be included.

Coder3333:
It looks like you copied my PR into new PR's, so I will close this one.

https://github.com/snowflakedb/snowflake-connector-net/pull/459
https://github.com/snowflakedb/snowflake-connector-net/pull/474

PR #461, opened 2022-05-18
SNOW-592041-disable-retry

sfc-gh-mheavin:
Sorted the table and added a new DISABLERETRY row for the new connection setting.

DISABLERETRY   N.  Disables the driver from automatically reconnecting when a connection fails or drops. Default: false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,28 +114,28 @@ The following table lists all valid connection properties:
 |----------------------------|----------|-------------------------------------------------------------------------------|
 | ACCOUNT                    | Yes      | Your full account name might include additional segments that identify the region and cloud platform where your account is hosted	|
 | APPLICATION                | No       | **_Snowflake partner use only_**: Specifies the name of a partner application to connect through .NET. The name must match the following pattern:  ^\[A-Za-z](\[A-Za-z0-9.-]){1,50}$ (one letter followed by 1 to 50 letter, digit, .,- or, \_ characters).   |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,27 +114,28 @@ The following table lists all valid connection properties:
...
+| AUTHENTICATOR              | No       | The method of authentication. Currently supports the following values: <br /> - snowflake (default): You must also set USER and PASSWORD. <br /> - [the URL for native SSO through Okta](https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-use.html#native-sso-okta-only): You must also set USER and PASSWORD. <br /> - [externalbrowser](https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-use.html#browser-based-sso): You must also set USER. <br /> - [snowflake_jwt](https://docs.snowflake.com/en/user-guide/key-pair-auth.html): You must also set PRIVATE_KEY_FILE or PRIVATE_KEY. <br /> - [oauth](https://docs.snowflake.com/en/user-guide/oauth.html): You must also set TOKEN.
+| CONNECTION_TIMEOUT         | No       | Total timeout in seconds when connecting to Snowflake. Default to 120 seconds |
 | DB                         | No       |                                                                               |
+| DISABLERETRY               | No       | Disables the driver from automatically reconnecting when a connection fails or drops. Default: false. |
```
sfc-gh-dyoshinaga, line 120: For boolean parameters, we should probably explain this in terms of when you would set this. e.g. -- as is done with INSECUREMODE --

```
Set this to true to prevent the driver from reconnecting automatically when the connection fails or drops. The default value is `false`.
```

PR #462, opened 2022-05-29
Fix for PR431 regression with backslashes

sfc-gh-kterada:
- Correct regression with extra backslashes added due to #431
- Consolidate standalone unit tests for #431 into SFReusableChunkTest


sfc-gh-hyu (approved):

sfc-gh-kterada:
@sfc-gh-hyu, would you mind reviewing the latest commit to this branch to correct the GCS-specific issues with unit test failures when run on the 4.7.2 framework runtime?

sfc-gh-hyu (approved):

PR #480, opened 2022-07-08
Add test case for CALL statement type

SimbaGithub:
Based on SF# 00383533

### DIFF
--- a/Snowflake.Data.Tests/SFStatementTest.cs
+++ b/Snowflake.Data.Tests/SFStatementTest.cs
@@ -6,9 +6,6 @@ namespace Snowflake.Data.Tests
 {
     using Snowflake.Data.Core;
     using NUnit.Framework;
-    using System.Data;
-    using Snowflake.Data.Client;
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/SFStatementTest.cs
```diff
@@ -6,12 +6,15 @@ namespace Snowflake.Data.Tests
...
+    using System.Data;
+    using Snowflake.Data.Client;
+    using System.Data.Common;
 
     /**
      * Mock rest request test
      */
     [TestFixture]
-    class SFStatementTest
+    class SFStatementTest : SFBaseTest
```
sfc-gh-hyu, line 17: I think original tests is more like a unit tests, where it mock a bunch of object. However, when extending a testsuite from SFBaseTest, it normally requires establishing connection with server. I would suggest put this tests in a testsuite that is already extend from SFBaseTest or create a new test fixture.

sfc-gh-hyu (approved):

PR #482, opened 2022-07-13
This should be public, not internal.

### LINKED ISSUE #481: Cannot configure connections pool as SnowflakeDbConnectionPool class has internal access modifier
### Issue description
According to documentation 
https://github.com/snowflakedb/snowflake-connector-net#using-connection-pools
it should be possible to configure connection pool.
But since the class SnowflakeDbConnectionPool is internal - the class is not visible to code of an application.

### Configuration
Driver version: 2.0.14

lee-chrisbell:
Closes #481

### DIFF
--- a/Snowflake.Data.Tests/HttpUtilTest.cs
+++ b/Snowflake.Data.Tests/HttpUtilTest.cs
@@ -1,0 +1,39 @@
+﻿/*
+ * Copyright (c) 2022 Snowflake Computing Inc. All rights reserved.
+ */
+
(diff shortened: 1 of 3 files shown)


jeff-smith-dvn (approved):
I have been using this library for connectivity and just recently upgraded and noticed the updates for connection pooling which I would love to use.  When reviewing it I realized the same as original issue creator that it can't be set because of the class not being PUBLIC.

vuder (approved):

PR #484, opened 2022-07-14
connection pool default disable, add enable/disable for connection pool

SimbaGithub:


Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -17,6 +18,7 @@ class SnowflakeDbConnectionPool
         private static long timeout;
         private const int MAX_POOL_SIZE = 10;
         private const long TIMEOUT = 3600;
+        private static bool pooling = false;
```
sfc-gh-hchaturvedi, line 21: Please set pooling back to the original default value. I am assuming that it was enabled by default. Any deviation from original default values constitute a behavior change.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -105,6 +108,36 @@ public void TestConnectionPoolIsFull()
...
+            var conn2 = new SnowflakeDbConnection();
+            conn2.ConnectionString = ConnectionString;
+            conn2.Open();
+            conn2.Close();
+            var conn3 = new SnowflakeDbConnection();
+            conn3.ConnectionString = ConnectionString;
+            conn3.Open();
+            conn3.Close();
+
+            Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize());
```
sfc-gh-igarish, line 135: Why do you expect 1 here?

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -145,5 +151,15 @@ public static int GetCurrentPoolSize()
         {
             return connectionPool.Count;
         }
+
+        public static void SetPooling(bool isEnable)
+        {
+            pooling = isEnable;
```
sfc-gh-igarish, line 157: what is expected behavior when pool was enabled and now user set it disable? Do we need to clear the pooled connections or throw an exception?

Snowflake.Data.Tests/SFConnectionPoolT.cs
SimbaGithub: Since those connections use the same connection string. please look at line 120, 127, 131. So the same connection will reuse from the pool. And this test case is provided by the customer.

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
SimbaGithub: Yes, I think after we enable or disable the pool, we should clear it. I will update this.

sfc-gh-igarish (approved):
Looks good

Issue #497, opened 2022-08-02
Connection Pool cleanExpiredConnections throws an exception


lee-chrisbell:
### Issue description
Exception detail -
```
An Exception occured while executing rule for jobId - "1ad9b1d2-4f8c-40f1-afbd-9e69f9f69317". Exception - "System.InvalidOperationException: Collection was modified; enumeration operation may not execute.\n   at Snowflake.Data.Client.SnowflakeDbConnectionPool.cleanExpiredConnections()\n   at Snowflake.Data.Client.SnowflakeDbConnectionPool.addConnection(SnowflakeDbConnection conn)\n   at Snowflake.Data.Client.SnowflakeDbConnection.Close()\n   at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command) in /_/Dapper/SqlMapper.Async.cs:line 457\n   at Avalara.DataWatchService.Infrastructure.ObservationsDomain.RulesRepository.ExecuteRuleAsync(Rule rule, String jobId, String parameters) in /builds/opus/data-watch/DataWatchService/src/Infrastructure/Avalara.DataWatchService.Infrastructure/ObservationsDomain/RulesRepository.cs:line 54\n   at Avalara.DataWatchService.Service.Services.ObservationsDomain.RulesService.ExecuteRuleAsync(Rule rule, String jobId, String parameters) in /builds/opus/data-watch/DataWatchService/src/Service/Avalara.DataWatchService.Service.Services/ObservationsDomain/RulesService.cs:line 33\n   at Avalara.DataWatchService.Service.Services.ObservationsDomain.RulesEngine.<>c__DisplayClass8_0.<<EnqueueRuleExecutionTask>b__0>d.MoveNext() in /builds/opus/data-watch/DataWatchService/src/Service/Avalara.DataWatchService.Service.Services/ObservationsDomain/RulesEngine.cs:line 63"
```

You should make a copy of the list and iterate through it instead of changing the list you're currently enumerating.

### Configuration
*Driver version:*

*.net6

*Client OS:* Ubutntu 20.04

PR #498, opened 2022-08-02
You cannot edit a list while the list is being enumerated.

lee-chrisbell:
- You cannot edit a list during a foreach.
- We can copy the list by using .ToList()


lee-chrisbell:
closes #497

PR #503, opened 2022-08-05
Flush block after copying decrypted stream

SimbaGithub:
Fix issue where block is not flushed after decrypted stream causing the downloaded file to be incomplete

### DIFF
--- a/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
+++ b/Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
@@ -133,6 +133,11 @@ namespace Snowflake.Data.Core
         private string destStagePath = null;
 
         /// <summary>
+        /// Placeholder threshold value.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -132,6 +132,11 @@ class SFFileTransferAgent
         private string streamDestFileName = null;
         private string destStagePath = null;
 
+        /// <summary>
+        /// Placeholder threshold value.
+        /// </summary>
+        private long DATA_SIZE_THRESHOLD = 9223372036854775807;
```
sfc-gh-igarish, line 138: what's special about this number? Can you put more comments for this specific value?

SimbaGithub: The number is equivalent to max long value. I'll edit it to use C#'s max long instead for clarity

sfc-gh-igarish (approved):
LGTM

PR #519, opened 2022-09-06
remove fakeodbc

SimbaGithub:
remove fakeodbc connection parameter when multiple statements feature is unblocked on server side.


sfc-gh-igarish (approved):
LGTM

PR #525, opened 2022-09-21
remove all expired sessions on getConnection

SimbaGithub:
salesforce 00399934, fix authentication token is expired issue.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
@@ -81,6 +81,7 @@ namespace Snowflake.Data.Client
                             {
                                 conn.SfSession.close();
                             }
+                            i--;
                         }
                         else
                         {
--- a/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
@@ -68,6 +68,7 @@ namespace Snowflake.Data.Client
             }
             lock (_connectionPoolLock)
             {
+                cleanExpiredConnections(); // clean up all expired connections first.
                 for (int i = 0; i < connectionPool.Count; i++)
                 {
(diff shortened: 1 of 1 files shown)


SimbaGithub:
salesforce 00399934, fix authentication token is expired issue.

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -68,6 +68,7 @@ internal static SnowflakeDbConnection getConnection(string connStr)
             }
             lock (_connectionPoolLock)
             {
+                cleanExpiredConnections(); // clean up all expired connections first.
```
sfc-gh-igarish, line 71: Will this cause delay in getting a connection? If yes, then I suggest when connection string match in the pool, do one more check whether it's expired or not. If expired, then put in a separate list, say expire list and continue look into pool. Don't change pool count. After the loop, which are in expire list, remove those only. This way we remove expire connection in batches and not holding to get a connection from a pool to cleanup all expired connections.

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -68,14 +68,25 @@ internal static SnowflakeDbConnection getConnection(string connStr)
...
                         connectionPool.RemoveAt(i);
-                        return conn;
+                        long timeNow = DateTimeOffset.Now.ToUnixTimeSeconds();
+                        if (conn._poolTimeout <= timeNow)
+                        {
+                            if (conn.SfSession != null)
+                            {
+                                conn.SfSession.close();
+                            }
+                            return null;
```
sfc-gh-igarish, line 84: Do we have only one connection in pool for matching connection string? If not, we need logic to continue and reduce i counter as we remove one element. Think about it.

sfc-gh-igarish (approved):
LGTM. Once tests pass, you can merge.

PR #531, opened 2022-10-06
fix 00402816 escape characters issue.

SimbaGithub:
fixed the escape character issue, this follow the microsoft .net escape character logic. Which allow the password contain ; ' ''


Snowflake.Data/Core/SFSessionProperty.cs
```diff
@@ -129,101 +131,21 @@ public override int GetHashCode()
...
-                        properties.Add(p, tokens[1]);
-                    }
-                    catch (ArgumentException e)
-                    {
-                        logger.Warn($"Property {tokens[0]} not found ignored.", e);
-                    }
-                }
+                SFSessionProperty p = (SFSessionProperty)Enum.Parse(
+                            typeof(SFSessionProperty), keys[i].ToUpper());
+                properties.Add(p, values[i]);
```
sfc-gh-igarish, line 148: Check "fakeodbc" case above. It was ignore. Also when it's true it was setting SFEnvironment stuff. You need to do the same with your new logic here.

sfc-gh-igarish (changes requested):
Please check failure about fakeodbc and my comment about it.

PR #534, opened 2022-10-12
Use WinHttpHandler for requests on Windows

SimbaGithub:
Based off SF# 00396351

Allows connecting through proxy with insecuremode=false


sfc-gh-igarish (approved):
LGTM. Did you compile on Linux? Does WinHttpHandler avail for compilation? Yes, the code won't execute at runtime, but make sure this code compile on all supported platforms.

SimbaGithub:
> LGTM. Did you compile on Linux? Does WinHttpHandler avail for compilation? Yes, the code won't execute at runtime, but make sure this code compile on all supported platforms.

The github tests include building and running on Linux and Mac platforms and it successfully compiled in those tests

PR #535, opened 2022-10-14
check command injection vulnerability

SimbaGithub:
add check to prevent command injection vulnerability

### DIFF
--- a/Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs
+++ b/Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Snowflake.Data.Log;
 using Snowflake.Data.Client;
+using System.Text.RegularExpressions;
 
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs
```diff
@@ -156,6 +157,20 @@ private static HttpListener GetHttpListener(int port)
 
         private static void StartBrowser(string url)
         {
+            string regexStr = @"^http(s?)\\:\\/\\/[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z@:])*(:(0-9)*)*(\\/?)([a-zA-Z0-9\\-\\.\\?\\,\\&\\(\\)\\/\\\\\\+&%\\$#_=@]*)?$";
```
sfc-gh-igarish, line 160: why we use negative regex?

SimbaGithub: sorry, it should be !m.success, thanks for catching this. don't know why my local version is !m.success, but submit one just m.success

PR #540, opened 2022-10-25
SNOW-682118 Add samples for special characters

sfc-gh-mheavin:
Would you please review the updates for technical accuracy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The following table lists all valid connection properties:
 | APPLICATION                | No       | **_Snowflake partner use only_**: Specifies the name of a partner application to connect through .NET. The name must match the following pattern:  ^\[A-Za-z](\[A-Za-z0-9.-]){1,50}$ (one letter followed by 1 to 50 letter, digit, .,- or, \_ characters).   |
 | DB                         | No       |                                                                               |
 | HOST                       | No       | Specifies the hostname for your account in the following format: \<ACCOUNT\>.snowflakecomputing.com. <br /> If no value is specified, the driver uses \<ACCOUNT\>.snowflakecomputing.com. |
-| PASSWORD                   | Depends  | Required if AUTHENTICATOR is set to `snowflake` (the default value) or the URL for native SSO through Okta. Ignored for all the other authentication types. <br /> <br /> See [Using Special Characters in Connection Passwords](#sample-connection-strings) for examples of including special characters in connection strings.|
+| PASSWORD                   | Depends  | Required if AUTHENTICATOR is set to `snowflake` (the default value) or the URL for native SSO through Okta. Ignored for all the other authentication types.|
 | ROLE                       | No       |                                                                               |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,6 +152,55 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+<a id="sample-connection-strings"></a>
+#### Using Special Characters in Connection Passwords
+
+The .NET connector now uses Microsoft [DbConnectionStringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.data.oledb.oledbconnection.connectionstring?view=dotnet-plat-ext-6.0#remarks) to following .NET specification for escaping characters in connection strings.
```
sfc-gh-dyoshinaga, line 158: Does "now" mean from a specific version number onward?
If so, should we just include that version number? (e.g. "For version n and more recent versions")

sfc-gh-dyoshinaga, line 158: Should "to following .NET specification" be "to follow the .NET specification"?

README.md
```diff
@@ -152,6 +152,55 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+
+  ``` cs
+  string connectionString = String.Format(
+    "account=testaccount; " +
+    "user=testuser; " +
+    "password=test=password;"
+  );
+  ```
+
+  Note that previously you needed to use a double equal sign (==) to escape the character, but it is no longer required.
```
sfc-gh-dyoshinaga, line 202: What does "it" refer to?
Should this be "Prior to version n, you needed to use two equal signs (==) to specify an equal sign in the password. In version n and more recent versions, you no longer need to use two equal signs."

README.md
```diff
@@ -152,6 +152,55 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+
 If you are using a different method for authentication, see the examples below:
```
sfc-gh-dyoshinaga, line 204: You need to add a new heading here.
Otherwise, all of the content after this will be part of "Using Special Characters in Connection Passwords".

README.md
sfc-gh-mheavin: Beginning with version 2.0.18, the .NET connector uses Microsoft

sfc-gh-mheavin: Done

README.md
sfc-gh-mheavin: Note that previously you needed to use a double equal sign (==) to escape the character. However, beginning with version 2.0.18, you can use a single equal size.

README.md
sfc-gh-mheavin: After some thought, I decided to remove that heading and add some minor reorganization.

**Password-based Authentication**
blah

**Other Authentication Methods**
blah

sfc-gh-dyoshinaga (approved):
Thanks for making these changes!

sfc-gh-igarish (approved):
LGTM

PR #542, opened 2022-10-26
SNOW-684-244 Add keep alive property

sfc-gh-mheavin:
Added the CLIENT_SESSION_KEEP_ALIVE


README.md
```diff
@@ -122,6 +122,7 @@ The following table lists all valid connection properties:
 | USER                       | Yes      | If AUTHENTICATOR is set to `externalbrowser` or the URL for native SSO through Okta, set this to the login name for your identity provider (IdP).     |
 | WAREHOUSE                  | No       |                                                                               |
 | CONNECTION_TIMEOUT         | No       | Total timeout in seconds when connecting to Snowflake. Default to 120 seconds |
+| CLIENT_SESSION_KEEP_ALIVE  | No       | Whether to keep the current session active after a period of inactivity, or to force the user to login again. If the value is `true`, Snowflake keeps the session active indefinitely, even if there is no activity from the user. If the value is `false`, the user must log in again after four hours of inactivity. Default is `false`. |
```
sfc-gh-dyoshinaga, line 125: "Default is" should be "The default value is"

sfc-gh-dyoshinaga (approved):
Note that there is also a session parameter [`CLIENT_SESSION_KEEP_ALIVE`](https://docs.snowflake.com/en/sql-reference/parameters.html#client-session-keep-alive).

In the docs for the clients / connectors / drivers that support keep alive, we describe how that server parameter interacts with the client property.

e.g. For the [Python connector](https://docs.snowflake.com/en/user-guide/python-connector-api.html#connect), we say:

```
When the value is None, the CLIENT_SESSION_KEEP_ALIVE session parameter
takes precedence. To override the session parameter, pass in True or False
for this argument.
```

For [SnowSQL](https://docs.snowflake.com/en/user-guide/snowsql-start.html#k-client-session-keep-alive), we say:

```
-K , --client-session-keep-alive

Internally the parameter CLIENT_SESSION_KEEP_ALIVE is set to true for the session.
```

Finally, if the .NET Driver now also honors the `CLIENT_SESSION_KEEP_ALIVE` session parameter, we should add the .NET Driver to the list of supported clients in https://docs.snowflake.com/en/sql-reference/parameters.html#client-session-keep-alive

README.md
sfc-gh-mheavin: Done

PR #543, opened 2022-10-27
Remove cloud storage SDKs and use the REST API

SimbaGithub:
Based off SF# 00396900

Remove the cloud storage SDKs of AWS, Azure, and GCS and use their REST API for upload/download


Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -687,15 +687,14 @@ private List<string> expandFileNames(string location)
                 location = location.Replace("~", homePath);
             }
 
+            string fileName = Path.GetFileName(location);
+            string directoryName = Path.GetDirectoryName(location);
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                location = Path.GetFullPath(location);
+                location = directoryName + "\\" + fileName;
```
sfc-gh-igarish, line 695: will \\ work on Linux/Mac?

Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
```diff
@@ -176,27 +148,46 @@ public RemoteLocation ExtractBucketNameAndPath(string stageLocation)
...
 
-            GetObjectRequest request = GetFileHeaderRequest(ref client, fileMetadata);
-            GetObjectResponse response = null;
             try
             {
-                // Issue the GET request
-                var task = client.GetObjectAsync(request);
-                task.Wait();
+                // Send the request
+                HttpWebRequest request = (HttpWebRequest)HttpWebRequest.Create(uri.AbsoluteUri);
```
sfc-gh-igarish, line 171: Previously it's an async operation. Are we getting same behavior after the change?

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -38,7 +34,7 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
sfc-gh-igarish, line 37: why net 6.0 here?

sfc-gh-igarish (commented):
Are we going to do same change in other driver? Why we are moving away from SDK and use REST API? Please benchmark before and after change.

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
SimbaGithub: Yes, the connector was able to build and run tests on Linux and Mac machines as well

Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
SimbaGithub: Previously the client.GetObjectAsync() function was used but task.Wait() was synchronously blocking it. So the behaviour is similar since they both execute as synchronous

Snowflake.Data/Snowflake.Data.csproj
SimbaGithub: I've reverted it back, it was just for testing purposes

sfc-gh-ext-simba-lf:

SimbaGithub:
> Are we going to do same change in other driver? Why we are moving away from SDK and use REST API?

This was requested as some of the customer projects also referenced the cloud SDKs which caused incompatibilities in their projects.

Another reason was due to being required to install the entire SDK when only upload/download was being used so it was requested to the REST API instead.

The python connector also uses the REST API

SimbaGithub:
> Please benchmark before and after change.

I will work on benchmarking and update again with the results

sfc-gh-mcepiga:
closed after years of inactivity.

PR #545, opened 2022-10-28
Remove unneeded dispose field copy

DaneGMorris:
Copying this field from a pooled connection was causing 2 problems:
Firstly it meant that the new connection would never get disposed
Secondly a connection can only be pooled once (as the disposed path would repool the connection) 

This only occurs when connections are disposed rather than closed explicitly


DaneGMorris:

sfc-gh-pbulawa:
Thank you for your contribution. The issue has been fixed with PR #666.

PR #549, opened 2022-11-07
Mheavin snow 620438 net multi statements

sfc-gh-mheavin:


README.md
```diff
@@ -449,6 +449,71 @@ Note that because this method is not available in the generic `IDataReader` inte
...
+
+### Specifying the Number of Statements in a Batch
+
+By default, the Snowflake database expects a .NET connector application to prepare and send a single statement for execution.
+
+You can override this by specifying the number of statements in a batch for a given request or by enabling multiple statements for
+the current session.
+
+To enable multiple statements for the current session or account, alter the session or account, and set the Snowflake
+  [MULTI_STATEMENT_COUNT <label-parameters__multi_statement_count>](https://docs.snowflake.com/en/sql-reference/parameters.html#label-parameters-multi-statement-count) parameter to ``0``.
```
sfc-gh-dyoshinaga, line 484: `<label-parameters__multi_statement_count>` should not be in the link text.

README.md
```diff
@@ -449,6 +449,71 @@ Note that because this method is not available in the generic `IDataReader` inte
 TimeSpan timeSpanTime = ((SnowflakeDbDataReader)reader).GetTimeSpan(13);
 ```
 
+Executing a Batch of SQL Statements (Multi-Statement Support)
+--------------------------------------------------------------
+
+With the .NET connector, you can send
```
sfc-gh-dyoshinaga, line 455: Which version of the .NET connector supports this?

README.md
sfc-gh-dyoshinaga, line 484: Do you also plan to update https://docs.snowflake.com/en/sql-reference/parameters.html#label-parameters-multi-statement-count to include the .NET connector in the list of supported clients?

README.md
```diff
@@ -449,6 +449,71 @@ Note that because this method is not available in the generic `IDataReader` inte
...
+To enable multiple statements for the current session or account, alter the session or account, and set the Snowflake
+  [MULTI_STATEMENT_COUNT <label-parameters__multi_statement_count>](https://docs.snowflake.com/en/sql-reference/parameters.html#label-parameters-multi-statement-count) parameter to ``0``.
+
+    Use:
+
+    ```
+    alter session set MULTI_STATEMENT_COUNT = 0;
+    ```
+
+    By default, `MULTI_STATEMENT_COUNT` is set to ``1``, which indicates that only one SQL statement can be executed.
```
sfc-gh-dyoshinaga, line 492: In Markdown, when you indent a line by four spaces, you're indicating that [you want the line to be in a code block](https://daringfireball.net/projects/markdown/syntax#precode).

So all of this text now appears as [literal text in a code block (including the backticks)](https://drive.google.com/a/snowflake.com/uc?export=view&id=1UDESB_YvJcv0GlYbUlI_bTGzntrAJ6m5).

README.md
sfc-gh-mheavin: Done

README.md
sfc-gh-mheavin: With the .NET 2.0.18 and later connector, you can send

README.md
sfc-gh-mheavin: Fixed

README.md
sfc-gh-mheavin: Updated in PR: https://github.com/snowflakedb/snowflake-prod-docs/pull/1989

README.md
```diff
@@ -449,6 +449,68 @@ Note that because this method is not available in the generic `IDataReader` inte
 TimeSpan timeSpanTime = ((SnowflakeDbDataReader)reader).GetTimeSpan(13);
 ```
 
+Executing a Batch of SQL Statements (Multi-Statement Support)
+--------------------------------------------------------------
+
+With the .NET 2.0.18 and later connector, you can send
```
sfc-gh-dyoshinaga, line 455: "later connector" sounds odd. Maybe "With version 2.0.18 and later of the .NET connector".

sfc-gh-mheavin: Done

README.md
```diff
@@ -449,6 +449,68 @@ Note that because this method is not available in the generic `IDataReader` inte
...
+
+By default, the Snowflake database expects a .NET connector application to prepare and send a single statement for execution.
+
+You can override this by specifying the number of statements in a batch for a given request or by enabling multiple statements for
+the current session.
+
+To enable multiple statements for the current session or account, alter the session or account, and set the Snowflake [MULTI_STATEMENT_COUNT](https://docs.snowflake.com/en/sql-reference/parameters.html#label-parameters-multi-statement-count) parameter to ``0``, as follows.
+
+```
+alter session set MULTI_STATEMENT_COUNT = 0;
```
sfc-gh-dyoshinaga, line 486: This actually conflicts with the previous statement ("you must specify the number of statements in the batch. The Snowflake database requires the exact number of statements in order to guard against SQL injection attacks").

I think this is supposed to say that you should set MULTI_STATEMENT_COUNT to the number of statements in the batch.

README.md
```diff
@@ -449,6 +449,68 @@ Note that because this method is not available in the generic `IDataReader` inte
...
```
sfc-gh-dyoshinaga, line 486: Using ALTER SESSION here is odd, given that the example in the beginning seems to indicate that you set the number of statements via an API, not by executing the ALTER SESSION command.

i.e. the example has:

```
var param = cmd.CreateParameter();
param.ParameterName = "MULTI_STATEMENT_COUNT";
param.DbType = DbType.Int16;
param.Value = 3;
cmd.Parameters.Add(param);
```

which appears to be a way to set the MULTI_STATEMENT_COUNT parameter to 3 without having to execute ALTER SESSION.

So the instructions in this section should probably explain how to use `CreateParameter` to set the MULTI_STATEMENT_COUNT parameter.

README.md
```diff
@@ -449,6 +449,68 @@ Note that because this method is not available in the generic `IDataReader` inte
...
+- If the statements have
+  [parameters](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/statement-parameters?view=sql-server-ver15),
+  calling the [SQLNumParams](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlnumparams-function?view=sql-server-ver15)
+  function returns the total number of parameters in all the statements in the batch.
+
+- Column information about the result set (e.g. data returned by `SQLNumResultCols`, `SQLDescribeCol`, `SQLColAttribute` and
+  `SQLColAttributes`) is available when you call `SQLExecute` or `SQLExecDirect`.
+
+  Although some column information is available when you call `SQLPrepare`, the information might not be completely accurate, and
+  subsequent calls to `SQLExecute` or `SQLExecDirect` might provide more accurate information.
```
sfc-gh-dyoshinaga, line 505: I think the functions like SQLPrepare are specific to ODBC.

I think ODBC and the .NET connector have very different APIs.

I don't think you can just copy and paste the ODBC documentation into the .NET Connector documentation.

I would check with the engineers on this.

README.md
```diff
@@ -449,6 +449,68 @@ Note that because this method is not available in the generic `IDataReader` inte
 TimeSpan timeSpanTime = ((SnowflakeDbDataReader)reader).GetTimeSpan(13);
 ```
 
+Executing a Batch of SQL Statements (Multi-Statement Support)
+--------------------------------------------------------------
+
+With version 2.0.18 and later of the .NET connector, you can send
+a [batch of SQL statements](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/batches-of-sql-statements?view=sql-server-ver15)  (separated by semicolons)
```
sfc-gh-dyoshinaga, line 456: I'm not sure if you should be linking to the ODBC documentation, since ODBC and .NET have different interfaces (I think).

README.md
sfc-gh-mheavin: Not quite. I also removed the mention of account (as .NET does not support it). The idea is that you must enable the feature by setting the param to 0. Not a good design choice IMHO, but the values are 0 (multiple) or 1 (single). It has nothing to do with the specific number of statements.

README.md
sfc-gh-mheavin: I will remove the link.

README.md
sfc-gh-mheavin: The engineers approved this content, but I will touch base to verify.

sfc-gh-dyoshinaga: As an example, in .NET, it looks like `DbCommand` has a `Prepare` method (not an SQLPrepare function) for preparing SQL statements:

https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.prepare?view=net-6.0

It looks like we don't actually implement that method:

https://github.com/snowflakedb/snowflake-connector-net/blame/e9673a50a3c69bb6a2c7807cf426699b253df9df/Snowflake.Data/Client/SnowflakeDbCommand.cs#L232

README.md
sfc-gh-dyoshinaga: So this statement in line 471 is not correct:

```
The Snowflake database requires the exact number of statements
```

0 is not the exact number of statements.

sfc-gh-mheavin: It requires the exact number in the call. This parameter simply enables or disables the ability to send more than one statement. Unfortunately, it is also backwards as 0 means yes and 1 means no.

sfc-gh-dyoshinaga: So the session is created and started by the .NET connector, right?
I write code that uses the .NET connector.
How in my code do I set the parameter for the session?

As an example, [in this test](https://github.com/snowflakedb/snowflake-connector-net/blob/e9673a50a3c69bb6a2c7807cf426699b253df9df/Snowflake.Data.Tests/SFMultiStatementsIT.cs#L34), I'm opening a connection, which I think creates the session.

Where in this code would I do the equivalent of `ALTER SESSION SET MULTI_STATEMENT_COUNT = 0`?

sfc-gh-dyoshinaga: Re-reading this again, I think the problem is that there's not enough of an explanation of how to do this per request. When I skimmed this section, the most prominent thing was that I should execute ALTER SESSION somewhere.

I think we need to make it easier to see that there are two options here. I think adding an example and more of an explanation of first option would help. Otherwise, to someone skimming this section, the takeaway is that you just use ALTER SESSION to set MULTI_STATEMENT_COUNT to 0.

sfc-gh-dyoshinaga: Here's a suggested way to do this.

```
To send a batch of statements with the Snowflake .NET connector, you can either:

- Specify the number of statements in the batch, or
- Override the `MULTI_STATEMENT_COUNT` parameter for the session to indicate that each batch contains more than one statement.

### Specifying the Number of Statements in a Batch

By default, the Snowflake database expects a .NET connector application to prepare and send a single statement for execution.

You can override this by specifying the number of statements in a batch for a given request.
The Snowflake database requires the exact number of statements in order to guard against
[SQL injection](https://en.wikipedia.org/wiki/SQL_injection>) attacks. 

To specify the number of statements in the batch, set the `MULTI_STATEMENT_COUNT` parameter. For example, if there are two statements in the batch:

    var param = cmd.CreateParameter();
    param.ParameterName = "MULTI_STATEMENT_COUNT";
    param.DbType = DbType.Int16;
    param.Value = 2;
    cmd.Parameters.Add(param);

### Enabling Multiple Statements in a Session

If you don't want to specify the number of statements in each request, you can enable multiple statements for the current session by setting the Snowflake [MULTI_STATEMENT_COUNT](https://docs.snowflake.com/en/sql-reference/parameters.html#label-parameters-multi-statement-count) parameter to ``0``, as follows.

   <example of .NET connector code that sets this>
```

sfc-gh-dyoshinaga: Otherwise, the code block is the most prominent element in the section, and [it's easy to miss the fact that the code block just covers the second of two options](https://drive.google.com/a/snowflake.com/uc?export=view&id=1ysK5FoX8_DbeAkJPupTF4Pbkzdwl-fqo).

sfc-gh-mheavin:
No longer applicable

PR #558, opened 2022-11-30
reduce one version number at csproj file

SimbaGithub:
reduce one version number at csproj file, we have version and AssemblyVersion on csproj file, after this fix, we only have to update version on csproj, AssemblyVersion will use the same version number automatically.


sfc-gh-igarish (approved):
LGTM

Issue #559, opened 2022-12-05
Unable to connect from local machine due to Timeout


myshon:
### Issue description
I downloaded repository and I'm trying to connect to database on my snowflake trial account. Before I run tests I put these parameters
```json
{
  "testconnection": {
    "SNOWFLAKE_TEST_USER": "myshon323",
    "SNOWFLAKE_TEST_PASSWORD": "{mypasswordhere}",
    "SNOWFLAKE_TEST_ACCOUNT": "mt24333.eu-central-1",  
    "SNOWFLAKE_TEST_WAREHOUSE": "SANDBOX_WH",
    "SNOWFLAKE_TEST_DATABASE": "SANDBOX_DB",
    "SNOWFLAKE_TEST_SCHEMA": "PUBLIC",
    "SNOWFLAKE_TEST_ROLE": "ACCOUNTADMIN",
    "SNOWFLAKE_TEST_HOST": "mt24333.eu-central-1.snowflakecomputing.com"
  }
}
```

Because it's not clear to me what value should be provied in "TEST_ACCOUNT" I tried also with account name `mt24333` (without region) or also account identifier `MKFRWIX-OU52049` with and without organization prefix. In all cases I got timeout error: 

### Error log
Stacktrace: 
```
Snowflake.Data.Client.SnowflakeDbException : Error: Snowflake Internal Error: Unable to connect. One or more errors occurred. (Error: Request reach it...

Snowflake.Data.Client.SnowflakeDbException : Error: Snowflake Internal Error: Unable to connect. One or more errors occurred. (Error: Request reach its timeout. SqlState: , VendorCode: 270007, QueryId: ) SqlState: 08006, VendorCode: 270001, QueryId: 
  ----> System.AggregateException : One or more errors occurred. (Error: Request reach its timeout. SqlState: , VendorCode: 270007, QueryId: )
  ----> Snowflake.Data.Client.SnowflakeDbException : Error: Request reach its timeout. SqlState: , VendorCode: 270007, QueryId: 
  ----> System.Threading.Tasks.TaskCanceledException : A task was canceled.
   at Snowflake.Data.Client.SnowflakeDbConnection.Open() in C:\dev\snowflake\snowflake-connector-net\Snowflake.Data\Client\SnowflakeDbConnection.cs:line 207
```

log file: 

```
[...]

PR #561, opened 2022-12-07
Use win http handler for net framework only

myshon:
Related with issue https://github.com/snowflakedb/snowflake-connector-net/issues/559


Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -109,10 +112,19 @@ private HttpClient RegisterNewHttpClientIfNecessary(HttpClientConfig config)
             return _HttpClients[name];
         }
 
+        private bool IsNetFramework()
+        {
+#if NETFRAMEWORK
```
sfc-gh-igarish, line 117: Why conditional compilation? Does it mean we will have two versions of drivers?

myshon: Yes you have. One for NetFramework, second for Net6.

myshon: @sfc-gh-igarish I am open to better idea how to solve the issue.

lawrence-vo:
This would solve this [issue](https://github.com/snowflakedb/snowflake-connector-net/issues/583)

myshon:
Is there any chance for this PR? 😄

Snowflake.Data/Core/HttpUtil.cs
myshon: There are multiple conditional compilations in this codebase. What's the issue with that?

sfc-gh-igarish: ok. I am checking internally and get back to you ASAP.

sfc-gh-ext-simba-hx: @myshon  I think it's better to check runtime version with SFEnvironment.ClientEnv.netRuntime which is set here:
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/RestParams.cs#L52
#if NETFRAMEWORK is used there as well but it would be easier to just replace the implementation there when we have better solution of checking .net version in the future.
@sfc-gh-igarish Other than that the change looks good to me.

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -10,9 +10,12 @@
 using System.Collections.Generic;
 using Snowflake.Data.Log;
 using System.Collections.Specialized;
+using System.Linq;
+using System.Reflection;
 using System.Web;
 using System.Security.Authentication;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
```
sfc-gh-ext-simba-hx, line 18: Please remove the using added as they are not actually being used.

Snowflake.Data/Core/HttpUtil.cs
myshon: Thanks for response. I just fixed it.

Snowflake.Data/Core/HttpUtil.cs
myshon: Removed

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-hx: Thanks. All look good to me now.

sfc-gh-igarish (approved):
LGTM

sfc-gh-ext-simba-hx:
Hi @myshon ,

Currently the test can't be run in your PR and I made another one with your changes to run all tests.
https://github.com/snowflakedb/snowflake-connector-net/pull/615
It was blocked for a while due to test failures and now merged.
Thanks for your contribution.

Thanks,
Harry

Issue #557, opened 2022-11-29
CLIENT_SESSION_KEEP_ALIVE=true causes 401 response


jimhoonan:
### Issue description
When I include client_session_keep_alive=true in my connection string, I receive a 401 response when calling OpenAsync.
If I include client_session_keep_alive=false in my connection string the connection is successful

This is occurring in driver version 2.0.19
.net 6.0

alitas:
The PR #553 aims to fix this.

alitas:
> Yes, I saw that. Do you know when it will be released?

No idea. I'm waiting for a response from the maintainers too.

deepak-vanco:
release this fix please....very getting ton 401 issues

sfc-gh-igarish:
Please try latest version.

PR #566, opened 2022-12-12
fix keep alive error issue

sfc-gh-ext-simba-dl:
CLIENT_SESSION_KEEP_ALIVE=true causes 401 response #557 
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/74

### DIFF
--- a/Snowflake.Data/Core/SFSession.cs
+++ b/Snowflake.Data/Core/SFSession.cs
@@ -412,7 +412,11 @@ namespace Snowflake.Data.Core
             {
                 HeartBeatBackground heartBeatBg = HeartBeatBackground.Instance;
                 if (this.masterValidityInSeconds == 0)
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Core/SFSession.cs
```diff
@@ -411,6 +411,8 @@ internal void startHeartBeatForThisSession()
             if (!this.isHeartBeatEnabled)
             {
                 HeartBeatBackground heartBeatBg = HeartBeatBackground.Instance;
+                if (this.masterValidityInSeconds == 0)
+                    this.masterValidityInSeconds = 14400;
```
sfc-gh-igarish, line 415: Could you please put comment here? why we change 0 to 14400 and what is 14400? Please use constant name instead of value.

sfc-gh-igarish (approved):
LGTM

Issue #571, opened 2022-12-19
Error: Index was outside the bounds of the array.   at Snowflake.Data.Core.SFStatement.TrimSql


Coder3333:
### Issue description
The 2.0.19 .Net driver is throwing this "Index was outside the bounds of the array" error when I run a script that creates a stored procedure.  This same scripts run without error using the 2.0.8 version of the .Net driver with a .Net Core 3.1 application.

The procedure code being run does have a "/* */" block at the end, which I assume the TrimSql method is trying to remove.

### Error log

System.IndexOutOfRangeException: Index was outside the bounds of the array.\r\n   at Snowflake.Data.Core.SFStatement.TrimSql(String originalSql)\r\n   at Snowflake.Data.Core.SFStatement.ExecuteAsync(Int32 timeout, String sql, Dictionary`2 bindings, Boolean describeOnly, CancellationToken cancellationToken)\r\n   at Snowflake.Data.Client.SnowflakeDbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)\r\n   at

### Configuration
2.0.19 .Net 6

*Server version:* 
6.41.2

*Client OS:* Windows 10

Coder3333:
I see from the source code that the "sqlQueryLen" variable is not updated inside the do loop.  I am quite certain this is what is causing the index out of range error with "sqlQueryBuf[idx + 1]" inside the do loop.  Perhaps just moving "var sqlQueryLen = sqlQueryBuf.Length;" inside the do loop would take care of the issue.

Coder3333:
I believe the problem line must be "builder.Append(sqlQueryBuf[idx]);", as the "idx" variable can be changed before this line to a value outside the bounds of the array.

Coder3333:
I have submitted a pull request to address this bug.

https://github.com/snowflakedb/snowflake-connector-net/pull/573

sfc-gh-igarish:
Please try latest version. It's fixed.

PR #573, opened 2022-12-20
Ef trim sql bug

Coder3333:
Addresses https://github.com/snowflakedb/snowflake-connector-net/issues/571 bug, which causes an exception from TrimSql when there is a comment at the end of the statement.

System.IndexOutOfRangeException: Index was outside the bounds of the array.\r\n at Snowflake.Data.Core.SFStatement.TrimSql(String originalSql)\r\n at Snowflake.Data.Core.SFStatement.ExecuteAsync(Int32 timeout, String sql, Dictionary`2 bindings, Boolean describeOnly, CancellationToken cancellationToken)\r\n at Snowflake.Data.Client.SnowflakeDbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)\r\n at


Coder3333:
It looks like you created a new PR with my changes, so I will cancel this one.

https://github.com/snowflakedb/snowflake-connector-net/pull/581

Issue #536, opened 2022-10-16
NETFRAMEWORK preprocessor directive incorrectly applied to .Net Core 


klewelling:
### Issue description
The NETFRAMEWORK preprocessor directive should apply to Windows only, however it is being applied to .NET Core as well. 
For example https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs and https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/RestParams.cs

### Example code
The following code from RestParams.cs when compiled using .Net Core SDK on MacOS will set the netRuntime to "NETFramework" rather than "NETCore" 
```cs
#if NETFRAMEWORK
                netRuntime = "NETFramework",
                netVersion = "4.7.2",
#else
                netRuntime = "NETCore",
                netVersion ="2.0",
#endif
```

The following code form ExternalBrowserAuthenticator.cs when compiled using .Net Core SDK on MacOS will exclude the multi-platform external browser code. This causes authentication=externalbrowser to fail on MacOS (and presumably Linux) with file not found errors. 
```cs

        private static void StartBrowser(string url)
        {
            // The following code is learnt from https://brockallen.com/2016/09/24/process-start-for-urls-on-net-core/
#if NETFRAMEWORK
            // .net standard would pass here
            Process.Start(url);
#else
            // hack because of this: https://github.com/dotnet/corefx/issues/10361
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
            {
                url = url.Replace("&", "^&");
                Process.Start(new ProcessStartInfo("cmd", $"/c start {url}") { CreateNoWindow = true });
            }
            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
            {
                Process.Start("xdg-open", url);
            }
            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
            {
                Process.Start("open", url);
            }
            else
            {
                throw new SnowflakeDbException(SFError.UNSUPPORTED_PLATFORM);
            }
#endif
        }
```

### Culprit 
[...]

PR #575, opened 2023-01-03
Removed redundant 'NETFRAMEWORK' preprocessor symbol from csproj

### LINKED ISSUE #308: SNOW-352097: ExternalBrowser authentication type does not work for .NET Core projects on Windows, Linux or Mac
## Issue Description
When using ExternalBrowser authentication type, URL to the SSO provider is constructed and a browser window is opened using default OS browser. The code to start it is in Snowflake.Data.Core.Authenticator.ExternalBrowserAuthenticator.StartBrowser
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs#L156-L174
```CS
        {
            // The following code is learnt from https://brockallen.com/2016/09/24/process-start-for-urls-on-net-core/
#if NETFRAMEWORK
            // .net standard would pass here
            Process.Start(url);
#else
            // hack because of this: https://github.com/dotnet/corefx/issues/10361
            {
                url = url.Replace("&", "^&");
                Process.Start(new ProcessStartInfo("cmd", $"/c start {url}") { CreateNoWindow = true });
            }
            else
            {
                throw new SnowflakeDbException(SFError.UNSUPPORTED_PLATFORM);
            }
#endif
        }
```

The logic of the https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/Authenticator/ExternalBrowserAuthenticator.cs#L159 is incorrect. 

The `NETFRAMEWORK` preprocessor symbol is used to conditionally compile in `Process.Start(url)` for what the comments suggest are .NET Standard. 
```cs
#if NETFRAMEWORK
            // .net standard would pass here
            Process.Start(url);
```

However, per https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework, `NETFRAMEWORK` preprocessor symbol is for ".NET Framework", not 'NET Standard". 

## Looking at Result of Compilation

When compiled, the DLL in the `C:\snowflake\snowflake-connector-net\Snowflake.Data\bin\Release\net472\Snowflake.Data.dll` as well as in the public nuget package `C:\Users\Daniel\Downloads\snowflake.data.2.0.0\lib\net472\Snowflake.Data.dll` that is targeting .NET Full Framework for Windows Only desktops, which works fine on Windows:

```CS
// Snowflake.Data.Core.Authenticator.ExternalBrowserAuthenticator
using System.Diagnostics;

{
    Process.Start(url);
}
```

Looking at the DLL that targets .NET Core in `C:\snowflake\snowflake-connector-net\Snowflake.Data\bin\Release\netstandard2.0\Snowflake.Data.dll` and in `C:\Users\Daniel\Downloads\snowflake.data.2.0.0\lib\netstandard2.0\Snowflake.Data.dll`, that would be used by code running on Mac and Linux, we can see that code that could only run on Windows (and it does run beautifully there), but throws an SnowflakeDbException for Mac and Linux:

```cs
// Snowflake.Data.Core.Authenticator.ExternalBrowserAuthenticator
using Snowflake.Data.Client;
using System.Diagnostics;

{
    {
        url = url.Replace("&", "^&");
        Process.Start(new ProcessStartInfo("cmd", $"/c start {url}")
        {
            CreateNoWindow = true
        });
        return;
    }
    throw new SnowflakeDbException(SFError.UNSUPPORTED_PLATFORM);
}
```

## Possible Solution 1
This seems like an honest mistake, and can most easily be fixed by changing to correct preprocessor symbol `NETSTANDARD2_1` or better yet `NET`:
```cs
#if NET
            // .net standard would pass here
            Process.Start(url);
```

## Possible Solution 2
Better yet, change the code to explicitly address 3 different OS versions. In fact https://brockallen.com/2016/09/24/process-start-for-urls-on-net-core/ that is mentioned in comments does just that.

colgreen-payroc:
Removed redundant 'NETFRAMEWORK' preprocessor symbol from csproj; this fixes #308 and #536

NETFRAMEWORK is a standard preprocessor symbol that is already defined for builds that target .NET Framework (but not .NET Core builds); therefore defining this symbol in the csproj (with an attempt to exclude it for dotnet-core builds using a regex) is not necessary.

The regex filter was not working as expecting, causing this symbol to be defined for all builds including dotnet core builds.

See here for a list of standard preprocessor symbols: https://learn.microsoft.com/en-us/dotnet/core/tutorials/libraries#how-to-multitarget

### DIFF
--- a/Snowflake.Data/Snowflake.Data.csproj
+++ b/Snowflake.Data/Snowflake.Data.csproj
@@ -17,10 +17,6 @@
     <DebugType>Full</DebugType>
   </PropertyGroup>
   
-  <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^net\d'))">
-	<DefineConstants>NETFRAMEWORK</DefineConstants>
-  </PropertyGroup>
-	
   <ItemGroup>
     <PackageReference Include="AWSSDK.S3" Version="3.7.0.4" />
     <PackageReference Include="Google.Cloud.Storage.V1" Version="2.5.0" />


davidhcar:
consider updating the regex pattern to `  <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^net\d{2,}'))">
	<DefineConstants>NETFRAMEWORK</DefineConstants>
  </PropertyGroup>`

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -17,10 +17,6 @@
     <DebugType>Full</DebugType>
   </PropertyGroup>
   
-  <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^net\d'))">
```
davidhcar, line 20: Instead of removing the property group update the regex pattern to `'^net\d{2,}'`

davidhcar (approved):
Approving based on @colgreen-payroc confirmation https://github.com/snowflakedb/snowflake-connector-net/issues/308#issuecomment-1414050223 that the fix will work for both net471 and .net6.

davidhcar:
@colgreen-payroc You will be bumping up the version as well right?

davidhcar (approved):

colgreen-payroc:
> @colgreen-payroc You will be bumping up the version as well right?

@davidhcar I think should be done either when PR is merged, or when the next release is ready. The target branch has already been incremented from 2.0.19 to 2.0.20 since I branched, and there might other PRs in the next release besides this one.

skarpecki:
Is there any chance this will be merged to next release?

sfc-gh-igarish (approved):
LGTM

colgreen-payroc:
@sfc-gh-igarish  Do I need to do anything else? I see there is a CLA check failing.

sfc-gh-igarish:
Wait until tests finish before you merge

colgreen-payroc:
@sfc-gh-igarish  I don't have access rights to merge... "Only those with write access to this repository can merge pull requests."

sfc-gh-igarish:
We tried running test and it fails. We are trying again. Once test passes, we will merge it.

sfc-gh-ext-simba-hx:
Merge as all tests passed

PR #579, opened 2023-01-10
issue101 adding error for missing bind parameter

sfc-gh-ext-simba-dl:


Snowflake.Data/Core/SFStatement.cs
```diff
@@ -325,6 +325,11 @@ internal SFBaseResultSet Execute(int timeout, string sql, Dictionary<string, Bin
             // Trim the sql query and check if this is a PUT/GET command
             string trimmedSql = TrimSql(sql);
 
+            if(trimmedSql.Contains('?') && bindings == null)
```
sfc-gh-igarish, line 328: do we want to worry about '?' in a comment or in a literal string in the SQL statement? Also please add more description in the PR, what was the issue and how this PR will fix it?

sfc-gh-ext-simba-dl: the problem is when user use a query with ?, but they set the bind to null, so the server return "Bind variable ? not set" error message

sfc-gh-ext-simba-dl: Since this error message from server isn't like regular error message, so I catch it before sending the query to server.
I said I can repro it when my query contains ? and set the bind to null. if not null, this error will not exist.

sfc-gh-igarish: Let me give you an example: SELECT * from my_table where c1 = 'hi how are you?'; Your code changes will throw an error as this question mark is not really a parameter.

PR #581, opened 2023-01-13
Fix out of bounds error when trimming SQL

sfc-gh-ext-simba-lf:
Regarding https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/100

Based off https://github.com/snowflakedb/snowflake-connector-net/pull/573

Checks if there are still characters after a closing comment to prevent index out of bounds error when trimming the SQL query.
Also add test cases for SQL queries with line and block comments


sfc-gh-igarish (approved):
LGTM

PR #585, opened 2023-01-23
Use GCS access token for upload/download

sfc-gh-ext-simba-lf:
Regarding issue 92

Support using GCS access token for PUT/GET queries

### DIFF
--- a/Snowflake.Data.Tests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/SFDbCommandIT.cs
@@ -810,5 +810,28 @@ namespace Snowflake.Data.Tests
                 conn.Close();
             }
         }
+
+        [Test]
+        [IgnoreOnEnvIs("snowflake_cloud_env",
+                       new string[] { "AWS", "AZURE" })]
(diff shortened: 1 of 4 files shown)


sfc-gh-igarish (commented):
Did you test using GCS downscope token as other driver?

sfc-gh-ext-simba-lf:
> Did you test using GCS downscope token as other driver?

I'll add a test case similar to the other drivers

sfc-gh-ext-simba-lf:
I added two test cases similar to the ODBC and the JDBC connector

The first test case based on ODBC is about executing a large query with the GCS downscoped token: https://github.com/snowflakedb/snowflake-odbc/pull/409/files#diff-9c59c382829b5ea28ce41ba69e6a64bd9e13bd29a1197f2c996710a68fc87976R2576

The second test case based on JDBC uploads and downloads a file with the GCS downscoped token: https://github.com/snowflakedb/snowflake-jdbc/pull/409/files#diff-65666f89fd973572d96644d93d201eeb917372a9cbab1519ab487a9561c5a063R753

sfc-gh-igarish (approved):
Thanks for adding test cases. LGTM

PR #598, opened 2023-02-06
Testing Azure.Storage.Blobs dependency update

sfc-gh-ext-simba-lf:
Running tests for https://github.com/snowflakedb/snowflake-connector-net/pull/493


sfc-gh-pbulawa:
The dependency is already updated since Feb 8, 2023.

PR #599, opened 2023-02-06
support to 4.7.1

sfc-gh-ext-simba-dl:


.github/workflows/main.yml
```diff
@@ -21,7 +21,7 @@ on:
 
 env:
   DOTNET_VERSION: 6.0
-  DOTNET_LEGACY_VERSION: 4.7.2
+  DOTNET_LEGACY_VERSION: 4.7.1
```
sfc-gh-igarish, line 24: Does it mean it only work for 4.7.1 or it can work with 4.7.1 and 4.7.2 also?

sfc-gh-ext-simba-dl: we set DOTNET_LEGACY_VERSION to 4.7.2, and on readme also support 4.7.3
i believed it means the connector support 4.7.1, 4.7.2 and 4.7.3

sfc-gh-igarish (approved):
LGTM. Please check test failures on aws.

PR #611, opened 2023-02-24
handle the special case for the fips for issue 75

sfc-gh-ext-simba-dl:
handle the url contains '[' and ']'


sfc-gh-igarish (approved):
LGTM

PR #617, opened 2023-03-15
bug fix .net 249 SNOW-756325: Calling Close() before Dispose() results in duplicate connections in the pool

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/249
test case is added, if the given connection string isn't able to create connection, it will not add to the connection pool after it closed.


sfc-gh-igarish (approved):
LGTM

sfc-gh-rjoshi:
@sfc-gh-ext-simba-dl What is the ETA for this.  Will it be a part of the March release?

sfc-gh-ext-simba-dl:
this will be on april release

Issue #621, opened 2023-03-16
Excessive memory usage when uploading/downloading files


Gonyoda:
### Issue description
When uploading or downloading files, the EncryptionProvider and related code work with the file contents entirely using byte arrays.  For large files, this can cause OutOfMemory exceptions and/or cause the process to crash due to Docker memory limits.

Instead, use Streams which can use files vs memory to accomplish the same goal.

### Example code
```cs
// write 200mb file to disk
var filename = "path/to/large/file.json";
this._snowflake.NonQuery($"PUT file://{filename} @~/STAGEFOLDER");
```

### Configuration
*Driver version:* Snowflake.Data 2.0.21

*Dotnet framework and version:* .NET Core 7.0.201

*Server version:* 7.8.1

*Client OS:* Windows and/or Ubuntu

Gonyoda:
I have read the CLA Document and I hereby sign the CLA

sfc-gh-igarish:
Thanks for reporting issue. We will look into it.

sfc-gh-dszmolka:
fix is now out with release [2.1.2](https://github.com/snowflakedb/snowflake-connector-net/tree/v2.1.2)

Gonyoda:
I ran the benchmarks from my pull request #622 against your changes and here are the results:

```
|           Method |     Mean |    Error |   StdDev |    Allocated |
|----------------- |---------:|---------:|---------:|-------------:|
|   RunSmallMemory | 744.0 ms | 14.10 ms | 32.96 ms |      19.4 KB |
| RunDefaultMemory | 208.3 ms |  2.38 ms |  4.04 ms |   2062.27 KB |
|   RunLargeMemory | 197.1 ms |  2.09 ms |  1.86 ms | 859397.59 KB |
```
RunSmallMemory: MaxBytesInMemory = 1024 bytes
RunDefaultMemory: MaxBytesInMemory =1mb
RunLargeMemory: MaxBytesInMemory = 220mb

So: very nice improvement!  Thank you!

PR #622, opened 2023-03-16
Encryption streams

Gonyoda:
An effort to reduce memory usage when uploading or download files.  Stop using MemoryStream/byte[] and instead use a temporary file stream.  The temporary files are automatically closed when the stream is closed/disposed.

Can resolve issue https://github.com/snowflakedb/snowflake-connector-net/issues/621

### DIFF
--- a/Snowflake.Data.Tests/EncryptionStreamTest.cs
+++ b/Snowflake.Data.Tests/EncryptionStreamTest.cs
@@ -32,16 +32,16 @@ namespace Snowflake.Data.Tests
             
             try
             {
-                var readStream = new MemoryStream(rawData, false);
                 // first, write the data to the crypt file
-                using (var encryptStream = new EncryptionStream(File.OpenWrite(randomFileNameCrypt), EncryptionStream.CryptMode.Encrypt, pgMaterial, metaData, false))
+                using (var encryptStream = EncryptionStream.Create(new MemoryStream(rawData, false), EncryptionStream.CryptMode.Encrypt, pgMaterial, metaData, false))
+                using (var dest = File.OpenWrite(randomFileNameCrypt))
                 {
-                    readStream.CopyTo(encryptStream);
+                    encryptStream.CopyTo(dest);
                 }
 
                 var newStream = new MemoryStream();
                 // next read the data from the encrypted file into the newStream
-                using (var decryptStream = new EncryptionStream(File.OpenRead(randomFileNameCrypt), EncryptionStream.CryptMode.Decrypt, pgMaterial, metaData, false))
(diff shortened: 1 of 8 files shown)


Gonyoda:

sfc-gh-igarish:
@Gonyoda Thanks for your contribution. Did you get a chance to check performance difference byte[] v/s temp file stream?

Gonyoda:
@sfc-gh-igarish There are 2 aspects of performance, so I'll try to answer both using BenchmarkDotNet.

I created [a project](https://github.com/Snagajob/snowflake-connector-net/blob/benchmarks/Snowflake.Data.Benchmarks/EncryptionStreamBenchmark.cs) that runs a benchmark for the memory stream (old) and the file stream (this PR).  I was not able to run both benchmarks simultaneously, I just switched between the two branches and commented out the benchmark that wouldn't compile.  I ran the benchmarks with a few different file sizes - small, medium and large.  We often load files between medium and large.

I merged the results below:

```
100,000 bytes
|          Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 |     Gen2 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|---------:|---------:|----------:|
|   RunFileStream | 10.77 ms | 0.1470 ms | 0.1310 ms |  15.6250 |  15.6250 |  15.6250 | 110.46 KB |
| RunMemoryStream | 5.551 ms | 0.0878 ms | 0.0822 ms | 500.0000 | 500.0000 | 500.0000 |   1.72 MB |


10,000,000 bytes
|          Method |     Mean |    Error |   StdDev |      Gen0 |      Gen1 |      Gen2 | Allocated |
|---------------- |---------:|---------:|---------:|----------:|----------:|----------:|----------:|
| RunFileStream   | 46.18 ms | 0.899 ms | 1.503 ms |  333.3333 |  333.3333 |  333.3333 |   9.58 MB |
| RunMemoryStream | 39.81 ms | 0.789 ms | 1.025 ms | 1875.0000 | 1875.0000 | 1875.0000 |  92.76 MB |

220,000,000 bytes
|          Method |      Mean |    Error |    StdDev |   Median |      Gen0 |      Gen1 |      Gen2 | Allocated |
|---------------- |----------:|---------:|----------:|---------:|----------:|----------:|----------:|----------:|
|   RunFileStream | 1036.0 ms | 69.70 ms | 205.40 ms |          |           |           |           | 209.95 MB |
| RunMemoryStream |  935.4 ms | 18.35 ms |  51.46 ms | 918.0 ms | 2000.0000 | 2000.0000 | 2000.0000 |    1.6 GB |

```
 
Conclusion:
No doubt the file-writing is a bit slower than in-memory buffers, but it's not significantly slower.  However, it should be plain to see that the original memory stream uses significantly more memory, especially for large files.

sfc-gh-pbulawa:
Thank you for your contribution. The issue has been addressed within PR #734.

PR #631, opened 2023-03-20
for issue 52, increase retry to 6 times to avoid incomplete json response

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/52


sfc-gh-igarish (approved):
LGTM. As such we have plan to do exponential retry and include some jitter in that also with max 90 sec total retry time. We can do this change later once we give you details about it.

sfc-gh-ext-simba-dl:
on March release 2.0.22

PR #634, opened 2023-03-29
Update README.md: ".NET Core 6.0" to just ".NET 6.0"

wdolek:
I know this is silly PR, but "Core" has been removed from framework name with .NET 5.0, see [Target frameworks](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions).

// look for image: ralph I'm helping


wdolek:

PR #636, opened 2023-03-30
issue 257 connection not terminate session when connection pooling enable

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/257
The root case of this issue is no clean up logic for connection pool. Since the connection pool functions are all static, so no instance for it to handle the application exit. Adding ConnectionPoolSingleton when connection pool being used. The destructor will run when the application exit.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
@@ -8,31 +8,51 @@ using System.Linq;
 
 namespace Snowflake.Data.Client
 {
-    public class SnowflakeDbConnectionPool
+    sealed class ConnectionPool : IDisposable
     {
         private static SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbConnection>();
-
-        private static List<SnowflakeDbConnection> connectionPool;
+        private static ConnectionPool instance = null;
         private static readonly object _connectionPoolLock = new object();
-        private static int maxPoolSize;
-        private static long timeout;
-        private const int MAX_POOL_SIZE = 10;
+
+        private List<SnowflakeDbConnection> connectionPool;
+        private int maxPoolSize;
+        private long timeout;
+        private int MAX_POOL_SIZE = 10;
         private const long TIMEOUT = 3600;
-        private static bool pooling = true;
+        private bool pooling = true;
 
-        SnowflakeDbConnectionPool()
+        ConnectionPool()
         {
-            initConnectionPool();
-        }
 
-        ~SnowflakeDbConnectionPool()
+        }
+        ~ConnectionPool()
         {
             ClearAllPools();
         }
 
-        private static void initConnectionPool()
+        public void Dispose()
         {
-            logger.Debug("SnowflakeDbConnectionPool::initConnectionPool");
+            ClearAllPools();
+        }
+
+        public static ConnectionPool Instance
+        {
+            get
+            {
+                lock (_connectionPoolLock)
+                {
+                    if(instance == null)
+                    {
+                        instance = new ConnectionPool();
+                    }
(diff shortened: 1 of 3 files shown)


sfc-gh-igarish:
Could you please describe how your changes will fix the issue?

sfc-gh-ext-simba-dl:
I used singleton for connection pool and added destructor, which will run clear all connections there. I ran a test program which will create some connections, and when the program finished, it will hit the break point on the connection pool destructor.

sfc-gh-ext-simba-hx:
@sfc-gh-igarish 
Please continue on the review. The changes passed internal review on Simba side.

sfc-gh-ext-simba-dl:
The root case of this issue is no clean up logic for connection pool. Since the connection pool functions are all static, so no instance for it to handle the application exit. Adding ConnectionPoolSingleton when connection pool being used. The destructor will run when the application exit. 
I put a breakpoint at the destructor, and open wireshark to record the https transactions. 
This is the wireshark image to show the message send out to the server for close session. 
wireshark1

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -8,47 +8,66 @@
 
 namespace Snowflake.Data.Client
 {
-    public class SnowflakeDbConnectionPool
+    sealed class ConnectionPoolSingleton : IDisposable
     {
         private static SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbConnection>();
-
-        private static List<SnowflakeDbConnection> connectionPool;
+        private static ConnectionPoolSingleton instance = null;
```
sfc-gh-igarish, line 14: will it be a breaking change, as from the list of pool change to a singleton?

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -8,47 +8,66 @@
 
 namespace Snowflake.Data.Client
 {
-    public class SnowflakeDbConnectionPool
```
sfc-gh-igarish, line 11: As we change class visibility this could be consider as BCR.

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -127,52 +136,96 @@ internal static bool addConnection(SnowflakeDbConnection conn)
             return false;
         }
 
-        public static void ClearAllPools()
+        internal void ClearAllPools()
```
sfc-gh-igarish, line 139: will it break applications as change method from public to internal?

Snowflake.Data/Client/SnowflakeDbConnectionPool.cs
```diff
@@ -127,52 +136,96 @@ internal static bool addConnection(SnowflakeDbConnection conn)
...
+                foreach (SnowflakeDbConnection conn in connectionPool)
+                {
+                    conn.PostClose();
+                }
                 connectionPool.Clear();
             }
         }
 
-        public static void SetMaxPoolSize(int size)
+        public void SetMaxPoolSize(int size)
```
sfc-gh-igarish, line 152: One more breaking change.

sfc-gh-ext-simba-hx:
@sfc-gh-igarish The existing interface become a wrapper of the internal singleton here and keep all interface unchanged.
https://github.com/snowflakedb/snowflake-connector-net/pull/636/files#diff-27546517714d05a20246f398e52ede33a1f692ec4eced98fd5c061a109e3951eR184

sfc-gh-igarish (approved):
@sfc-gh-ext-simba-hx Thanks for clarification. LGTM.

PR #638, opened 2023-04-07
issue 258, fix COMMITs interrupted by an unnecessary ROLLBACK

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/258
adding bool isCommittedOrRollbacked to take care transaction already commit or commit twice issue and prevent commit interrupted by dispose function. 
Test case TestTransactionRollback is used to reproduce a COMMIT and ROLLBACK race condition. 
Test case TestTransactionRollbackOn2Transactions is used to test the transaction COMMIT and ROLLBACK in normal cases.


sfc-gh-igarish (approved):
LGTM

PR #641, opened 2023-04-13
Bug fix: issue 308, issue 311 

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/308
USER property is mandatory for external browser auth.
root cause is after using DbConnectionStringBuilder, when we have user=;password=; DbConnectionStringBuilder will not add this to connection string properties, but the old version we still add it with empty string. 
solve it by reading the input connection string, if user or password exist, and not exist from DbConnectionStringBuilder, add user or password with empty string.

https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/311
Customer can't access lowercase database from connection string. 
root cause is DbConnectionStringBuilder will remove all double quotes in front or after the property value. So for those properties which will convert to upper case by default (if not double quotes wrapped), such as database, schema, warehouse and role. We will get the value from the old way.

### DIFF
--- a/Snowflake.Data/Core/SFSessionProperty.cs
+++ b/Snowflake.Data/Core/SFSessionProperty.cs
@@ -177,7 +177,8 @@ namespace Snowflake.Data.Core
                 if(keyVal.Length > 0)
                 {
                     string[] tokens = keyVal.Split(new string[] { "=" }, StringSplitOptions.None);
-                    if(tokens[0].ToUpper() == "DB" || tokens[0].ToLower() == "SCHEMA" || tokens[0].ToLower() == "WAREHOUSE")
+                    if(tokens[0].ToUpper() == "DB" || tokens[0].ToLower() == "SCHEMA" ||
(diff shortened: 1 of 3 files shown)


Snowflake.Data/Core/SFSessionProperty.cs
```diff
@@ -170,6 +170,35 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
                 }
             }
 
+            //handle DbConnectionStringBuilder missing cases
```
sfc-gh-igarish, line 173: We need to investigate why DbConnectionStringBuilder missing this and we need to hand code this stuff.

sfc-gh-igarish (approved):
LGTM

Snowflake.Data/Core/SFSessionProperty.cs
sfc-gh-ext-simba-dl: DbConnectionStringBuilder is good for password and special characters. It doesn't handle using double quote to warp the database name means case sensitive.

Issue #210, opened 2020-06-02
SNOW-164194: Connect to Snowflake using .NET driver


SarekaAvinash:
### Issue description
As per the official documentation [LINK](https://github.com/snowflakedb/snowflake-connector-net)
it was stating that we can connect to SnowFlake using VisualStudion2017 only and no other version of VisualStudio won't support it. But I created a .NET Core API using V.S 2019 and tried to connect to Snowflake and succeeded in fetching data. 

My question is like, SnowFlake drivers will support for VS2017, VS2019?

sfc-gh-igarish:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of April 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

PR #642, opened 2023-04-14
SNOW-777930 Bug fix: .NET application terminates when client_session_keep_alive=true due to unhandled exception

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/307

There were mainly two issues:
1. unhandled exception from heartbeat session renew crashes upper application
2. heartbeat queue growth possibly uncontrolled

The root cause of issue 1 is that when trying to send heart beat for a session, there could be a bad timing that the session token expired so renew session is triggered, and the session is closed right after so renew session could fail, and the unhandled exception was thrown there. 
Removing the exception throw will fix this issue.

There are multiple reason for issue 2 though.
- SFSession.Close() doesn't remove the session out of heart beat queue. https://github.com/snowflakedb/snowflake-connector-net/blob/v2.0.22/Snowflake.Data/Core/SFSession.cs#L258
- SFSnowflakeConnection.Dispose() set disposed flag even when the connection is pooled, so when the connection is reused by a new connection and disposed again, nothing would happen as the disposed flag is set to true already so the session was leaked. https://github.com/snowflakedb/snowflake-connector-net/blob/v2.0.22/Snowflake.Data/Client/SnowflakeDbConnection.cs#L335
To fix issue 2, always call stopHeartBeatForThisSession() when close the session, also in SFSnowflakeConnection.Dispose() only set disposed when it's not pooled.

There are other bug fixes as well:
- CLIENT_SESSION_KEEP_ALIVE in connection string. It's actually not working as the driver didn't send it to server as session parameter
- pooling timeout was reset each time when the connection is reused and pooled again. This could make the connection never expire and application could get error when the session expired. Set pooling timeout only at the first time it's pooled.

Some other minor changes:
- add logging of session id.
- add some internal functions in HeartBeatBackground for test purpose.
- remove usage of codecov as it's deprecated and fail the test run.
- Ignore known test failure as it's been there for months and we have [sdk issue #210](https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/210) for that already.

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -132,7 +132,6 @@ namespace Snowflake.Data.Client
         {
             if (SfSession != null)
             {
-                SfSession.stopHeartBeatForThisSession();
                 SfSession.close();
             }
         }
@@ -141,7 +140,6 @@ namespace Snowflake.Data.Client
         {
             if (SfSession != null)
             {
-                SfSession.stopHeartBeatForThisSession();
                 SfSession.CloseAsync(cancellationToken).ContinueWith(
                 previousTask =>
                 {
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -24,7 +24,7 @@ namespace Snowflake.Data.Client
 
         internal int _connectionTimeout;
 
-        internal long _poolTimeout;
+        internal long _poolTimeout = 0;
 
         private bool disposed = false;
 
@@ -362,6 +362,13 @@ namespace Snowflake.Data.Client
             {
                 externalCancellationToken.Register(() => { _connectionState = ConnectionState.Closed; });
             }
+        }
+
+        // Called by connection pooling when the connection is removed out of pool
+        // could be reused or destroyed when expire
+        internal void Unpool()
+        {
+            pooled = false;
         }
 
         ~SnowflakeDbConnection()
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -340,10 +340,14 @@ namespace Snowflake.Data.Client
                 // Prevent an exception from being thrown when disposing of this object
                 logger.Error("Unable to close connection", ex);
             }
-            
-            disposed = true;
 
-            base.Dispose(disposing);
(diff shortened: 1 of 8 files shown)


Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -342,10 +340,14 @@ protected override void Dispose(bool disposing)
                 // Prevent an exception from being thrown when disposing of this object
                 logger.Error("Unable to close connection", ex);
             }
-            
-            disposed = true;
 
-            base.Dispose(disposing);
+            // only dispose the connection when it's not pooled
```
sfc-gh-igarish, line 344: what will happen when pool wanted to destroy the connection?

Snowflake.Data/Core/SFSession.cs
```diff
@@ -486,22 +486,25 @@ internal void heartbeat()
...
                             }
                             catch (Exception ex)
                             {
-                                logger.Error("renew session failed.", ex);
-                                throw;
+                                // Since we don't lock the heart beat queue when sending
+                                // the heart beat, it's possible that the session get
+                                // closed when sending renew request and caused exception
+                                // thrown from renewSession(), simply ignore that
+                                logger.Error($"renew session (ID: {sessionId}) failed.", ex);
```
sfc-gh-igarish, line 500: Do we need to break the loop after some try?

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-ext-simba-hx: Good point. Will add fix for such case. Connection pool should be able to unset the pooled flag when removing the connection out of the queue and destroy it.

sfc-gh-ext-simba-hx: Actually that won't cause issue as the connection status will be closed so Dispose() won't add it to pool again so it will be destroyed.

sfc-gh-ext-simba-hx: But add reset of the pooled flag anyway to make the status more clear. Also make the change to always call Dispose() do destroy the connection.

Snowflake.Data/Core/SFSession.cs
sfc-gh-ext-simba-hx: fixed

sfc-gh-igarish (approved):
LGTM

PR #645, opened 2023-04-18
DOC-661 Add batch execution feature

sfc-gh-mheavin:
Add batch SQL statement execution

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,16 +469,16 @@ You can execute multiple statements as a batch in the same way you execute queri
 You can set this parameter at the session level using the following command:
 
 ```
-ALTER SESSION SET MULTI_STATEMENT_COUNT = <n>;
+ALTER SESSION SET MULTI_STATEMENT_COUNT = <0/1>;
 ```
 
-where _n_ is:
+where:
 
-- 0: Enable an unspecified number of SQL statements in a query. 
+- **0**: Enable an unspecified number of SQL statements in a query. 
 
     Using this value allows batch queries to contain any number of SQL statements without needing to specify the MULTI_STATEMENT_COUNT statement parameter. However, be aware that using this value reduces the protection against SQL injection attacks.
 
-- 1: Allow one SQL statement or a specified number of statement in a query string (default).
+- **1**: Allow one SQL statement or a specified number of statement in a query string (default).
 
     You must include MULTI_STATEMENT_COUNT as a statement parameter to specify the number of statements included when the query string contains more than one statement. If the number of statements sent in the query string does not match the  MULTI_STATEMENT_COUNT value, the .NET driver rejects the request. You can, however, omit this parameter if you send a single statement.
 
--- a/README.md
+++ b/README.md
@@ -7,10 +7,9 @@ Snowflake Connector for .NET
 
 The Snowflake .NET connector supports the the following .NET framework and libraries versions:
 
-- .NET Framework 4.7.1
 - .NET Framework 4.7.2
 - .NET Framework 4.7.3
-- .NET 6.0
+- .NET Core 6.0
 
 Please refer to the Notice section below for information about safe usage of the .NET Driver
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -590,71 +671,6 @@ Here is a sample app.config file that uses [log4net](http://logging.apache.org/l
   </log4net>
 ```
 
-Getting the code coverage
```
sfc-gh-igarish, line 593: Why you removed this section? This is needed to get the code coverage of the driver.

sfc-gh-igarish (approved):
LGTM

README.md
sfc-gh-mheavin: commit blip. Content is there

README.md
```diff
@@ -450,6 +450,88 @@ Note that because this method is not available in the generic `IDataReader` inte
...
+
+By setting the value the session level, you do not need to set it when you execute each time you execute a batch statement. The following example sets the number of statements at the session level to three and then executes three SQL statements:
+
+```cs
+using (IDbConnection conn = new SnowflakeDbConnection())
+{
+	conn.ConnectionString = ConnectionString;
+	conn.Open();
+	IDbCommand cmd = conn.CreateCommand();
+	cmd.CommandText = "ALTER SESSION SET MULTI_STATEMENT_COUNT = 3;";
```
sfc-gh-ext-simba-hx, line 485: This is incorrect. Session parameter can only be used to turn multiple statements feature on/off, the valid values there can only be 0 or 1.
To set number of statements you have to set it through the parameter per query, as the following example.
Please take ODBC doc as a reference. https://docs.snowflake.com/en/developer-guide/odbc/odbc-using#specifying-the-number-of-statements-in-a-batch

sfc-gh-ext-simba-hx (changes requested):

README.md
sfc-gh-mheavin: Updated for review

README.md
```diff
@@ -464,50 +464,32 @@ By default, Snowflake returns an error for queries issued with multiple statemen
...
-For example, if you set MULTI_STATEMENT_COUNT=3, a batch statement must include precisely three statements. If you submit a batch statement with any other number of statements, the Node.js driver rejects the request. You can set MULTI_STATEMENT_COUNT=0 to allow batch queries to contain any number of statements. However, be aware that using this value reduces the protection against SQL injection attacks.
-
-You can set this parameter at the session level using the following command, or you can set the value separately each time you submit a query.
+You can set this parameter at the session level using the following command:
 
 ```
 ALTER SESSION SET MULTI_STATEMENT_COUNT = <n>;
 ```
 
-By setting the value the session level, you do not need to set it when you execute each time you execute a batch statement. The following example sets the number of statements at the session level to three and then executes three SQL statements:
```
sfc-gh-ext-simba-hx: Sorry for the misleading. I think we need some correction here. `By setting the value the session level, you do not need to set it when you execute each time you execute a batch statement.` This is true and need to be mentioned in some way, since setting MULTI_STATEMENT_COUNT=0 in session level is not just enable multiple statements feature, it allows any number of statements in a query, and that could be vulnerable to SQL injection. Better to mention that and suggest setting number of statements per query through is a more safe way.

sfc-gh-ext-simba-hx: @sfc-gh-mheavin In another word, using parameter to set number of statements per query doesn't require setting MULTI_STATEMENT_COUNT=0 in session level, it works when MULTI_STATEMENT_COUNT=1 in session level so basically the we suggest customer to keep MULTI_STATEMENT_COUNT=1 in session level and set number of statements per query through parameter.

sfc-gh-mheavin: Rewrote section for review.

sfc-gh-ext-simba-hx (approved):
LGTM

sfc-gh-ext-simba-dl (approved):

PR #647, opened 2023-04-21
Re-enable Stage binding async test for .net 

sfc-gh-ext-simba-dl:
re-enable the stage binding async test case, reduce the bind data size to prevent mac machine failed. 
reduce the number of rows for binding to 15000, and 25000 is a workaround for Mac VM limit.


Snowflake.Data.Tests/SFDbCommandIT.cs
```diff
@@ -751,22 +752,22 @@ public void TestPutArrayBindAsyncMultiThreading()
 
         static void ThreadProcess1(string connstr)
         {
-            ArrayBindTest(connstr, "testPutArrayBind1", 250000);
+            ArrayBindTest(connstr, "testPutArrayBind1", 15000);
```
sfc-gh-jtang, line 755: can you explain to me why change the value from 250,000 to 15,000 and the other changes from 1,250,000 to 25,000?

sfc-gh-ext-simba-dl: For github Mac vm, usually we got random error when the data size is too large, but on a physical machine it works fine. I just reduce to number of rows to prevent it use up too much vm memory.

sfc-gh-jtang (approved):

PR #651, opened 2023-05-11
A command may be executed within an unexpected transaction.

asiwjp:
If Close or Dispose is called for a Connection with an uncomplete transaction and this Connection is reused from the ConnectionPool, SQL commands will be executed in the old transaction scope. 
Uncompleted transaction should be rolled back before Close. In addition, if there is any chance of a transaction surviving, it should not be added to the Pool. 
In other cases, an explicit BEGIN may be unknowingly returned to the pool, so connections should be checked before they are reused.

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -412,6 +412,68 @@ namespace Snowflake.Data.Tests
         }
 
         [Test]
+        public void TestConnectionWithZombieTransaction()
+        {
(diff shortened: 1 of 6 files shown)


asiwjp:
recheck

asiwjp:

sfc-gh-dszmolka:
first of all, thank you for your contribution @asiwjp ! after a long internal discussion and considering some factors which are tied to the internal workings of the Snowflake engine itself, we chose a different approach for the sake of performance. the change is merged on #691 so I'm now closing this one too.

PR #659, opened 2023-05-25
SNOW-827383 .NET Driver Unit Tests failing on non-default account's settings

sfc-gh-dstempniak:
Fixes of tests:
- **SFDbTransactionIT.TestTransactionRollback, SFDbTransactionIT.TestTransactionRollbackOn2Transactions**
If account's TIMESTAMP_TYPE_MAPPING is set to TIMESTAMP_LTZ then in testTransation TIMESTAMP is treated as TIMESTAMP_LTZ and reader.GetDateTime throws an exception. Of course we can (a) set session's TIMESTAMP_TYPE_MAPPING to TIMESTAMP_NTZ or (b) change account's parameter. But as we don't test here TIMESTAMP_* variations I think explicit creation of TIMESTAMP_NTZ is the simplest solution.
- **SFMultiStatementsIT.testWithAllQueryTypes**
On our dev envs we get: SQL compilation error: error line 1 at position 4 Invalid explain plan format 'tabular'.
Since that is not the main purpose of the test (we test here multi line statements) I've change it to 'text'.
One more info - setting FORCE_USING_EXTERNAL_EXPLAIN = true fixes the problem with 'tabular' on local env.

### DIFF
--- a/Snowflake.Data.Tests/SFDbTransactionIT.cs
+++ b/Snowflake.Data.Tests/SFDbTransactionIT.cs
@@ -68,7 +68,7 @@ namespace Snowflake.Data.Tests
             conn.Open();
 
             DbCommand command = conn.CreateCommand();
-            command.CommandText = "create or replace table testTransaction(x timestamp, a integer)";
+            command.CommandText = "create or replace table testTransaction(x timestamp_ntz, a integer)";
             command.ExecuteNonQuery();
             Console.WriteLine("Created table");
 
@@ -118,7 +118,7 @@ namespace Snowflake.Data.Tests
             conn.Open();
 
             DbCommand command = conn.CreateCommand();
-            command.CommandText = "create or replace table testTransaction(x timestamp, a integer)";
+            command.CommandText = "create or replace table testTransaction(x timestamp_ntz, a integer)";
             command.ExecuteNonQuery();
             Console.WriteLine("Created table");
 
--- a/Snowflake.Data.Tests/SFMultiStatementsIT.cs
+++ b/Snowflake.Data.Tests/SFMultiStatementsIT.cs
@@ -302,6 +302,7 @@ namespace Snowflake.Data.Tests
                 {
                     cmd.CommandText = "select 1;" +
                                       "create or replace temporary table t1(c1 varchar);" +
+                                      "alter session set force_using_external_explain = true;" +
                                       "explain using tabular select * from t1;" +
                                       "show parameters;" +
                                       "insert into t1 values ('str1');" +
@@ -316,7 +317,7 @@ namespace Snowflake.Data.Tests
                     var stmtCountParam = cmd.CreateParameter();
                     stmtCountParam.ParameterName = "MULTI_STATEMENT_COUNT";
                     stmtCountParam.DbType = DbType.Int16;
-                    stmtCountParam.Value = 11;
+                    stmtCountParam.Value = 12;
                     cmd.Parameters.Add(stmtCountParam);
 
                     DbDataReader reader = cmd.ExecuteReader();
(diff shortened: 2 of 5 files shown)


sfc-gh-igarish:
@sfc-gh-dstempniak Please provide the description.

Snowflake.Data.Tests/SFDbTransactionIT.cs
```diff
@@ -68,7 +68,7 @@ public void TestTransactionRollback()
             conn.Open();
 
             DbCommand command = conn.CreateCommand();
-            command.CommandText = "create or replace table testTransaction(x timestamp, a integer)";
+            command.CommandText = "create or replace table testTransaction(x timestamp_ntz, a integer)";
```
sfc-gh-igarish, line 71: Why change the data type?

Snowflake.Data.Tests/SFDbTransactionIT.cs
```diff
@@ -118,7 +118,7 @@ public void TestTransactionRollbackOn2Transactions()
...
```
sfc-gh-igarish, line 121: why change the data type?

Snowflake.Data.Tests/SFMultiStatementsIT.cs
```diff
@@ -302,7 +302,7 @@ public void testWithAllQueryTypes()
                 {
                     cmd.CommandText = "select 1;" +
                                       "create or replace temporary table t1(c1 varchar);" +
-                                      "explain using tabular select * from t1;" +
+                                      "explain using text select * from t1;" +
```
sfc-gh-igarish, line 305: Why change the SQL?

Snowflake.Data/Core/FastParser.cs
```diff
@@ -100,7 +101,7 @@ public static decimal FastParseDecimal(byte[] s, int offset, int len)
                 catch (OverflowException)
                 {
                     // Fallback to regular decimal constructor from string instead.
-                    return decimal.Parse(UTF8Buffer.UTF8.GetString(s, offset, len));
+                    return decimal.Parse(UTF8Buffer.UTF8.GetString(s, offset, len), CultureInfo.InvariantCulture);
```
sfc-gh-igarish, line 104: Before making code changes, please create a doc for short coming of existing code and what we are going to do to support Locale. We need to find out all areas missing Locale support.

Snowflake.Data.Tests/SFDbTransactionIT.cs
sfc-gh-dstempniak: If account's TIMESTAMP_TYPE_MAPPING is set to TIMESTAMP_LTZ then in testTransation TIMESTAMP is treated as TIMESTAMP_LTZ and reader.GetDateTime throws an exception. Of course we can (a) set session's TIMESTAMP_TYPE_MAPPING to TIMESTAMP_NTZ or (b) change account's parameter. But as we don't test here TIMESTAMP_* variations I think explicit creation of TIMESTAMP_NTZ is the simplest solution.

Snowflake.Data.Tests/SFDbTransactionIT.cs
sfc-gh-dstempniak: Same as above. Maybe such explanation should be given somewhere in jira? Or github issue comment?

Snowflake.Data.Tests/SFMultiStatementsIT.cs
sfc-gh-dstempniak: On our dev envs we get: SQL compilation error: error line 1 at position 4 Invalid explain plan format 'tabular'.
Since that is not the main purpose of the test (we test here multi line statements) I've change it to 'text'.
One more info - setting FORCE_USING_EXTERNAL_EXPLAIN = true fixes the problem with 'tabular' on local env.

sfc-gh-igarish (approved):
LGTM

PR #661, opened 2023-05-26
SNOW-828109: Remove unnecessary restrictions on database host in tests

sfc-gh-knozderko:
Test SFConnectionIT.TestInvalidProxySettingWithByPassListFromConnectionString() imposes unnecessary restrictions on database host address. This test is passing only if the database host is localhost (or testConfig.account + ".snowflakecomputing.com"). The goal is to make the test pass also for other database hosts.

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -1285,7 +1285,7 @@ namespace Snowflake.Data.Tests
                 = ConnectionString
                 + String.Format(
                     ";useProxy=true;proxyHost=Invalid;proxyPort=8080;nonProxyHosts={0}",
-                    "*.foo.com %7C" + testConfig.account + ".snowflakecomputing.com|localhost");
+                    "*.foo.com %7C" + testConfig.account + ".snowflakecomputing.com|" + testConfig.host);
                 conn.Open();
                 // Because testConfig.host is in the bypass list, the proxy should not be used
             }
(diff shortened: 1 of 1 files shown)


Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -1285,7 +1285,7 @@ public void TestInvalidProxySettingWithByPassListFromConnectionString()
                 = ConnectionString
                 + String.Format(
                     ";useProxy=true;proxyHost=Invalid;proxyPort=8080;nonProxyHosts={0}",
-                    "*.foo.com %7C" + testConfig.account + ".snowflakecomputing.com|localhost");
+                    "*.foo.com %7C" + testConfig.account + ".snowflakecomputing.com|" + testConfig.host);
```
sfc-gh-igarish, line 1288: Please read comment below. This is a negative test for a proxy connection. So it's hard coded localhost in the connection string. am I missing anything?

sfc-gh-knozderko: That's right, the proxy will not be used, but the reason is because of this: "proxyHost=Invalid" together with "nonProxyHosts=...|localhost".

So when I run it with the database on localhost the test passes because testConfig.host = localhost.
But once I change the database to some other host it is failing because database is not on nonProxyHosts list.  
And the test should pass always regardless of what database host name is used.

I'm not sure what are you asking me about for about this comment? Do you want me to delete the comment?

sfc-gh-igarish (approved):
LGTM. Thanks for the explanation.

PR #662, opened 2023-05-29
Do not log proxy credentials

sfc-gh-ext-simba-hx:
Bug fix for issue: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/400

Went through source code and removed all logging of proxy password or entire connection string. 
Added proxypassword to SecretDetector as well to prevent it to be logged in the future, but that would only work in case it's logged with "proxypassword" keyword.


sfc-gh-igarish (approved):
LGTM

Issue #604, opened 2023-02-17
Reader.HasRows throws NullReferenceException when next result is not available


myshon:
### Issue description
reader.HasRows throws NullReferenceException when next result is not available. 

### Example code
Here is unit tests that reproduced the issue
```cs
    [Fact]
    public void ExecuteReader_WhenNoCommandLeft_HasRowsShouldReturnFalse()
    {
        using SnowflakeDbConnection conn = new SnowflakeDbConnection();
        conn.ConnectionString = "...";

        conn.Open();

        DbCommand cmd = conn.CreateCommand();
        var insertTemplate = "insert into db.public.output_table_01 select '{0}', '{1}', '{2}'";
        var cmdCommandText = string.Format(insertTemplate,
            Guid.NewGuid().ToString(),
            Guid.NewGuid().ToString(),
            Guid.NewGuid().ToString());
        
        var stmtCountParam = cmd.CreateParameter();
        stmtCountParam.ParameterName = "MULTI_STATEMENT_COUNT";
        stmtCountParam.DbType = DbType.Int16;
        stmtCountParam.Value = 3;
        cmd.Parameters.Add(stmtCountParam);
        
        cmd.CommandText = $"{cmdCommandText};{cmdCommandText};{cmdCommandText};";

        DbDataReader reader = cmd.ExecuteReader();
        
        Assert.False(reader.HasRows);
        Assert.Equal(1, reader.RecordsAffected);
        
        Assert.True(reader.NextResult());
        Assert.False(reader.HasRows);
        Assert.Equal(1, reader.RecordsAffected);
        
        Assert.True(reader.NextResult());
        Assert.False(reader.HasRows);
        Assert.Equal(1, reader.RecordsAffected);
        
        Assert.False(reader.NextResult());
        Assert.False(reader.HasRows); // --> throws NRE !!!
        
        reader.Close();
        
        conn.Close();
    }
```

### Error log
```
System.NullReferenceException
Object reference not set to an instance of an object.
   at Snowflake.Data.Core.ResultSetUtil.HasResultSet(SFBaseResultSet resultSet)
   at Snowflake.Data.Client.SnowflakeDbDataReader.get_HasRows()
```

### Configuration
*Driver version:* 2.0.20
*Dotnet framework and version:* NET6
*Server version:* 7.5.2
*Client OS:* Windows 10.

sfc-gh-igarish:
Thanks for reporting issue. We are looking into it.

[...]

Issue #413, opened 2022-01-21
SnowflakeDbDataReader.HasRows() always returns true


DR486:
### Issue description
SnowflakeDbDataReader.HasRows() should return true when there are rows, and false when there are no rows.  [Currently it's only returning true.  ](https://github.com/snowflakedb/snowflake-connector-net/blob/80b595164ce89c6678b3d16ec5e99610edb65caa/Snowflake.Data/Client/SnowflakeDbDataReader.cs#L71)

The comment for that function states that one row will always be returned. When using `while(reader.Read())` to read the results of a select query when no records are found, the while loop is skipped.

### Example code
```cs
//Always returns true, see https://github.com/snowflakedb/snowflake-connector-net/blob/80b595164ce89c6678b3d16ec5e99610edb65caa/Snowflake.Data/Client/SnowflakeDbDataReader.cs#L71
var exists = reader.HasRows(); 

while(reader.Read()){
//This (obviously) never executes if the query executed returns no results....  
//but that means that reader.HasRows() should indeed return false.
   var test = reader.GetValue(0); 
}


```

### Error log
```
N/A
```
There is an example in READMD.md file showing you how to enable logging.

### Configuration
*Driver version:* 2.0.8

*Dotnet framework and version:* .NET Core 3.1 & .NET 6

*Server version:* N/A
You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```

*Client OS:* Windows 10

Christian-Oleson:
Here is where this is hard coded:
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Client/SnowflakeDbDataReader.cs#L77

JCurto:
This is still being hard coded to true for all SELECT statement types (and others)?:
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/ResultSetUtil.cs#L72

sfc-gh-jtang:
@sfc-gh-igarish can you double check to see if it's fixed already?

sfc-gh-dszmolka:
hi and thank you for submitting this issue, also apologies for the long period without response! at least can post some good news now; this seems to be fixed with #665 and the fix is merged, so awaiting driver release.
will update this issue once again once the new driver is available.

sfc-gh-igarish:
fix is released in 2.0.25

PR #665, opened 2023-05-30
SNOW-828708 Fix for SnowflakeDbDataReade.HasRows

sfc-gh-dstempniak:
Fix for SnowflakeDbDataReade.HasRows. 
Issues:
https://github.com/snowflakedb/snowflake-connector-net/issues/413
https://github.com/snowflakedb/snowflake-connector-net/issues/604

### DIFF
--- a/Snowflake.Data.Tests/SFDbDataReaderIT.cs
+++ b/Snowflake.Data.Tests/SFDbDataReaderIT.cs
@@ -1323,5 +1323,59 @@ namespace Snowflake.Data.Tests
                 conn.Close();
             }
         }
+        
+        [Test]
+        public void TestHasRows()
+        {
+            using (DbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                conn.Open();
+
+                DbCommand cmd = conn.CreateCommand();
+                cmd.CommandText = "select 1 where 1=2";
+                
+                DbDataReader reader = cmd.ExecuteReader();
+                Assert.IsFalse(reader.HasRows);
+                reader.Close();
+
+                cmd.CommandText = "select 1;" +
+                                  "select 1 where 1=2;" +
+                                  "select 1;" + 
+                                  "select 1 where 1=2;";
+                
+                DbParameter param = cmd.CreateParameter();
+                param.ParameterName = "MULTI_STATEMENT_COUNT";
+                param.DbType = DbType.Int16;
+                param.Value = 4;
+                cmd.Parameters.Add(param);
+                
+                reader = cmd.ExecuteReader();
+                
+                // select 1
+                Assert.IsTrue(reader.HasRows);
+                reader.Read();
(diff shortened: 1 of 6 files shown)


Snowflake.Data/Core/SFResultSet.cs
```diff
@@ -183,6 +183,16 @@ internal override async Task<bool> NextResultAsync(CancellationToken cancellatio
             return await Task.FromResult(false);
         }
 
+        internal override bool HasRows()
+        {
+            if (isClosed)
+            {
+                throw new SnowflakeDbException(SFError.DATA_READER_ALREADY_CLOSED);
```
sfc-gh-igarish, line 190: do we need to throw an exception or we can just return false?

sfc-gh-igarish (approved):
Just check my one comment otherwise LGTM.

Snowflake.Data/Core/SFResultSet.cs
sfc-gh-dstempniak: I was based on Next(), but you're right. Changed to return false.

Snowflake.Data.Tests/SFDbDataReaderIT.cs
```diff
@@ -1323,5 +1323,59 @@ public void TestResultSetMetadata()
...
+                conn.Open();
+
+                DbCommand cmd = conn.CreateCommand();
+                cmd.CommandText = "select 1 where 1=2";
+                
+                DbDataReader reader = cmd.ExecuteReader();
+                Assert.IsFalse(reader.HasRows);
+                reader.Close();
+
+                cmd.CommandText = "select 1;" +
```
sfc-gh-mhofman, line 1342: I would maybe split it into separate TestMultiStatementHasRows

Snowflake.Data.Tests/SFMultiStatementsIT.cs
```diff
@@ -355,12 +355,12 @@ public void testWithAllQueryTypes()
 
                     // result of list
                     Assert.IsTrue(reader.NextResult());
-                    Assert.IsTrue(reader.HasRows);
+                    Assert.IsFalse(reader.HasRows); // no files staged for table t1
```
sfc-gh-mhofman, line 358: Maybe we could think of splitting this as well into separated tests

Snowflake.Data.Tests/SFDbDataReaderIT.cs
sfc-gh-dstempniak: Splitted

Snowflake.Data.Tests/SFMultiStatementsIT.cs
sfc-gh-dstempniak: It tests multi statement SQL so it is difficult to split it into separated tests.

sfc-gh-mhofman: My bad - your right

PR #666, opened 2023-05-31
Bug fix for "Session no longer exists" issue with connection pooling

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/416

The root cause is the improper design of connection pooling feature. It was pooling SnowflakeDBConnection instances and making copy when it's reused from pooling, so it's always possible that the connection hold by different threads could point to the same underlying session.

In the case of issue 416 with `Session no longer exists` error, thread A calls close() to close the connection and the connection get pooled, thread B reuse the connection right after and get the connection out of the pool. Then thread A calls Dispose() on the connection and since it's not pooled, it close the session while it still being used by thread B, and that's why thread B got error of `Session no longer exists`

The error was `Authentication token has expired` in earlier versions I think that's because we made fix for other issues of connection pooling which changed the behavior but the issue still there.

Pool the underlying session instead of the connection instance itself should be able to solve all the timing issue and make the logic more simple and clear. No copy on the session instance and a session should be always either in pool or hold by one thread.

Test cases added for connection pooling with concurrent connections.

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -15,6 +15,7 @@ namespace Snowflake.Data.Tests
     using System.Diagnostics;
     using Snowflake.Data.Tests.Mock;
     using System.Runtime.InteropServices;
+    using System.Data.Common;
 
     [TestFixture]
     class SFConnectionPoolT : SFBaseTest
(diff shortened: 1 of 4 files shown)


sfc-gh-igarish:
why we change pooling concept from connection object to a session object? Can we not directly use connection object without copy? Which information we may loose if we only do session pooling? as such Connection object should have session reference, in that case connection object has more info than session object.

sfc-gh-ext-simba-hx:
@sfc-gh-igarish 
Connection pooling feature should be transparent to application, which means:
- the application can use the same implementation, create a new connection object, call open() to establish the connection, run queries and then call close() or simply leave it to .NET GC to destroy the connection object, regardless whether connection pooling is enabled.
- the connection object should be as if it's newly created every time, even it's being reused internally
- connection pooling feature is just for performance improvement, saving overhead on establish a new connection through login request, no other behavior change expected

So I think pooling the underlying session object would be a better design and should be done initially. SnowfakeDBConnection is what the application can see and should keep as what it is as before. SFSession is the actual underlying connection that fully controlled by the driver, which is the only thing needs to be reused. This was actually raised in PR review by a customer when adding this feature but got ignored. https://github.com/snowflakedb/snowflake-connector-net/pull/430#issuecomment-1093261761

shawner18:
This may or may not be useful...
I was having issues with [2.0.24] in SnowflakeDBConnection. 
Debugging was showing the destruction/finalizer of SnowflakeDBConnection setting the SFSession.sessionToken to null, even though another SnowflakeDBConnection in the pool had what I believe was a reference to the same SFSession. Which made the SFSession invalid. I assume this had to do with parallel tasks/queries using the same SnowflakeDBConnection.

I was able to reproduce every time in my application unit tests against 2.0.24. 
I did pull this branch and test with my unit tests and it passed. Looking forward to seeing this merged.

sfc-gh-igarish:
@sfc-gh-ext-simba-hx please check comment above from @shawner18.

sfc-gh-ext-simba-hx:
> @sfc-gh-ext-simba-hx please check comment above from @shawner18.

@sfc-gh-igarish I believe that's actually the similar issue of issue#416 with the same root cause.
Setting SFSession.sessionToken to null is not the root cause, the driver only do that after close the session. Before the change of setting SFSession.sessionToken to null, the driver close the session but doesn't clear the session token, so other connection reference to the same session object could get error of "Session no longer exists" afterwards.
The root cause is what mentioned in the description in this PR, we should pool the session object instead of SnowflakeDBConnection, which would prevent any duplicated reference to the same session.
The design concept actually mentioned in MSFT documentation here as well:
https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-connection-pooling
`The pooler maintains ownership of the physical connection.`
In Snowflake, I think `physical connection` should be session, not SnowflakeDBConnection that visible to application and I think should be `logical connection`

sfc-gh-igarish (approved):
LGTM

PR #668, opened 2023-06-01
SNOW-830959: Refactor of custom http handler setup

sfc-gh-knozderko:
Refactor of HttpUtil class to extract logic of creating http handler to external classes.
It improves the code by applying Single Responsibility Principle and Open-Closed Principle to the refactored code.
A test was added for creating or not a web proxy based on configuration. The need to add this test was inspired by this ticket: SNOW-828572.

### DIFF
--- a/Snowflake.Data.Tests/Http/HttpUtilTest.cs
+++ b/Snowflake.Data.Tests/Http/HttpUtilTest.cs
@@ -36,5 +36,27 @@ namespace Snowflake.Data.Tests
 
             Assert.AreEqual(expectedIsRetryable, actualIsRetryable);
         }
+        
+        [Test]
+        public void ShouldCreateHttpClient()
+        {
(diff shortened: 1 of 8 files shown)


sfc-gh-igarish (approved):
LGTM

sfc-gh-ext-simba-hx:
Hi @sfc-gh-knozderko 

I think better to hold on this change as we are considering removing the use of WinHttpHandler. 
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/365
It was introduced to the driver for fixing a proxy issue while it doesn't fully fix the issue (the proxy issue was reported on Linux as well while WinHttpHandler only available on Windows) so we might need to switch back to use HttpClientHandler in all cases.

sfc-gh-knozderko:
Will not be merged

PR #669, opened 2023-06-02
SNOW-831638 removed FORCEPARSEERROR flag used for testing chunk download

sfc-gh-mhofman:
Improving testing of Chunk Downloader without necessity of providing special flag FORCEPARSEERROR affecting production code.
Enables ReusableChunkTest to validate download of whole query output on mocked Chunk Parser factory which is configurable to throw (in the test only code and no longer in production) requested number of exceptions.


sfc-gh-igarish:
@sfc-gh-mhofman could you please description?

sfc-gh-mhofman:
> @sfc-gh-mhofman could you please add description?

Done.

Snowflake.Data/Core/SFChunkDownloaderV2.cs
```diff
@@ -16,6 +16,7 @@
 
 namespace Snowflake.Data.Core
 {
+    [Obsolete("SFChunkDownloaderV2 is deprecated", false)]
```
sfc-gh-dstempniak, line 19: Since we don't want to remove V1 and V2 yet, maybe we could add some tests for them?
As I see there are tests commented out: TestUseV1ResultParser and TestUseV2ChunkDownloader. Please consider rewriting them, it should be easy now and will increase code coverage.

sfc-gh-mhofman: will do, thx @sfc-gh-dstempniak good point

PR #619, opened 2023-03-15
bug fit issue 52, SNOW-701875: .net connector throwing error: System.Net.Http.WinHttpException (80072EE2, 12002): Error 12002 calling WINHTTP_CALLBACK_STATUS_REQUEST_ERROR

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/52
add retry logic on put and get request. 
added local debug code for testing, but comment out for release version.

PR #677, opened 2023-06-08
Bug fix for retry on chunk downloading

sfc-gh-ext-simba-hx:
Revise the fix for issue#52 to fix bugs there
https://github.com/snowflakedb/snowflake-connector-net/pull/610
https://github.com/snowflakedb/snowflake-connector-net/pull/619
https://github.com/snowflakedb/snowflake-connector-net/pull/631

- rollback the retry added for request sending (in requester.cs) since we have retry on request sending in HttpClient.RetryHandler already. Extra retry here is unnecessary
- Reset chunk before retry on chunk downloading in case there could be garbage data left from previous attempt which could fail the chunk parsing
- Fix the bug that backoff time is not set correctly

### DIFF
--- a/Snowflake.Data/Core/HttpUtil.cs
+++ b/Snowflake.Data/Core/HttpUtil.cs
@@ -65,6 +65,7 @@ namespace Snowflake.Data.Core
     public sealed class HttpUtil
     {
         static internal readonly int MAX_RETRY = 6;
+        static internal readonly int MAX_BACKOFF = 16;
         private static readonly SFLogger logger = SFLoggerFactory.GetLogger<HttpUtil>();
 
         private HttpUtil() { }
@@ -295,7 +296,6 @@ namespace Snowflake.Data.Core
                 HttpResponseMessage response = null;
                 int backOffInSec = 1;
                 int totalRetryTime = 0;
-                int maxDefaultBackoff = 16;
 
                 ServicePoint p = ServicePointManager.FindServicePoint(requestMessage.RequestUri);
                 p.Expect100Continue = false; // Saves about 100 ms per request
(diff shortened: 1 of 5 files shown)


Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -387,8 +387,8 @@ internal RetryHandler(HttpMessageHandler innerHandler, bool disableRetry, bool f
                     await Task.Delay(TimeSpan.FromSeconds(backOffInSec), cancellationToken).ConfigureAwait(false);
                     totalRetryTime += backOffInSec;
                     // Set next backoff time
-                    backOffInSec = backOffInSec >= maxDefaultBackoff ?
-                            maxDefaultBackoff : backOffInSec * 2;
+                    backOffInSec = backOffInSec >= MAX_BACKOFF ?
+                            MAX_BACKOFF : backOffInSec * 2;
```
sfc-gh-igarish, line 391: Do we terminate based on max retry? Am I missing any thing here as outer loop is while(true), but I am not seeing terminate based on time or count.

sfc-gh-ext-simba-hx: Currently the retry on request sending is only limited by timeout.
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/HttpUtil.cs#L335
The timeout is set here:
https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/RestRequester.cs#L173
This is existing behavior has been there for years and won't be a problem in most cases since
- it only retry on retry-able errors https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/HttpUtil.cs#L367
- we have limited timeout by default
It's good to have a max number of retry though, which I think we could add later.

sfc-gh-igarish: But if we get retry-able errors continuously in the loop and never succeed, that will go in infinite loop. right?

sfc-gh-ext-simba-hx: Yes, it would be infinite loop if we keep getting retry-able errors. I agree a max number of retry should be added and maybe configurable. But that's existing behavior, not introduced by this change, so I think we should make a different PR for that later.

sfc-gh-igarish:
Please make the change. Please see JDBC PR open here: https://github.com/snowflakedb/snowflake-jdbc/pull/1406

sfc-gh-ext-simba-hx:
> Please make the change. Please see JDBC PR open here: [snowflakedb/snowflake-jdbc#1406](https://github.com/snowflakedb/snowflake-jdbc/pull/1406)

You mean in this PR or a different PR?

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-igarish: Discuss with Harry to do it as part of another PR.

sfc-gh-igarish (approved):
LGTM

PR #679, opened 2023-06-13
SNOW-831638 Fixed problem in chunk retry downloading process and improved testing of those retries

sfc-gh-mhofman:
- Fixed reset of chunk download process when exceptions are thrown during the download
- Removed FORCEPARSEERROR from production code (introduced for testing) 
- Improved testing of chunk downloading process
- Enabled tests for V1 Result Parser and v2 Chunk Downloader

### DIFF
--- a/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
+++ b/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
@@ -189,12 +189,7 @@ namespace Snowflake.Data.Core
                             retry = true;
                             // reset the chunk before retry in case there could be garbage
                             // data left from last attempt
-                            ExecResponseChunk chunkInfo = new ExecResponseChunk()
-                            {
-                                rowCount = chunk.RowCount,
-                                url = chunk.Url
-                            };
-                            chunk.Reset(chunkInfo, chunk.chunkIndexToDownload);
+                            chunk.Reset(chunkInfos[chunk.chunkIndexToDownload], chunk.chunkIndexToDownload);
                             await Task.Delay(TimeSpan.FromSeconds(backOffInSec), downloadContext.cancellationToken).ConfigureAwait(false);
                             ++retryCount;
                             // Set next backoff time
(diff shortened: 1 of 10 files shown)


Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
```diff
@@ -184,25 +181,15 @@ private async Task<IResultChunk> DownloadChunkAsync(DownloadContextV3 downloadCo
...
                             retry = true;
                             // reset the chunk before retry in case there could be garbage
                             // data left from last attempt
-                            ExecResponseChunk chunkInfo = new ExecResponseChunk()
-                            {
-                                rowCount = chunk.RowCount,
-                                url = chunk.Url
-                            };
-                            chunk.Reset(chunkInfo, chunk.chunkIndexToDownload);
+                            chunk.Reset(chunkInfos[chunk.chunkIndexToDownload], chunk.chunkIndexToDownload);
```
sfc-gh-ext-simba-hx, line 192: Thanks @sfc-gh-mhofman for catching the issue here. I agree uncompressedSize needs to be reset correctly as well (although I'm not sure if it would cause customer facing issue as the test passed previously).
I added the reset here to fix the issue that when chunk downloading failed in the middle, without reset the chunk the data left from the failed attempt will break all following retry attempts.
My concern of using chunkInfos here is it was used in main thread only before and the existing implementation carefully not passing chunk, not chunk index to the downloading thread.
https://github.com/snowflakedb/snowflake-connector-net/blob/v2.0.24/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs#L113
It's possible that application stop fetching in the middle and close the connection, while downloading thread still working. In such case if downloading thread try to access the resource in ChunkDownloader, which could be destroyed already, in C++ it would crash the driver for sure, but I'm not sure what would happen in .NET. If that's not a problem in .NET the current fix should be fine, otherwise maybe adding a new function in ExecResponeChunk like Reset4Retry(), simply reset internal data buffer using existing uncompressedSize.

sfc-gh-mhofman:

sfc-gh-igarish (approved):
LGTM. Please update the PR description with new changes and test changes, if any.

Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
sfc-gh-ext-simba-hx: LGTM

sfc-gh-ext-simba-hx (approved):

PR #683, opened 2023-06-20
Add tests for ChunkParserFactory and ChunkDeserializer

sfc-gh-ext-simba-lf:
Regarding issue 113

The PR adds tests for ChunkParserFactory and ChunkDeserializer to increase code coverage

### DIFF
--- a/Snowflake.Data.Tests/ChunkDeserializerTest.cs
+++ b/Snowflake.Data.Tests/ChunkDeserializerTest.cs
@@ -37,14 +37,19 @@ namespace Snowflake.Data.Tests
             // Do nothing;
         }
 
+        public IChunkParser getParser(string data)
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(data);
+            Stream stream = new MemoryStream(bytes);
+            return ChunkParserFactory.Instance.GetParser(stream);
+        }
+
         [Test]
         public async Task TestParsingEmptyChunk()
         {
             // Create sample data for parser
             string data = "[ ]";
-            byte[] bytes = Encoding.UTF8.GetBytes(data);
-            Stream stream = new MemoryStream(bytes);
-            IChunkParser parser = ChunkParserFactory.Instance.GetParser(stream);
+            IChunkParser parser = getParser(data);
 
             SFResultChunk chunk = new SFResultChunk(new string[1, 1]);
 
@@ -60,9 +65,7 @@ namespace Snowflake.Data.Tests
         {
             // Create sample data for parser
             string data = "[ [],  [] ]";
-            byte[] bytes = Encoding.UTF8.GetBytes(data);
-            Stream stream = new MemoryStream(bytes);
-            IChunkParser parser = ChunkParserFactory.Instance.GetParser(stream);
+            IChunkParser parser = getParser(data);
 
             SFResultChunk chunk = new SFResultChunk(new string[1, 1]);
 
@@ -78,9 +81,7 @@ namespace Snowflake.Data.Tests
         {
             // Create a sample data using non-JSON instead
             string data = "[ \"1\", \"1.234\", \"abcde\" ]";
-            byte[] bytes = Encoding.UTF8.GetBytes(data);
-            Stream stream = new MemoryStream(bytes);
-            IChunkParser parser = ChunkParserFactory.Instance.GetParser(stream);
+            IChunkParser parser = getParser(data);
 
             SFResultChunk chunk = new SFResultChunk(new string[1, 1]);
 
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/ChunkDeserializerTest.cs
```diff
@@ -0,0 +1,80 @@
...
+    using Snowflake.Data.Configuration;
+    using Snowflake.Data.Core;
+    using System.IO;
+    using System.Text;
+    using System.Threading.Tasks;
+
+    [TestFixture]
+    class ChunkDeserializerTest
+    {
+        int ChunkParserVersionDefault = SFConfiguration.Instance().GetChunkParserVersion();
```
sfc-gh-mhofman, line 17: Unless we don't have any mock library (we're on this) this kind of push/pop of the factory state makes me worried when thinking about running such tests in parallel (depending on who and how runs them). 
In such a case one test can change chunk parser delivered for the rest of the tests. 
Maybe we should indicate such tests with [NonParallelizable].
I should probably have done that within my previous PR for SFReusableChunkTest.
What do you think @sfc-gh-ext-simba-lf?

Snowflake.Data.Tests/ChunkParserFactoryTest.cs
```diff
@@ -0,0 +1,83 @@
...
+            SFConfiguration.Instance().UseV2JsonParser = useV2JsonParser;
+            SFConfiguration.Instance().ChunkParserVersion = chunkParserVersion;
+
+            // Get parser using sample stream
+            byte[] bytes = Encoding.UTF8.GetBytes("test");
+            Stream stream = new MemoryStream(bytes);
+
+            IChunkParser parser = null;
+
+            // GetParser() throws an error when ChunkParserVersion is not 1-3
```
sfc-gh-mhofman, line 50: Unless we can finally remove from the code all those parser versions, my further idea would be to introduce an enum to be reused within the code to get rid of magic numbers of parser/downloader versions.

Snowflake.Data.Tests/ChunkDeserializerTest.cs
sfc-gh-igarish: Yes, we should have serialize v/s parallel annotation, so we can run tests in parallel as needed.

sfc-gh-igarish:
@sfc-gh-ext-simba-lf Please check comments from sfc-gh-mhofman.

Snowflake.Data.Tests/ChunkParserFactoryTest.cs
sfc-gh-ext-simba-lf: Yeah that sounds good, making an enum will help make sure the version is within the proper range

Snowflake.Data.Tests/ChunkDeserializerTest.cs
sfc-gh-ext-simba-lf: I agree, since the parser version could affect other tests, then adding NonParallelizable should help with that

Would you like for me to add [NonParallelizable] to this test now or add it later when the tests are running in parallel

Snowflake.Data.Tests/ChunkDeserializerTest.cs
```diff
@@ -0,0 +1,80 @@
...
+            await parser.ParseChunk(chunk);
+
+            Assert.AreEqual(null, chunk.ExtractCell(0, 0).SafeToString());
+            Assert.AreEqual("1.234", chunk.ExtractCell(0, 1).SafeToString());
+            Assert.AreEqual(null, chunk.ExtractCell(0, 2).SafeToString());
+            Assert.AreEqual("2", chunk.ExtractCell(1, 0).SafeToString());
+            Assert.AreEqual(null, chunk.ExtractCell(1, 1).SafeToString());
+            Assert.AreEqual("fghi", chunk.ExtractCell(1, 2).SafeToString());
+        }
+    }
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: 1. I'm not sure if it's possible to get an empty array but I'll add a test case for that
2. It should throw an error since it's expecting to parse a JSON file. I'll add a test case for that as well

sfc-gh-igarish (approved):
LGTM

Snowflake.Data.Tests/ChunkDeserializerTest.cs
sfc-gh-mhofman: imho lets maybe add this for now so we don't loose track on tests which rely on this

sfc-gh-ext-simba-lf: Alright, added

sfc-gh-mhofman (approved):

PR #686, opened 2023-06-22
SNOW-845774 - Add Jenkins user to allow list for cla assistant check

sfc-gh-anavalos:
CLA Assistant check has been failing for release branches. The failure is due the jenkins user used to create the branch and do the commit is not in the allow list in that check.
The fix is to add it.


sfc-gh-igarish (approved):
LGTM

Issue #653, opened 2023-05-12
UseProxy property in the connection string can be misleading


pucefti:
When you create the HttpClientHandler in HttpUtil, you don't set the UseProxy property and leave it to default, but by default the value is true.
It's obviously not a problem when we want to set a proxy or if your system hasn't a proxy configured, but when your system has a proxy configured by default and you don't want to use it, you have no way to do it because no matter if you specify in the connection string UseProxy to false or not, the proxy will still be used.

So I think that the value of the connection string's UseProxy should match the value of the HttpClientHandler, in order to be able to disable the proxy.

1. What version of .NET driver are you using?
2.0.23
   
2. What operating system and processor architecture are you using?
Windows server 2019
   
3. What version of .NET framework are you using?
.NET 6
   
4. What did you do?
In the connection string set UseProxy to false

5. What did you expect to see?
The connection is perform without using the proxy

6. Can you set logging to DEBUG and collect the logs?
Not relevant

sfc-gh-igarish:
I cannot reproduce the issue, I need more info from the reporter.
What I would ask about:

If you create a new session with a connection string, you can specify "useProxy=true" or "useProxy=false" and it works as expected. The default value for useProxy is false.

What are the exact steps to reproduce the issue?
How exactly do you create and configure HttpUtil object?
Do you create SFSession object by passing in the connection string useProxy parameter?

pucefti:
Hello,

Let me clarify, the problem is not in the UseProxy of the connection string which works as intended, but the way HttpClientHandler is build in HttpUtil.
For exemple:
I use this connectionstring:
account=test1;user=testuser;password=123;db=TEST_DB;useProxy=false;

With this one HttpClientHandler in HttpUtil will be construct as:
httpHandler = new HttpClientHandler()
{
	// Verify no certificates have been revoked
	CheckCertificateRevocationList = config.CrlCheckEnabled,
	// Enforce tls v1.2
	SslProtocols = SslProtocols.Tls12,
	AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
	UseCookies = false // Disable cookies
};

So with no proxy. But by default HttpClientHandler has his internal value UseProxy to true and so if a default proxy is configured in the system, it will use it even if you have specified in the connection string UseProxy=false, because this value is not passed to HttpClientHandler.

So, I think that the correct way to do it, is to add something like:
httpHandler = new HttpClientHandler()
{
[...]

PR #689, opened 2023-06-30
SNOW-828572 Set UseProxy on HttpClientHandler

sfc-gh-knozderko:
It is a bug fix that resolves the issue: https://github.com/snowflakedb/snowflake-connector-net/issues/653
The fix is about setting properly UseProxy attribute of HttpClientConfig to indicate whether a proxy is being used.

### DIFF
--- a/Snowflake.Data.Tests/HttpUtilTest.cs
+++ b/Snowflake.Data.Tests/HttpUtilTest.cs
@@ -47,7 +47,7 @@ namespace Snowflake.Data.Tests
                 true,
                 "snowflake.com",
                 "123",
-                "Bartek",
+                "testUser",
                 "proxyPassword",
                 "localhost", 
                 false,
--- a/Snowflake.Data.Tests/HttpUtilTest.cs
+++ b/Snowflake.Data.Tests/HttpUtilTest.cs
@@ -2,6 +2,8 @@
  * Copyright (c) 2022 Snowflake Computing Inc. All rights reserved.
  */
 
+using System.Net.Http;
+
 namespace Snowflake.Data.Tests
 {
     using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/HttpUtilTest.cs
```diff
@@ -36,5 +38,51 @@ public async Task TestIsRetryableHTTPCode(HttpStatusCode statusCode, bool forceR
 
             Assert.AreEqual(expectedIsRetryable, actualIsRetryable);
         }
+
+        [Test]
+        public void ShouldCreateHttpClientHandlerWithProxy()
+        {
+            // given
```
sfc-gh-dprzybysz, line 45: is it a new practice to write comments give/when/then

sfc-gh-knozderko: It is quite a new practice (previously I had added this convention in a few tests).
But even though previously it was not followed why not to introduce a good practice?

Snowflake.Data.Tests/HttpUtilTest.cs
```diff
@@ -36,5 +38,51 @@ public async Task TestIsRetryableHTTPCode(HttpStatusCode statusCode, bool forceR
...
+            var config = new HttpClientConfig(
+                true,
+                "snowflake.com",
+                "123",
+                "Bartek",
```
sfc-gh-dstempniak, line 50: maybe use "test" instead of Bartek :)

sfc-gh-igarish (approved):
LGTM

Snowflake.Data.Tests/HttpUtilTest.cs
sfc-gh-knozderko: changed

sfc-gh-mhofman (approved):

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -127,7 +128,8 @@ private HttpMessageHandler setupCustomHttpHandler(HttpClientConfig config)
                 httpHandler = new HttpClientHandler()
                 {
                     AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
-                    UseCookies = false // Disable cookies
+                    UseCookies = false, // Disable cookies
```
sfc-gh-mhofman, line 131: I would remove this comment

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -117,7 +117,8 @@ private HttpMessageHandler setupCustomHttpHandler(HttpClientConfig config)
                     // Enforce tls v1.2
                     SslProtocols = SslProtocols.Tls12,
                     AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
-                    UseCookies = false // Disable cookies
+                    UseCookies = false, // Disable cookies
```
sfc-gh-mhofman, line 120: I would remove this comment

PR #691, opened 2023-07-03
SNOW-835635 rollback session on connection pooled

sfc-gh-mhofman:
This change introduces an automated rollback during connection close and being pooled.
The rollback will be done on condition that a transaction got started with its constructor or BeginTransaction method of a connection (explicit transaction). If an error on such a rollback occurs the connection won't get added back to the pool to ensure that pool does never provide a dirty (with active transaction) connection.
In case of a transaction started with SnowflakeDbCommand the driver won't keep track of it.

### DIFF
--- a/Snowflake.Data.Tests/App.config
+++ b/Snowflake.Data.Tests/App.config
@@ -21,7 +21,7 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
       </layout>
       <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
         <layout type="log4net.Layout.PatternLayout">
-          <conversionPattern value="%date [%thread] [%-5level] [%method:%line] - %message%newline" />
+          <conversionPattern value="%date [%thread] [%-5level] [%ClassName:%line] - %message%newline" />
         </layout>
       </appender>
     </appender>
@@ -29,6 +29,9 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
     <root>
       <level value="ALL" />
       <appender-ref ref="RollingFileAppender" />
+    </root>    
+    <root>
+      <level value="WARN" />
       <appender-ref ref="ConsoleAppender" />
     </root>
   </log4net>
--- a/Snowflake.Data.Tests/App.config
+++ b/Snowflake.Data.Tests/App.config
@@ -8,7 +8,7 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
   </configSections>
   
   <log4net>
-    <appender name="MyRollingFileAppender" type="log4net.Appender.RollingFileAppender">
+    <appender name="RollingFileAppender" type="log4net.Appender.RollingFileAppender">
       <file type="log4net.Util.PatternString" value="test_%property{framework}.log" />
       <appendToFile value="true"/>
       <rollingStyle value="Size" />
@@ -19,11 +19,17 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
         <!-- <header value="[DateTime]  [Thread]  [Level]  [ClassName] Message&#13;&#10;" /> -->
         <conversionPattern value="[%date] [%t] [%-5level] [%logger] %message%newline" />
       </layout>
+      <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
+        <layout type="log4net.Layout.PatternLayout">
+          <conversionPattern value="%date [%thread] [%-5level] [%method:%line] - %message%newline" />
+        </layout>
+      </appender>
     </appender>
 
     <root>
       <level value="ALL" />
-      <appender-ref ref="MyRollingFileAppender" />
+      <appender-ref ref="RollingFileAppender" />
+      <appender-ref ref="ConsoleAppender" />
     </root>
   </log4net>
 
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -551,6 +551,7 @@ namespace Snowflake.Data.Tests
                 connection.Open();
                 firstOpenedSessionId = connection.SfSession.sessionId;
                 connection.BeginTransaction();
+                Assert.AreEqual(true, connection.HasActiveTransaction());
                 Assert.Throws<SnowflakeDbException>(() =>
                 {
                     using (var command = connection.CreateCommand())
@@ -567,14 +568,32 @@ namespace Snowflake.Data.Tests
                 connectionWithSessionReused.Open();
                 
                 Assert.AreEqual(firstOpenedSessionId, connectionWithSessionReused.SfSession.sessionId);
+                Assert.AreEqual(false, connectionWithSessionReused.HasActiveTransaction());
                 using (var cmd = connectionWithSessionReused.CreateCommand())
                 {
                     cmd.CommandText = "SELECT CURRENT_TRANSACTION()";
                     Assert.AreEqual(DBNull.Value, cmd.ExecuteScalar());
                 }
-            } 
+            }
             
             Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection should be reused and any pending transaction rolled back before it gets back to the pool");
+        }
+
+        [Test]
+        public void TestTransactionStatusNotTrackedForNonExplicitTransactionCalls()
+        {
+            SnowflakeDbConnectionPool.SetMaxPoolSize(1);
+            using (var connection = new SnowflakeDbConnection())
+            {
+                connection.ConnectionString = ConnectionString;
+                connection.Open();
+                using (var command = connection.CreateCommand())
+                {
+                    command.CommandText = "BEGIN"; // in general can be put as a part of a multi statement call and mixed with commit as well 
+                    command.ExecuteNonQuery();
+                    Assert.AreEqual(false, connection.HasActiveTransaction()); 
+                }
+            }
         }
 
         [Test]
@@ -589,11 +608,11 @@ namespace Snowflake.Data.Tests
                 connection1.ConnectionString = ConnectionString;
                 connection1.Open();
                 Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session is added to the pool after close connection");
+                connection1.BeginTransaction();
+                Assert.AreEqual(true, connection1.HasActiveTransaction());
                 using (var command = connection1.CreateCommand())
                 {
                     firstOpenedSessionId = connection1.SfSession.sessionId;
-                    command.CommandText = "BEGIN TRANSACTION";
-                    command.ExecuteNonQuery();
                     command.CommandText = "SELECT CURRENT_TRANSACTION()";
                     Assert.AreNotEqual(DBNull.Value, command.ExecuteScalar());
                 }
@@ -605,6 +624,7 @@ namespace Snowflake.Data.Tests
                 connection2.ConnectionString = ConnectionString;
                 connection2.Open();
                 Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session should be now removed from the pool");
+                Assert.AreEqual(false, connection2.HasActiveTransaction());
                 using (var command = connection2.CreateCommand())
                 {
                     Assert.AreEqual(firstOpenedSessionId, connection2.SfSession.sessionId);
@@ -633,7 +653,8 @@ namespace Snowflake.Data.Tests
                 connection.ConnectionString = ConnectionString;
                 connection.Open();
                 connection.BeginTransaction();
-                // no Rollback or Commit; during internal Rollback exception will be thrown
+                Assert.AreEqual(true, connection.HasActiveTransaction());
+                // no Rollback or Commit; during internal Rollback while closing a connection a mocked exception will be thrown
             }
             
             Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Should not return connection to the pool");
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -563,6 +563,20 @@ namespace Snowflake.Data.Tests
             Assert.AreEqual(DBNull.Value, command.ExecuteScalar());
             Assert.AreEqual(false, conn.HasTransactionInProgress());
             conn.Close();
+            
+            Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Should return connection to the pool");
+        }
+
+        private class TestSnowflakeDbFactory : DbProviderFactory 
+        {
+            public override SnowflakeDbCommand CreateCommand()
+            {
+                var commandThrowingExcOnlyForRollback = new Mock<SnowflakeDbCommand>().As<IDbCommand>();
+                commandThrowingExcOnlyForRollback.CallBase = true;
+                commandThrowingExcOnlyForRollback.SetupSet(it => it.CommandText = "ROLLBACK")
+                    .Throws(new SnowflakeDbException(SFError.INTERNAL_ERROR, "Unexpected failure on transaction rollback when connection is returned to the pool with pending transaction"));
+                return (SnowflakeDbCommand)commandThrowingExcOnlyForRollback.Object;
+            }
         }
 
         [Test]
@@ -571,32 +585,14 @@ namespace Snowflake.Data.Tests
             SnowflakeDbConnectionPool.SetPooling(true);
             SnowflakeDbConnectionPool.SetMaxPoolSize(10);
             Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize());
-            var connSpy = new Mock<SnowflakeDbConnection>().As<IDbConnection>();
-            connSpy.CallBase = true;
-            Mock<IDbCommand> mockDbCommand = new Mock<IDbCommand>();
-            mockDbCommand.SetupSequence(it => it.ExecuteNonQuery())
-                .Throws(new SnowflakeDbException(SFError.INTERNAL_ERROR, "syntax error"))
-                .Throws(new SnowflakeDbException(SFError.STATEMENT_ALREADY_RUNNING_QUERY, "error on transaction rollback"));
-            connSpy.SetupSequence(it => it.CreateCommand())
-                .CallBase() // BEGIN TRANS
-                .Returns(mockDbCommand.Object) // INVALID SQL
-                .CallBase() // for conn.Close() => GET TRANS ID
-                .Returns(mockDbCommand.Object); // for conn.Close() => ROLLBACK
-
-            var conn = connSpy.Object;
+            
+            var conn = new SnowflakeDbConnection();
+            conn.ProviderFactory = new TestSnowflakeDbFactory();
             conn.ConnectionString = ConnectionString;
             conn.Open();
             conn.BeginTransaction();
-            Assert.Throws<SnowflakeDbException>(() =>
-                {
-                    using (IDbCommand command = conn.CreateCommand())
-                    {
-                        command.CommandText = "INVALID SQL";
-                        command.ExecuteNonQuery();
-                    }
-                });
-            conn.Close();
-            
+            conn.Close(); // without Rollback or Commit; during internal Rollback exception will be thrown
+
             Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Should not return connection to the pool");
         }
 
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -588,6 +588,7 @@ namespace Snowflake.Data.Tests
             {
                 connection1.ConnectionString = ConnectionString;
                 connection1.Open();
+                Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session is added to the pool after close connection");
                 using (var command = connection1.CreateCommand())
                 {
                     firstOpenedSessionId = connection1.SfSession.sessionId;
@@ -603,6 +604,7 @@ namespace Snowflake.Data.Tests
             {
                 connection2.ConnectionString = ConnectionString;
                 connection2.Open();
+                Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session should be now removed from the pool");
                 using (var command = connection2.CreateCommand())
                 {
                     Assert.AreEqual(firstOpenedSessionId, connection2.SfSession.sessionId);
@@ -618,9 +620,15 @@ namespace Snowflake.Data.Tests
         {
             SnowflakeDbConnectionPool.SetPooling(true);
             SnowflakeDbConnectionPool.SetMaxPoolSize(10);
-            Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize());
+            var commandThrowingExceptionOnlyForRollback = new Mock<SnowflakeDbCommand>();
+            commandThrowingExceptionOnlyForRollback.CallBase = true;
+            commandThrowingExceptionOnlyForRollback.SetupSet(it => it.CommandText = "ROLLBACK")
+                .Throws(new SnowflakeDbException(SFError.INTERNAL_ERROR, "Unexpected failure on transaction rollback when connection is returned to the pool with pending transaction"));
+            var mockDbProviderFactory = new Mock<DbProviderFactory>();
+            mockDbProviderFactory.Setup(p => p.CreateCommand()).Returns(commandThrowingExceptionOnlyForRollback.Object);
 
-            using (var connection = new TestSnowflakeDbConnection())
+            Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize());
+            using (var connection = new TestSnowflakeDbConnection(mockDbProviderFactory.Object))
             {
                 connection.ConnectionString = ConnectionString;
                 connection.Open();
@@ -728,19 +736,12 @@ namespace Snowflake.Data.Tests
         
         private class TestSnowflakeDbConnection : SnowflakeDbConnection
         {
-            protected override DbProviderFactory DbProviderFactory => new TestSnowflakeDbFactory();
-        }
-        
-        private class TestSnowflakeDbFactory : DbProviderFactory 
-        {
-            public override DbCommand CreateCommand()
+            public TestSnowflakeDbConnection(DbProviderFactory dbProviderFactory)
             {
-                var commandThrowingExceptionOnlyForRollback = new Mock<SnowflakeDbCommand>().As<IDbCommand>();
-                commandThrowingExceptionOnlyForRollback.CallBase = true;
-                commandThrowingExceptionOnlyForRollback.SetupSet(it => it.CommandText = "ROLLBACK")
-                    .Throws(new SnowflakeDbException(SFError.INTERNAL_ERROR, "Unexpected failure on transaction rollback when connection is returned to the pool with pending transaction"));
-                return (DbCommand)commandThrowingExceptionOnlyForRollback.Object;
+                DbProviderFactory = dbProviderFactory;
             }
+
+            protected override DbProviderFactory DbProviderFactory { get; }
         }
     }
 }
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -551,7 +551,7 @@ namespace Snowflake.Data.Tests
                 connection.Open();
                 firstOpenedSessionId = connection.SfSession.sessionId;
                 connection.BeginTransaction();
-                Assert.AreEqual(true, connection.HasActiveTransaction());
+                Assert.AreEqual(true, connection.HasActiveExplicitTransaction());
                 Assert.Throws<SnowflakeDbException>(() =>
                 {
                     using (var command = connection.CreateCommand())
@@ -568,7 +568,7 @@ namespace Snowflake.Data.Tests
                 connectionWithSessionReused.Open();
                 
                 Assert.AreEqual(firstOpenedSessionId, connectionWithSessionReused.SfSession.sessionId);
-                Assert.AreEqual(false, connectionWithSessionReused.HasActiveTransaction());
+                Assert.AreEqual(false, connectionWithSessionReused.HasActiveExplicitTransaction());
                 using (var cmd = connectionWithSessionReused.CreateCommand())
                 {
                     cmd.CommandText = "SELECT CURRENT_TRANSACTION()";
@@ -591,7 +591,7 @@ namespace Snowflake.Data.Tests
                 {
                     command.CommandText = "BEGIN"; // in general can be put as a part of a multi statement call and mixed with commit as well 
                     command.ExecuteNonQuery();
-                    Assert.AreEqual(false, connection.HasActiveTransaction()); 
+                    Assert.AreEqual(false, connection.HasActiveExplicitTransaction()); 
                 }
             }
         }
@@ -609,7 +609,7 @@ namespace Snowflake.Data.Tests
                 connection1.Open();
                 Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session is added to the pool after close connection");
                 connection1.BeginTransaction();
-                Assert.AreEqual(true, connection1.HasActiveTransaction());
+                Assert.AreEqual(true, connection1.HasActiveExplicitTransaction());
                 using (var command = connection1.CreateCommand())
                 {
                     firstOpenedSessionId = connection1.SfSession.sessionId;
@@ -624,7 +624,7 @@ namespace Snowflake.Data.Tests
                 connection2.ConnectionString = ConnectionString;
                 connection2.Open();
                 Assert.AreEqual(0, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection session should be now removed from the pool");
-                Assert.AreEqual(false, connection2.HasActiveTransaction());
+                Assert.AreEqual(false, connection2.HasActiveExplicitTransaction());
                 using (var command = connection2.CreateCommand())
                 {
                     Assert.AreEqual(firstOpenedSessionId, connection2.SfSession.sessionId);
@@ -653,7 +653,7 @@ namespace Snowflake.Data.Tests
                 connection.ConnectionString = ConnectionString;
                 connection.Open();
                 connection.BeginTransaction();
-                Assert.AreEqual(true, connection.HasActiveTransaction());
+                Assert.AreEqual(true, connection.HasActiveExplicitTransaction());
                 // no Rollback or Commit; during internal Rollback while closing a connection a mocked exception will be thrown
             }
             
--- a/Snowflake.Data.Tests/SFConnectionPoolT.cs
+++ b/Snowflake.Data.Tests/SFConnectionPoolT.cs
@@ -15,44 +15,47 @@ namespace Snowflake.Data.Tests
     using Snowflake.Data.Tests.Mock;
     using System.Data.Common;
     using Moq;
-
     
-    readonly struct PoolConfig {
-        public bool Pooling { get; }
-        public long Timeout { get; }
-        public int MaxPoolSize { get; }
+    class PoolConfig {
+        private readonly bool _pooling;
+        private readonly long _timeout;
+        private readonly int _maxPoolSize;
 
-        public PoolConfig(int maxPoolSize, long timeout, bool poolingEnabled)
+        public PoolConfig()
         {
-            MaxPoolSize = maxPoolSize;
-            Timeout = timeout;
-            Pooling = poolingEnabled;
+            _maxPoolSize = SnowflakeDbConnectionPool.GetMaxPoolSize();
+            _timeout = SnowflakeDbConnectionPool.GetTimeout();
+            _pooling = SnowflakeDbConnectionPool.GetPooling();
+        }
+
+        public void Reset()
+        {
+            SnowflakeDbConnectionPool.SetMaxPoolSize(_maxPoolSize);
+            SnowflakeDbConnectionPool.SetTimeout(_timeout);
+            SnowflakeDbConnectionPool.SetPooling(_pooling);
         }
     }
     
     [TestFixture, NonParallelizable]
-    // [Apartment(ApartmentState.STA)]
     class SFConnectionPoolT : SFBaseTest
     {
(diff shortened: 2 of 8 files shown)


sfc-gh-igarish:
Please add description.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
             ConcurrentPoolingAsyncHelper(connStr, true);
         }
 
+        [Test]
+        public void TestRollbackTransactionOnPooledWhenExceptionOccurred()
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
```
sfc-gh-dprzybysz, line 510: those parameters are static? who cleans them between test methods?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+            SnowflakeDbConnectionPool.SetMaxPoolSize(10);
+
+            object firstOpenedSession = null;
+            using (var connection = new SnowflakeDbConnection())
+            {
+                connection.ConnectionString = ConnectionString;
+                connection.Open();
+                firstOpenedSession = connection.SfSession.sessionId;
+                connection.BeginTransaction();
+                Assert.Throws<SnowflakeDbException>(() =>
```
sfc-gh-dprzybysz, line 520: should we check also the message of the exception?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+                        command.ExecuteNonQuery();
+                    }
+                });
+            }
+
+            using (var connectionWithSessionReused = new SnowflakeDbConnection())
+            {
+                connectionWithSessionReused.ConnectionString = ConnectionString;
+                connectionWithSessionReused.Open();
+                Assert.AreEqual(firstOpenedSession, connectionWithSessionReused.SfSession.sessionId);
```
sfc-gh-dprzybysz, line 534: pool has 10 elements - are we sure that it will be always the same connection?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+                using (var cmd = connectionWithSessionReused.CreateCommand())
+                {
+                    cmd.CommandText = "SELECT CURRENT_TRANSACTION()";
+                    Assert.AreEqual(DBNull.Value, cmd.ExecuteScalar());
+                    Assert.AreEqual(false, connectionWithSessionReused.HasTransactionInProgress());
+                }
+                connectionWithSessionReused.Close();
```
sfc-gh-dprzybysz, line 541: isn't it close via using in line 530?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+            }
+        }
+
+        [Test]
+        public void TestRollbackTransactionOnPooledWhenConnectionClose()
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnectionPool.SetMaxPoolSize(10);
+            var conn = new SnowflakeDbConnection();
```
sfc-gh-dprzybysz, line 550: why don't we use using here?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -22,6 +22,14 @@ public class SnowflakeDbConnection : DbConnection
 
         internal ConnectionState _connectionState;
 
+        private DbProviderFactory _snowflakeDbFactory = new SnowflakeDbFactory();
```
sfc-gh-dprzybysz, line 25: why don't we call private field `_ dbProviderFactory` but with the name of specific class?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
             }
         }
 
-        public override string ServerVersion => _connectionState == ConnectionState.Open ? SfSession.serverVersion : "";
+        public override string ServerVersion => IsOpen() ? SfSession.serverVersion : "";
 
         public override ConnectionState State => _connectionState;
 
+        private object? GetTransactionId()
```
sfc-gh-dprzybysz, line 106: are we able to better name the returned type?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
...
+                {
+                    command.CommandText = "ROLLBACK";
+                    command.ExecuteNonQuery(); 
+                    logger.Warn("Rolled back transaction: " + transactionInProgressId + " in session: " + SfSession.sessionId);
+                    return true;
+                }
+            }
+            catch (SnowflakeDbException exception)
+            {
+                logger.Error("Unable to rollback transaction: " + transactionInProgressId + " in session: " + SfSession.sessionId + ", exception: " + exception.Message);
```
sfc-gh-dprzybysz, line 137: should we pass exception to logging method? We are loosing stack trace now?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
...
+        {
+            if (transactionInProgressId == DBNull.Value)
+                return false;
+            try
+            {
+                using (IDbCommand command = CreateCommand())
+                {
+                    command.CommandText = "ROLLBACK";
+                    command.ExecuteNonQuery(); 
+                    logger.Warn("Rolled back transaction: " + transactionInProgressId + " in session: " + SfSession.sessionId);
```
sfc-gh-dprzybysz, line 131: why is it warn? someone ordered transaction termination on purpose

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -134,9 +189,16 @@ public Task CloseAsync(CancellationToken cancellationToken)
             }
             else
             {
-                if ((_connectionState != ConnectionState.Closed) && SfSession != null)
+                if (IsNonClosedWithSession())
                 {
-                    if (SnowflakeDbConnectionPool.addSession(SfSession))
+                    var transactionInProgress = GetTransactionId();
```
sfc-gh-dprzybysz, line 194: isn't it duplicated code with this one some lines above? are we able to extract method?

Snowflake.Data.Tests/App.config
```diff
@@ -19,11 +19,17 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
...
       </layout>
+      <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
+        <layout type="log4net.Layout.PatternLayout">
+          <conversionPattern value="%date [%thread] [%-5level] [%method:%line] - %message%newline" />
+        </layout>
+      </appender>
     </appender>
 
     <root>
       <level value="ALL" />
```
sfc-gh-pfus, line 30: If we configure console logging for tests, maybe we can change here to info or, even better, to warn? If anyone needs more logging, always can be changed.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnectionPool.SetMaxPoolSize(10);
+
+            object firstOpenedSession = null;
+            using (var connection = new SnowflakeDbConnection())
+            {
+                connection.ConnectionString = ConnectionString;
+                connection.Open();
+                firstOpenedSession = connection.SfSession.sessionId;
```
sfc-gh-pfus, line 518: It's not session, it's sessionId?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+                {
+                    using (var command = connection.CreateCommand())
+                    {
+                        command.CommandText = "invalid command will throw";
+                        command.ExecuteNonQuery();
+                    }
+                });
+            }
+
+            using (var connectionWithSessionReused = new SnowflakeDbConnection())
```
sfc-gh-pfus, line 530: Are we sure that this is the same session reused and not some random one? This strongly depends on algorithm of choosing sessions, which is not a part of this test I suppose. Small suggestion - maybe we can decrease maxPoolSize to 1 to fix my concerns?

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-pfus, line 534: Thanks for this assert, but still - it would be more maintainable if this pool is 1-sized.

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-pfus, line 541: If assert above fails, this will never happen. Except for - can we use `using` and close automatically?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnectionPool.SetMaxPoolSize(10);
+            var conn = new SnowflakeDbConnection();
+            conn.ConnectionString = ConnectionString;
+            
+            conn.Open();
+            IDbCommand command = conn.CreateCommand();
+            object firstOpenedSession = conn.SfSession.sessionId;
+            command.CommandText = "BEGIN TRANSACTION ";
```
sfc-gh-pfus, line 556: Can we strip this string?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+                }
+                connectionWithSessionReused.Close();
+            }
+        }
+
+        [Test]
+        public void TestRollbackTransactionOnPooledWhenConnectionClose()
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnectionPool.SetMaxPoolSize(10);
```
sfc-gh-pfus, line 549: Same comment.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+            IDbCommand command = conn.CreateCommand();
+            object firstOpenedSession = conn.SfSession.sessionId;
+            command.CommandText = "BEGIN TRANSACTION ";
+            command.ExecuteNonQuery();
+            conn.Close();
+            
+            conn.Open();
+            Assert.AreEqual(firstOpenedSession, conn.SfSession.sessionId);
+            command.CommandText = "SELECT CURRENT_TRANSACTION()";
+            Assert.AreEqual(DBNull.Value, command.ExecuteScalar());
```
sfc-gh-pfus, line 563: I would move asserts to `using`.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+            conn.Close();
+            
+            Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Should return connection to the pool");
+        }
+
+        private class TestSnowflakeDbFactory : DbProviderFactory 
+        {
+            public override SnowflakeDbCommand CreateCommand()
+            {
+                var commandThrowingExcOnlyForRollback = new Mock<SnowflakeDbCommand>().As<IDbCommand>();
```
sfc-gh-pfus, line 574: `Exc` is a C# standard? I would use `Ex`.

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +504,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+            Assert.AreEqual(firstOpenedSession, conn.SfSession.sessionId);
+            command.CommandText = "SELECT CURRENT_TRANSACTION()";
+            Assert.AreEqual(DBNull.Value, command.ExecuteScalar());
+            Assert.AreEqual(false, conn.HasTransactionInProgress());
+            conn.Close();
+            
+            Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Should return connection to the pool");
+        }
+
+        private class TestSnowflakeDbFactory : DbProviderFactory
```
sfc-gh-pfus, line 570: Is this a c# standard, that private classes are anywhere in the file?

sfc-gh-pfus, line 570: Also, I would say, that I have always seen mock of mock, as creating mock which creates mock using library :) Like:
```
factory = Mock<Factory>
command = Mock<Command>
factory.createCommand(return command)
```
WDYT?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -22,6 +22,14 @@ public class SnowflakeDbConnection : DbConnection
 
         internal ConnectionState _connectionState;
 
+        private DbProviderFactory _snowflakeDbFactory = new SnowflakeDbFactory();
+
+        internal DbProviderFactory ProviderFactory
```
sfc-gh-pfus, line 27: Do we need it? Can't we use internal property interchangably?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -61,10 +69,15 @@ public SecureString Password
 
         public bool IsOpen()
         {
-            return _connectionState == ConnectionState.Open;
+            return _connectionState == ConnectionState.Open && SfSession != null;
```
sfc-gh-pfus, line 72: What is the case when state is open and session is null?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
             }
         }
 
-        public override string ServerVersion => _connectionState == ConnectionState.Open ? SfSession.serverVersion : "";
+        public override string ServerVersion => IsOpen() ? SfSession.serverVersion : "";
```
sfc-gh-pfus, line 102: You use above `string.Empty`, why `""` here?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -61,10 +69,15 @@ public SecureString Password
...
+            return _connectionState == ConnectionState.Open && SfSession != null;
         }
 
-        public override string Database => _connectionState == ConnectionState.Open ? SfSession.database : string.Empty;
+        private bool IsNonClosedWithSession()
+        {
+            return _connectionState != ConnectionState.Closed && SfSession != null;
+        }
+
+        public override string Database => IsOpen() ? SfSession.database : string.Empty;
```
sfc-gh-pfus, line 80: Why empty string instead of null?

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus, line 106: What do you think about GHActions comments below?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
...
 
+        private object? GetTransactionId()
+        {
+            if (!IsOpen())
+            {
+                return DBNull.Value;
+            }
+            using (IDbCommand command = CreateCommand())
+            {
+                command.CommandText = "SELECT CURRENT_TRANSACTION()";
```
sfc-gh-pfus, line 114: We do a roundtrip to SF here every time this function is invoked. I would say, that I would rather expect the `GetTransactionId` call did not go to server. But I'm not fully convinced, maybe this is a premature optimization...

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +99,53 @@ public override string DataSource
...
+                return command.ExecuteScalar(); 
+            }
+        }
+
+        public bool HasTransactionInProgress() => GetTransactionId() != DBNull.Value;
```
sfc-gh-pfus, line 119: `HasActiveTransaction`? WDYT?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -134,9 +189,16 @@ public Task CloseAsync(CancellationToken cancellationToken)
             }
             else
             {
-                if ((_connectionState != ConnectionState.Closed) && SfSession != null)
+                if (IsNonClosedWithSession())
```
sfc-gh-pfus, line 192: This looks similar to the above. Maybe some extract function?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -106,10 +155,16 @@ public override void ChangeDatabase(string databaseName)
...
             logger.Debug("Close Connection.");
-            if ((_connectionState != ConnectionState.Closed) &&
-                (SfSession != null))
+            if (IsNonClosedWithSession())
             {
-                if (SnowflakeDbConnectionPool.addSession(SfSession))
+                var transactionInProgress = GetTransactionId();
+                if (transactionInProgress != DBNull.Value)
+                {
+                    if (TerminateTransaction(transactionInProgress))
```
sfc-gh-pfus, line 163: Ok, I remember our conversation about that topic. But what do you think about throwing exception after rolling back if someone forgot to end transaction? Fail fast... We can discuss it today.

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-dstempniak: Pooling is turned on by default. Maybe we should use SetUp and TearDown to store are restore pooling attribute?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -12,12 +12,12 @@ namespace Snowflake.Data.Tests
...
     using Snowflake.Data.Log;
-    using System.Diagnostics;
     using Snowflake.Data.Tests.Mock;
-    using System.Runtime.InteropServices;
     using System.Data.Common;
+    using Moq;
 
-    [TestFixture]
+    [TestFixture, NonParallelizable]
+    [Apartment(ApartmentState.STA)]
```
sfc-gh-mhofman, line 20: own observation - to fix class name

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: will do

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: I will set pool size to one as @sfc-gh-pfus  gave good suggestion on this. This way even if we change future logic of the pool reusing a single connection should be covered in this test.

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: yep, fixed

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: fixed according to your suggestion @sfc-gh-pfus, thx!

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: of course, removing this

sfc-gh-mhofman: yes, @sfc-gh-dprzybysz  actually pointed that out, fixing this

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: fixed to have pool size = 1

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: If that kind of opening a connection is possible within a language some tests can be/should be written differently? It's not a recommended way but at least possible and provided in the readme along with using pattern.
But going this way I'm wondering why don't we have it further which would lead to:
using (conn)
  using (transaction)
     using (command)

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: Since it's entirely for this particular test I would keep it this way but when tends to be widely used I would move it out, ok?

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: right, but actually I'm removing since DbProviderFactory is already in the base class

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Basically whenever IsOpen was used code actually looked into underlying SfSession so introducing this additional check is also a protection against NPE and is just a refactor

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: better, thanks

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: The problem is not only user can run SnowflakeDbConnection.BeginDbTransaction but can also create a DbCommand, provide it's CommandText = "BEGIN" and trigger the command without this operation outcome being noticed by SnowflakeDbConnection.
I had same worries here but stateful connection might be a thing to discuss in the future.
We need also to remember that right now there is just one transaction at a time available but introducing a new feature in the Snowflake DB can lead to related changes in the SnowflakeDbConnection if we make it stateful. Not sure about this idea then.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: It's a private operation used within closing a connection. If it was a publicly available interace I would agree with you but right now this is just to point out for a user that he's returning connection to the pool without calling ROLLBACK/COMMIT on his own - that's why it's a WARN.
Maybe I change a name of this function like: TerminateTransactionForDirtyConnectionClose?

sfc-gh-mhofman: I changed messages and TerminateTransaction name, check if this is clear now

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: We should see just the reason for not terminating transaction during "dirty" connection being closed.
Either way this operation succeeds (and connection gets its rollback and returns to the pool) or fails (connection gets destroyed with underlying http session and doesn't return to the pool.
We can care here for the stack trace but user should worry about that what was suggested in the ticket.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Check my reply to @sfc-gh-dprzybysz comments

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Again - baby steps. I don't want to do a huge refactor so everyone would be confused

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: There's a plan to have bigger changes in the connection pool. That's why I just wanted to do minimal changes here

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: And this type should be?

sfc-gh-mhofman: Maybe string, not sure of the CurrentTransaction return type, maybe some Int64 would suite???

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus: I see your concerns, but on the other hand - it's private implementation, even if something changes we can refactor it. I thought about caching this value after first call to getting transaction id and clearing on transaction end. See, my main concern is that it's easy for instance to have a client write code like this:

```
for (record in records) {
    log.debug("session: {}, transactionId: {}, record:", getSessionId(), getTransactionId(), record)
}
```

And now, even if debug is not logged, we still have a lot of roundtrips.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus: String, int, whatever :)

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-pfus: Ok, fair enough.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus: On the other hand, isn't it a code that you just introduce? :)

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus: On the other hand - logs without stack traces are good if this is the only place from which this message can be written (like logs specific for some domain logic etc). Now you have a very generic method which may be closed from multiple places, and based on the message itself user may have difficulties in locating which function/operation/query/whatever caused the error. If we have a stack trace here, it would be easier to know. What do you think?

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-pfus: Ok, fair enough.

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-dprzybysz: Mock returning mock is very often considered as antipattern :)

sfc-gh-pfus: Still, it's better than mixing library mocks and custom code mocks IMHO :)

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -12,16 +12,47 @@ namespace Snowflake.Data.Tests
...
+        public PoolConfig(int maxPoolSize, long timeout, bool poolingEnabled)
+        {
+            MaxPoolSize = maxPoolSize;
+            Timeout = timeout;
+            Pooling = poolingEnabled;
+        }
+    }
+    
+    [TestFixture, NonParallelizable]
+    // [Apartment(ApartmentState.STA)]
```
sfc-gh-dprzybysz, line 34: can we get rid of commented code?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -12,16 +12,47 @@ namespace Snowflake.Data.Tests
...
     class SFConnectionPoolT : SFBaseTest
     {
         private static SFLogger logger = SFLoggerFactory.GetLogger<SFConnectionPoolT>();
+        private PoolConfig previousPoolConfig;
 
+        [SetUp]
+        public void BeforeTest()
+        {
+            previousPoolConfig = new PoolConfig(SnowflakeDbConnectionPool.GetMaxPoolSize(),
```
sfc-gh-dprzybysz, line 43: you can create previous version once in BeforeAll and save in static field?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -12,16 +12,47 @@ namespace Snowflake.Data.Tests
...
     using Snowflake.Data.Log;
-    using System.Diagnostics;
     using Snowflake.Data.Tests.Mock;
-    using System.Runtime.InteropServices;
     using System.Data.Common;
+    using Moq;
 
-    [TestFixture]
+    
+    readonly struct PoolConfig {
```
sfc-gh-dprzybysz, line 20: 1. it can be a class
2. it can has parameterless constructor reading defaults from SnowflakeDbConnectionPool once before all tests
3. it can have method `setDefaultSnowflakeDbConnectionPool` which work on SnowflakeDbConnectionPool
4. then all the fields could be private

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +541,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
 
+        [Test]
+        public void TestRollbackTransactionOnPooledWhenExceptionOccurred()
+        {
+            SnowflakeDbConnectionPool.SetMaxPoolSize(1);
+
+            object firstOpenedSessionId = null;
+            using (var connection = new SnowflakeDbConnection())
+            {
+                connection.ConnectionString = ConnectionString;
```
sfc-gh-dprzybysz, line 552: just curious why we cannot pass connection string in `SnowflakeDbConnection` constructor?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -504,6 +541,98 @@ public void TestConcurrentConnectionPoolingAsync()
...
+                using (var command = connection2.CreateCommand())
+                {
+                    Assert.AreEqual(firstOpenedSessionId, connection2.SfSession.sessionId);
+                    command.CommandText = "SELECT CURRENT_TRANSACTION()";
+                    Assert.AreEqual(DBNull.Value, command.ExecuteScalar());
+                    Assert.AreEqual(false, connection2.HasActiveTransaction());
+                }
+            }
+
+            Assert.AreEqual(1, SnowflakeDbConnectionPool.GetCurrentPoolSize(), "Connection should be returned to the pool");
```
sfc-gh-dprzybysz, line 615: the second connection came back to pool? what about first connection that has active transaction? in line ~601 we should have some assertion of pool size?

Snowflake.Data.Tests/SFConnectionPoolT.cs
```diff
@@ -598,5 +727,35 @@ static async Task InvalidConnectionTaskAsync(string connectionString)
...
+            {
+                var commandThrowingExceptionOnlyForRollback = new Mock<SnowflakeDbCommand>().As<IDbCommand>();
+                commandThrowingExceptionOnlyForRollback.CallBase = true;
+                commandThrowingExceptionOnlyForRollback.SetupSet(it => it.CommandText = "ROLLBACK")
+                    .Throws(new SnowflakeDbException(SFError.INTERNAL_ERROR, "Unexpected failure on transaction rollback when connection is returned to the pool with pending transaction"));
+                return (SnowflakeDbCommand)commandThrowingExceptionOnlyForRollback.Object;
+            }
+        }
+
+        private readonly struct PoolConfig {
```
sfc-gh-dprzybysz, line 748: this struct was already defined in different file in this review - why don't reuse it if necessary?

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Done

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: added new asserts

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Transaction can be started without SnowflakeDbConnection being aware of it. Won't do it this way

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Already refactored.

sfc-gh-dprzybysz (approved):

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +93,55 @@ public override string DataSource
...
+            {
+                command.CommandText = "SELECT CURRENT_TRANSACTION()";
+                var result = command.ExecuteScalar();
+                return result == null || result == DBNull.Value ? null : result.ToString();
+            }
+        }
+
+        internal bool HasActiveTransaction() => GetTransactionId() != null;
+        
+        private bool TerminateTransactionForDirtyConnectionReturningToPool()
```
sfc-gh-igarish, line 114: Can we comment on method what's meaning of true/false return value?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -314,7 +362,9 @@ protected override DbTransaction BeginDbTransaction(IsolationLevel isolationLeve
 
         protected override DbCommand CreateDbCommand()
         {
-            return new SnowflakeDbCommand(this);
+            var command = DbProviderFactory.CreateCommand();
```
sfc-gh-igarish, line 365: what is the relation of this code change with the connection pool change?

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -86,17 +93,55 @@ public override string DataSource
             }
         }
 
-        public override string ServerVersion => _connectionState == ConnectionState.Open ? SfSession.serverVersion : "";
+        public override string ServerVersion => IsOpen() ? SfSession.serverVersion : String.Empty;
 
         public override ConnectionState State => _connectionState;
 
+        private string GetTransactionId()
```
sfc-gh-igarish, line 100: will this cause delay in closing the connection? Is there a way we can keep transaction status in the driver without executing a SQL command? Will this execute only when connection pool is enable?

sfc-gh-mhofman: I'll change it to execute only when pool is enabled.
It does introduce a delay since there is a call to the database. In theory we could avoid it and store the transaction reference in the connection. However, if one decides to start a transaction with SnowflakeDbConnection.ExecuteNonQuery the connection will never be aware of such a transaction nor have a reference to the transaction object. 
We could of course track such commands within ExecuteNonQuery and some regexp and force users to use SnowflakeDbConnection.BeginDbTransaction but it would affect current apps as the breaking change.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: allows to inject a different objects factory in the tests to provide mocked commands and test edge case scenarios with exceptions thrown during rollback of transaction

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Keeping transaction status is a tricky thing. We can think about it in the future.

sfc-gh-igarish: Can we at least check for auto commit on/off and if auto commit is ON then no need to check for txn id or rolling back the transaction. Most of the driver has API for auto commit ON or OFF. e.g. in JDBC it's setAutoCommit(boolean) API.

sfc-gh-mhofman: When autocommit is enabled and user starts an explicit transaction we might still need to check if this has been rolled back or committed:
"autocommit determines whether a DML statement, when executed without an active transaction, is automatically committed after the statement successfully completes. ..."
Autocommit can be also driven by account settings in which case we need to query database to get this setting as the state might not always be achieved with connection string/explicit statement?

sfc-gh-mhofman: As a result of our discussions only ROLLBACK will be submitted. I removed this method.

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -35,6 +37,14 @@ public class SnowflakeDbConnection : DbConnection
         // Will fix that in a separated PR though as it's a different issue
         private static Boolean _isArrayBindStageCreated;
 
+        private enum TransactionRollbackStatus
+        {
+            Undefined,
```
sfc-gh-igarish, line 42: do we need Undefined and NotNeeded status? If yes, please put comment when it's going to use.

sfc-gh-igarish (approved):
LGTM. Nit: Cleanup TransactionRollbackStatus enum.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-mhofman: Right now just one, I will comment on the value

Snowflake.Data.Tests/SFConnectionPoolT.cs
sfc-gh-mhofman: It's actually both - factory and DbCommand a spied object because we need to check specific behaviour during integration test :(

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -103,22 +104,31 @@ public override string DataSource
         public override string ServerVersion => IsOpen() ? SfSession.serverVersion : String.Empty;
 
         public override ConnectionState State => _connectionState;
+        internal SnowflakeDbTransaction ExplicitTransaction { get; set; } // tracks only explicit transaction operations
+        
+        internal bool HasActiveTransaction() => ExplicitTransaction != null && ExplicitTransaction.IsActive;
```
sfc-gh-igarish, line 109: Can we rename as HasActiveExplicitTransaction?

PR #692, opened 2023-07-05
Add tests for ConcatenatedStream and FastMemoryStream

sfc-gh-ext-simba-lf:
Regarding issue 113

The PR adds tests for ConcatenatedStream and FastMemoryStream to increase code coverage

### DIFF
--- a/Snowflake.Data.Tests/ConcatenatedStreamTest.cs
+++ b/Snowflake.Data.Tests/ConcatenatedStreamTest.cs
@@ -72,7 +72,7 @@ namespace Snowflake.Data.Tests
         {
             try
             {
-                long l = concatStream.Length;
+                _ = concatStream.Length;
                 Assert.Fail("A NotImplementedException should've been thrown");
             }
             catch (Exception ex)
@@ -96,7 +96,7 @@ namespace Snowflake.Data.Tests
 
             try
             {
-                long l = concatStream.Position;
+                _ = concatStream.Position;
                 Assert.Fail("A NotImplementedException should've been thrown");
             }
             catch (Exception ex)
--- a/Snowflake.Data.Tests/ConcatenatedStreamTest.cs
+++ b/Snowflake.Data.Tests/ConcatenatedStreamTest.cs
@@ -1,0 +1,159 @@
+﻿/*
+ * Copyright (c) 2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+namespace Snowflake.Data.Tests
+{
+    using NUnit.Framework;
+    using Snowflake.Data.Core;
+    using System;
+    using System.IO;
+    using System.Text;
+
+    [TestFixture]
+    class ConcatenatedStreamTest
+    {
+        Stream concatStream;
+
+        [SetUp]
+        public void BeforeTest()
+        {
+            string data = "12345678";
+            byte[] bytes = Encoding.UTF8.GetBytes(data);
+            Stream stream = new MemoryStream(bytes);
+
+            Stream openBracket = new MemoryStream(Encoding.UTF8.GetBytes(data));
+            Stream closeBracket = new MemoryStream(Encoding.UTF8.GetBytes(data));
+
+            concatStream = new ConcatenatedStream(new Stream[3] { openBracket, stream, closeBracket });
+        }
+
+        [Test]
+        [Ignore("ConcatenatedStreamTest")]
+        public void ConcatenatedStreamTestDone()
+        {
+            // Do nothing;
+        }
+
+        [Test]
+        public void TestCanRead()
+        {
+            Assert.IsTrue(concatStream.CanRead);
+        }
+
+        [Test]
+        public void TestCanSeek()
+        {
+            Assert.IsFalse(concatStream.CanSeek);
+        }
+
+        [Test]
+        public void TestCanWrite()
+        {
+            Assert.IsFalse(concatStream.CanWrite);
+        }
+
+        [Test]
+        public void TestFlush()
+        {
+            try
+            {
+                concatStream.Flush();
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestLength()
+        {
+            try
+            {
+                long l = concatStream.Length;
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestPosition()
+        {
+            try
+            {
+                concatStream.Position = 1;
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+
+            try
+            {
+                long l = concatStream.Position;
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestSeek()
+        {
+            try
+            {
+                concatStream.Seek(1, new SeekOrigin());
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestSetLength()
+        {
+            try
+            {
+                concatStream.SetLength(1);
+            }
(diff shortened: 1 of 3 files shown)


Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            Assert.IsFalse(concatStream.CanWrite);
+        }
+
+        [Test]
+        public void TestFlush()
+        {
+            try
+            {
+                concatStream.Flush();
```
sfc-gh-dprzybysz, line 61: if we expect this line to throw exception then just after it we should add always failing assertion because if that line won't fail then the test will be green

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,159 @@
...
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestLength()
+        {
+            try
+            {
+                long l = concatStream.Length;
```
sfc-gh-dprzybysz, line 74: do we need temporary variable?
we should have always failing assertion at the end of try block - the same problem happens also in the tests below

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            try
+            {
+                concatStream.Position = 1;
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+
+            try
```
sfc-gh-dprzybysz, line 94: should we split this test?

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            Assert.AreEqual(0, concatStream.Read(buffer, 0, 0)); // Read 0 byte
+        }
+
+        [Test]
+        public void TestReadBytes()
+        {
+            byte[] buffer = new byte[3];
+            Assert.AreEqual(1, concatStream.Read(buffer, 0, 1)); // Read 1 byte
+            Assert.AreEqual(2, concatStream.Read(buffer, 0, 2)); // Read 2 bytes
+            Assert.AreEqual(3, concatStream.Read(buffer, 0, 3)); // Read 3 bytes
```
sfc-gh-dprzybysz, line 156: do we have a test where we try to read 4 bytes?

Snowflake.Data.Tests/FastMemoryStreamTest.cs
```diff
@@ -0,0 +1,77 @@
...
+        [Ignore("FastMemoryStreamTest")]
+        public void FastMemoryStreamTestDone()
+        {
+            // Do nothing;
+        }
+
+        [Test]
+        public void TestDefaultValues()
+        {
+            // FastMemoryStream defaults to size of 0 and buffer of size 256
```
sfc-gh-dprzybysz, line 32: is 256 defined somewhere as constant? can we reuse such constant then?

Snowflake.Data.Tests/FastMemoryStreamTest.cs
```diff
@@ -0,0 +1,77 @@
...
+                fastMemoryStream.WriteByte(val);
+            }
+
+            Assert.AreEqual(byteArr, fastMemoryStream.GetBuffer());
+        }
+
+        [Test]
+        public void TestClear()
+        {
+            fastMemoryStream.Clear();
```
sfc-gh-dprzybysz, line 56: should we write anything to the stream before to test if clear method clears the stream?

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-dprzybysz: in other tests I saw usage of `Assert.Throws<SpecificExceptionClass>` - maybe this method can be used instead of explicit try/catch?

sfc-gh-ext-simba-lf: Ok

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-ext-simba-lf: Removed the temporary variable and added `Assert.Fail` at the end of try block for all the relevant tests

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-ext-simba-lf: Ok

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-ext-simba-lf: I'll add a test for that

Snowflake.Data.Tests/FastMemoryStreamTest.cs
sfc-gh-ext-simba-lf: There was no constant but I made one for it

Snowflake.Data.Tests/FastMemoryStreamTest.cs
sfc-gh-ext-simba-lf: I'll add that to the test

sfc-gh-igarish (approved):
LGTM. Please resolve all conversation as such it's a merge gate.

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,189 @@
...
+    using NUnit.Framework;
+    using Snowflake.Data.Core;
+    using System;
+    using System.IO;
+    using System.Text;
+
+    [TestFixture]
+    class ConcatenatedStreamTest
+    {
+        Stream concatStream;
```
sfc-gh-dstempniak, line 16: maybe change the name of private member to _ concatStream

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-dstempniak: instead of try..catch it rather should be something like:
Assert.Throws< NotImplementedException>(() => concatStream.Flush());

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
```diff
@@ -0,0 +1,189 @@
...
+            catch (Exception ex)
+            {
+                Assert.IsTrue(ex is NotImplementedException);
+            }
+        }
+
+        [Test]
+        public void TestLength()
+        {
+            try
```
sfc-gh-dstempniak, line 73: `Assert.Throws<NotImplementedException>(() => _ = concatStream.Length);`
and so on...

Snowflake.Data.Tests/FastMemoryStreamTest.cs
```diff
@@ -0,0 +1,87 @@
...
+namespace Snowflake.Data.Tests
+{
+    using NUnit.Framework;
+    using Snowflake.Data.Core;
+    using System.Linq;
+
+    [TestFixture]
+    class FastMemoryStreamTest
+    {
+        FastMemoryStream fastMemoryStream;
```
sfc-gh-dstempniak, line 14: maybe _fastMemoryStream

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-ext-simba-lf: Ok

Snowflake.Data.Tests/ConcatenatedStreamTest.cs
sfc-gh-ext-simba-lf: Replaced the try catch with Assert.Throws

Snowflake.Data.Tests/FastMemoryStreamTest.cs
sfc-gh-ext-simba-lf: Ok

PR #703, opened 2023-07-17
SNOW-861926 Build and test parallelisation and cleanup

sfc-gh-pbulawa:
Prepare the test code to be able to run in parallel and enable the parallelisation within GitHub actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,6 @@ jobs:
     name: Tests on Windows
     runs-on: windows-latest
     strategy:
-      max-parallel: 1
       matrix:
         #dotnet: ['2.2.402']
         dotnet: ['net6.0', 'net472', 'net471']
@@ -135,7 +134,6 @@ jobs:
     name: Tests on Linux
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 1
       matrix:
         #dotnet: ['2.2.402']
         dotnet: ['net6.0']
@@ -175,7 +173,6 @@ jobs:
     name: Tests on MAC
     runs-on: macos-latest
     strategy:
-      max-parallel: 1
       matrix:
         #dotnet: ['2.2.402']
         dotnet: ['net6.0']
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,7 +83,7 @@ jobs:
     - name: Build with MSBuild
       run: msbuild snowflake-connector-net.sln -p:Configuration=Release
 
-  test:
+  test-windows:
     needs: [build-net, build-framework]
     name: Tests on Windows
     runs-on: windows-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,6 @@
-# This is a basic workflow to help you get started with Actions
-
 name: DotNet Build and Test
 
-# Controls when the action will run. Triggers the workflow on push or pull request
+# Triggers the workflow on push or pull request
 # events but only for the master branch
 on:
   push:
@@ -23,25 +21,17 @@ env:
   DOTNET_VERSION: 6.0
   DOTNET_LEGACY_VERSION: 4.7.1
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  # This workflow contains a single job called "build"
   build-net:
     name: Build-Net
-    # The type of runner that the job will run on
     runs-on: windows-latest
     strategy:
       matrix:
         dotnet: ['6.0.100']
         cloud_env: ['AZURE', 'GCP', 'AWS']
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
-
       #Setup .NET
-      - uses: actions/checkout@main
       - name: Setup Dotnet
         uses: actions/setup-dotnet@v1
         with:
@@ -64,7 +54,6 @@ jobs:
     strategy:
       matrix:
         cloud_env: ['AZURE', 'GCP', 'AWS']
-
     steps:
     - uses: actions/checkout@v1
     - name: Setup Nuget.exe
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -83,12 +83,11 @@ jobs:
     - name: Build with MSBuild
       run: msbuild snowflake-connector-net.sln -p:Configuration=Release
 
-  test:
+  test-windows:
```
sfc-gh-mhofman, line 86: 👍

.github/workflows/main.yml
```diff
@@ -171,11 +169,10 @@ jobs:
...
 
   test-mac:
-    needs: test-linux
+    needs: [build-net, build-framework]
     name: Tests on MAC
     runs-on: macos-latest
     strategy:
-      max-parallel: 1
       matrix:
         #dotnet: ['2.2.402']
```
sfc-gh-mhofman, line 177: can we remove this commented line?

.github/workflows/main.yml
```diff
@@ -83,12 +83,11 @@ jobs:
...
 
-  test:
+  test-windows:
     needs: [build-net, build-framework]
     name: Tests on Windows
     runs-on: windows-latest
     strategy:
-      max-parallel: 1
       matrix:
         #dotnet: ['2.2.402']
```
sfc-gh-mhofman, line 92: can we remove this commented line?

.github/workflows/main.yml
```diff
@@ -131,11 +130,10 @@ jobs:
 #        run: codecov -f Snowflake.Data.Tests\${{ matrix.dotnet }}_${{ matrix.cloud_env }}_coverage.xml -t ffc6a21d-8176-4849-9047-e3a631dcd35a
```
sfc-gh-mhofman, line 130: here I'm not asking to remove since I belive we will be adding it sooner or later

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -36,20 +38,22 @@ public static void TearDownContext()
...
      */
-    [SetUpFixture]
+    [TestFixture]
     public class SFBaseTestAsync
     {
-        private const string connectionStringWithoutAuthFmt = "scheme={0};host={1};port={2};" +
-            "account={3};role={4};db={5};schema={6};warehouse={7}";
+        private const string ConnectionStringWithoutAuthFmt = "scheme={0};host={1};port={2};" +
+                                                              "account={3};role={4};db={5};schema={6};warehouse={7}";
+        private const string ConnectionStringSnowflakeAuthFmt = ";user={0};password={1};";
```
sfc-gh-mhofman, line 49: Since we are touching this there are multiple ConnectionsStrings in the code base:
- without auth but later on added with user password
- with user and pwd
- with jwt authenticator
- with external browser auth
- with application name
If that's not much of a work why don't we move all of those here and reuse them in the tests.
Some special case for just one test I would leave as is.

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -58,24 +62,25 @@ protected string ConnectionStringWithoutAuth
...
+                                                 testConfig.user,
+                                                 testConfig.password);
 
+        protected TestConfig testConfig { get; }
+    }
+    
+    [SetUpFixture]
+    public class TestEnvironmentSetup
+    {
+        private const string ConnectionStringFmt = "scheme={0};host={1};port={2};" +
```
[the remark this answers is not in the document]
sfc-gh-pbulawa: Prior to the initialisation we do not have the schema thus we cannot use the `SFBaseTestAsync.ConnectionStringWithoutAuth` - the driver fails to create the connection with `ConnectionStringWithoutAuth` as the schema is not found within the SnowflakeDB.

.github/workflows/main.yml
sfc-gh-pbulawa: Done

.github/workflows/main.yml
sfc-gh-pbulawa: Done

.github/workflows/main.yml
```diff
@@ -1,8 +1,6 @@
-# This is a basic workflow to help you get started with Actions
-
 name: DotNet Build and Test
 
-# Controls when the action will run. Triggers the workflow on push or pull request
+# Triggers the workflow on push or pull request
```
sfc-gh-pfus, line 3: Can we join lines?

Snowflake.Data.Tests/SFBaseTest.cs
sfc-gh-pfus: I don't think we should include this in this PR. And I'm not sure if I agree. This are specific auths and requires specific cases. I wouldn't pull everything to the uber base class. Let's keep only common auth data here?

Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -188,6 +188,7 @@ public void TestCrlCheckSwitchConnection()
         [Test]
         public void TestConnectString()
         {
+            var schemaName = "dlSchema_" + Guid.NewGuid().ToString().Replace("-", "_");
```
sfc-gh-pfus, line 191: Why not provide epoch time here instead? It's easier to sort, easier to find the last one/current one.

sfc-gh-pbulawa: Epoch time is not unique and can be skewed when the system is synchronising the time with some server. I know there is a very little chance that collision will happen, but I think using the UUID is a better approach here.

.github/workflows/main.yml
sfc-gh-pbulawa: Done

Snowflake.Data.Tests/SFBaseTest.cs
sfc-gh-pbulawa: I agree with Piotr that we should keep the common data here as it is.

Snowflake.Data.Tests/SFConnectionIT.cs
sfc-gh-pbulawa: To find the newest schemas you can use `describe database <db>;` and sort it by `created_on` column.

sfc-gh-mhofman (approved):

PR #709, opened 2023-07-19
[GithubAction] Issue write permission on the issues for github action

sfc-gh-skumbham:
[GithubAction] Issue write permission on the issues for github action


sfc-gh-igarish (approved):
LGTM

PR #711, opened 2023-07-19
SNOW-871778: Update GitHub actions

sfc-gh-pbulawa:
Update the github actions versions to the newest ones available and keep the version consistency between the different actions

### DIFF
--- a/.github/workflows/snyk-pr.yml
+++ b/.github/workflows/snyk-pr.yml
@@ -5,8 +5,7 @@ on:
       - master
 jobs:
   whitesource:
-    permissions: 
-      issues: write
+    permissions: write-all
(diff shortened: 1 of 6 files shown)


.github/workflows/snyk-pr.yml
```diff
@@ -5,8 +5,7 @@ on:
       - master
 jobs:
   whitesource:
-    permissions: 
-      issues: write
+    permissions: write-all
```
sfc-gh-dprzybysz, line 8: should it be part of this PR?

sfc-gh-pbulawa: You are right, reverted it

PR #713, opened 2023-07-20
add max retry on http request to prevent infinite retry

sfc-gh-ext-simba-hx:
sdk issue 457

add new connection parameter `maxHttpRetries` to limit the number of retry attempts.


sfc-gh-igarish (approved):
LGTM

PR #715, opened 2023-07-21
SNOW-871776: Revert the version of CLA Assistant to @master

sfc-gh-pbulawa:
Rules do not allow to use `@v2` for CLAAssistant action thus reverting to `@master`.

### DIFF
--- a/.github/workflows/cla_bot.yml
+++ b/.github/workflows/cla_bot.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - name: "CLA Assistant"
         if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
-        uses: contributor-assistant/github-action/@master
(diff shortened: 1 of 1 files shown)


.github/workflows/cla_bot.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     steps:
       - name: "CLA Assistant"
         if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
-        uses: contributor-assistant/github-action@v2
+        uses: contributor-assistant/github-action/@master
```
sfc-gh-dprzybysz, line 19: can you add a comment why we should stay with `@master`?

sfc-gh-dprzybysz (changes requested):

.github/workflows/cla_bot.yml
sfc-gh-pbulawa: Done

sfc-gh-dprzybysz (approved):

PR #717, opened 2023-07-24
SNOW-828646: Test connection is not reused when not created successfully

sfc-gh-dprzybysz:
- Add tests to verify that connection is not marked as open (and taken from the pool) when before was unsuccessfully opened because of e.g. invalid user name

### DIFF
--- a/Snowflake.Data.Tests/SFConnectionIT.cs
+++ b/Snowflake.Data.Tests/SFConnectionIT.cs
@@ -23,8 +23,6 @@ namespace Snowflake.Data.Tests
     {
         private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<SFConnectionIT>();
 
-        private const string ConnectionFailureErrorCode = "08006";
-
         [Test]
         [Ignore("ConnectionIT")]
         public void ConnectionITDone()
@@ -97,7 +95,7 @@ namespace Snowflake.Data.Tests
                     {
                         // Expected
                         s_logger.Debug("Failed opening connection ", e);
-                        Assert.AreEqual(ConnectionFailureErrorCode, e.SqlState);
+                        AssertIsConnectionFailure(e);
                     }
 
                     Assert.AreEqual(ConnectionState.Closed, conn.State);
@@ -134,7 +132,7 @@ namespace Snowflake.Data.Tests
                 {
                     // Expected
                     s_logger.Debug("Failed opening connection ", e);
-                    Assert.AreEqual(ConnectionFailureErrorCode, e.SqlState);
+                    AssertIsConnectionFailure(e);
                 }
 
                 Assert.AreEqual(ConnectionState.Closed, conn.State);
@@ -159,7 +157,7 @@ namespace Snowflake.Data.Tests
                 }
                 catch (SnowflakeDbException e)
                 {
-                    Assert.AreEqual(ConnectionFailureErrorCode, e.SqlState);
+                    AssertIsConnectionFailure(e);
                     AssertConnectionIsNotOpen(snowflakeConnection);
                     if (explicitClose)
                     {
@@ -188,7 +186,7 @@ namespace Snowflake.Data.Tests
                 }
                 catch (SnowflakeDbException e)
                 {
-                    Assert.AreEqual(ConnectionFailureErrorCode, e.SqlState);
+                    AssertIsConnectionFailure(e);
                     AssertConnectionIsNotOpen(snowflakeConnection);
                     snowflakeConnection = null;
                 }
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -23,6 +23,8 @@ class SFConnectionIT : SFBaseTest
     {
         private static SFLogger logger = SFLoggerFactory.GetLogger<SFConnectionIT>();
 
+        private static readonly string ConnectionFailureErrorCode = "08006";
```
sfc-gh-pmotacki, line 26: It is possible to use SnowflakeDbException.CONNECTION_FAILURE_SSTATE static variable? Im not sure because it is internal static string.

sfc-gh-dprzybysz: It should work - changed

sfc-gh-pmotacki (approved):
LGTM

Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -132,13 +134,75 @@ public void TestIncorrectUserOrPasswordBasicConnection()
...
+                        snowflakeConnection.Close();
+                        AssertConnectionIsNotOpen(snowflakeConnection);
+                    }
+                    snowflakeConnection = null;
+                }
+            }
+        }
+
+        [Test]
+        public void TestConnectionIsNotMarkedAsOpenWhenWasNotCorrectlyOpenedWithUsing()
```
sfc-gh-knozderko, line 173: what do you think of renaming to ... WithUsingClause or ...WithUsingStatement?

Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -132,13 +134,75 @@ public void TestIncorrectUserOrPasswordBasicConnection()
...
         }
 
+        [Test]
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestConnectionIsNotMarkedAsOpenWhenWasNotCorrectlyOpenedBefore(bool explicitClose)
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnection snowflakeConnection = null;
+            for (var i = 0; i < 2; ++i)
```
sfc-gh-knozderko, line 149: what do you think of using Enumerable.Range(0, 2) instead of a for loop?

Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -132,13 +134,75 @@ public void TestIncorrectUserOrPasswordBasicConnection()
...
 			}
         }
 
+        [Test]
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestConnectionIsNotMarkedAsOpenWhenWasNotCorrectlyOpenedBefore(bool explicitClose)
+        {
+            SnowflakeDbConnectionPool.SetPooling(true);
+            SnowflakeDbConnection snowflakeConnection = null;
```
sfc-gh-knozderko, line 148: this variable is used only inside the loop and its value doesn't preserve to the next iteration so it could be considered to declare inside the loop

Snowflake.Data.Tests/SFConnectionIT.cs
```diff
@@ -132,13 +134,75 @@ public void TestIncorrectUserOrPasswordBasicConnection()
...
+                    snowflakeConnection = null;
+                }
+            }
+        }
+
+        private static void AssertConnectionIsNotOpen(SnowflakeDbConnection snowflakeDbConnection)
+        {
+            Assert.NotNull(snowflakeDbConnection);
+            Assert.IsFalse(snowflakeDbConnection.IsOpen()); // check via public method
+            Assert.AreEqual(snowflakeDbConnection.State, ConnectionState.Closed); // ensure internal state is expected
```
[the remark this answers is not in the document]
sfc-gh-dprzybysz: replaced

Snowflake.Data.Tests/SFConnectionIT.cs
sfc-gh-dprzybysz: Used suffix `WithUsingClause`

Snowflake.Data.Tests/SFConnectionIT.cs
sfc-gh-dprzybysz: moved down to loop in this and next test

Snowflake.Data.Tests/SFConnectionIT.cs
sfc-gh-dprzybysz: I don't see a lot of value in such change - it would add new `using` and replace:
```
for (int i = 0; i < 2; ++i)
```
with 
```
foreach (int i in Enumerable.Range(0, 2))
```
In my opinion is does not increase readability

sfc-gh-knozderko (approved):
LGTM

PR #732, opened 2023-08-01
SNOW-879621: Fix artifact path

sfc-gh-pbulawa:
Fix the artifact path as for linux and mac these artifacts are not yet picked up.

### DIFF
--- a/Snowflake.Data.Tests/SFBaseTest.cs
+++ b/Snowflake.Data.Tests/SFBaseTest.cs
@@ -176,8 +176,10 @@ namespace Snowflake.Data.Tests
             
             var dotnetVersion = Environment.GetEnvironmentVariable("net_version");
             var cloudEnv = Environment.GetEnvironmentVariable("snowflake_cloud_env");
+
+            var separator = Path.DirectorySeparatorChar;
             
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -176,8 +176,10 @@ public void CreateTestTimeArtifact()
             
             var dotnetVersion = Environment.GetEnvironmentVariable("net_version");
             var cloudEnv = Environment.GetEnvironmentVariable("snowflake_cloud_env");
+
+            var separator = Path.DirectorySeparatorChar;
             
-            File.WriteAllText($"../../../{GetOs()}_{dotnetVersion}_{cloudEnv}_performance.csv", resultText);
+            File.WriteAllText($"..{separator}..{separator}..{separator}{GetOs()}_{dotnetVersion}_{cloudEnv}_performance.csv", resultText);
```
sfc-gh-dprzybysz, line 182: maybe let's add a comment why do we need to go three times down the tree?

sfc-gh-pbulawa: Done

sfc-gh-dprzybysz (approved):

sfc-gh-pfus (approved):

PR #737, opened 2023-08-04
SNOW-880335: Use separate instance per test case

sfc-gh-pbulawa:
Use separate instance for each test case to be sure that tests are properly separated from eachother.

### DIFF
--- a/Snowflake.Data.Tests/SFBaseTest.cs
+++ b/Snowflake.Data.Tests/SFBaseTest.cs
@@ -55,7 +55,7 @@ namespace Snowflake.Data.Tests
         private Stopwatch _stopwatch;
         
         [SetUp]
-        public void SetupStopwatch()
+        public void BeforeTest()
         {
             _stopwatch = new Stopwatch();
             _stopwatch.Start();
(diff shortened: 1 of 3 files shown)


sfc-gh-dprzybysz (approved):

Snowflake.Data.Tests/SFBaseTest.cs
```diff
@@ -54,14 +55,14 @@ public class SFBaseTestAsync
         private Stopwatch _stopwatch;
         
         [SetUp]
-        public void BeforeTest()
+        public void SetupStopwatch()
```
sfc-gh-pfus, line 58: Why renaming this method? As I can see `BeforeTest` is the common name used across the project.

sfc-gh-pbulawa: I have renamed this method to make it more specific what is it's purpose. How about `SetupStopwatchBeforeTest`?

sfc-gh-pfus: I'm not sure. For me it's more like a standard and I would like to preserve the name consistent with other methods. See, this method is public, publicly it doesn't matter if internally some stopwatch is set or not. Publicly it is important to run it before each test.

sfc-gh-pbulawa: Renamed it.

sfc-gh-pfus (approved):

sfc-gh-dstempniak (approved):

PR #760, opened 2023-08-28
SNOW-726736 [HTAP] Add GetQueryId function in DbCommand

sfc-gh-ext-simba-hx:
### Description
sdk issue121
Add GetQueryId() function in SnowflakeDbCommand, as well as the existing one in SnowflakeDbDataReader. 

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbCommand.cs
+++ b/Snowflake.Data/Client/SnowflakeDbCommand.cs
@@ -238,6 +238,15 @@ namespace Snowflake.Data.Client
             throw new NotImplementedException();
         }
 
+        public string GetQueryId()
+        {
+            SnowflakeDbConnection conn = (SnowflakeDbConnection)connection;
+            if ((conn != null) && (conn.SfSession != null))
+            {
+                return conn.SfSession.lastQueryId;
+            }
+            return null;
+        }
         protected override DbParameter CreateDbParameter()
         {
             return new SnowflakeDbParameter();
--- a/Snowflake.Data/Client/SnowflakeDbCommand.cs
+++ b/Snowflake.Data/Client/SnowflakeDbCommand.cs
@@ -240,13 +240,13 @@ namespace Snowflake.Data.Client
 
         public string GetQueryId()
         {
-            SnowflakeDbConnection conn = (SnowflakeDbConnection)connection;
-            if ((conn != null) && (conn.SfSession != null))
+            if (sfStatement != null)
             {
-                return conn.SfSession.lastQueryId;
+                return sfStatement.GetQueryId();
             }
             return null;
         }
+
         protected override DbParameter CreateDbParameter()
         {
             return new SnowflakeDbParameter();
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -151,6 +151,7 @@ namespace Snowflake.Data.Client
             logger.Debug("Close Connection.");
             if (IsNonClosedWithSession())
             {
+                SfSession.lastQueryId = null;
                 var transactionRollbackStatus = SnowflakeDbConnectionPool.GetPooling() ? TerminateTransactionForDirtyConnectionReturningToPool() : TransactionRollbackStatus.Undefined;
                 
                 if (CanReuseSession(transactionRollbackStatus) && SnowflakeDbConnectionPool.addSession(SfSession))
@@ -180,6 +181,7 @@ namespace Snowflake.Data.Client
             {
                 if (IsNonClosedWithSession())
                 {
+                    SfSession.lastQueryId = null;
                     var transactionRollbackStatus = SnowflakeDbConnectionPool.GetPooling() ? TerminateTransactionForDirtyConnectionReturningToPool() : TransactionRollbackStatus.Undefined;
 
                     if (CanReuseSession(transactionRollbackStatus) && SnowflakeDbConnectionPool.addSession(SfSession))
@@ -321,17 +323,26 @@ namespace Snowflake.Data.Client
                 cancellationToken);
         }
 
-        public Mutex GetArrayBindingMutex()
+        public string GetQueryId()
+        {
+            if (SfSession != null)
+            {
+                return SfSession.lastQueryId;
+            }
+            return null;
+        }
+
+        internal Mutex GetArrayBindingMutex()
         {
             return _arraybindingMutex;
         }
 
-        public bool IsArrayBindStageCreated()
+        internal bool IsArrayBindStageCreated()
         {
             return _isArrayBindStageCreated;
         }
 
-        public void SetArrayBindStageCreated()
+        internal void SetArrayBindStageCreated()
         {
             _isArrayBindStageCreated = true;
         }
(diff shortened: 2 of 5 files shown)


Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -321,17 +323,26 @@ public override Task OpenAsync(CancellationToken cancellationToken)
                 cancellationToken);
         }
 
-        public Mutex GetArrayBindingMutex()
+        public string GetQueryId()
```
sfc-gh-igarish, line 326: why do we need on connection?

sfc-gh-ext-simba-hx: That's just in case that application might have the DbCommand instance for previous query execution destroyed but need to get the last query id.

sfc-gh-igarish: If other driver don't have at connection, then please remove it. Otherwise it will seen as feature GAP.

sfc-gh-ext-simba-hx: I see, will remove.

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -238,6 +238,15 @@ public override void Prepare()
             throw new NotImplementedException();
         }
 
+        public string GetQueryId()
+        {
+            SnowflakeDbConnection conn = (SnowflakeDbConnection)connection;
+            if ((conn != null) && (conn.SfSession != null))
+            {
+                return conn.SfSession.lastQueryId;
```
sfc-gh-pfus, line 246: Are you sure that this is the best way? What would happen in the following situation (pseudocode):
```
command1 = conn.createCommand
command2 = conn.createCommand
command1.execute
command2.execute
command1.GetQueryId() <---- here will be the query id of the second command?
```

sfc-gh-ext-simba-hx: make sense. Will put queryId at statement level.

sfc-gh-ext-simba-hx: done.

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-ext-simba-hx: done

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -60,6 +60,9 @@ class SFStatement
 
         string _bindStage = null;
 
+        // the query id of the last succeeded query
```
sfc-gh-mhofman, line 63: its also for unsucceeded query? like executed, parsed but failed regarding syntax or inability to ie. update db, right?

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -182,6 +185,7 @@ private SFBaseResultSet BuildResultSet(QueryExecResponse response, CancellationT
         {
             if (response.success)
             {
+                _lastQueryId = response.data.queryId;
```
sfc-gh-mhofman, line 188: what about non successful scenario, do we happen to have queryId as well and can provide even though result set might not have any rows?

sfc-gh-mhofman (approved):

Snowflake.Data/Core/SFStatement.cs
sfc-gh-mhofman: @sfc-gh-ext-simba-hx @sfc-gh-igarish if it makes sense let's provide queryId in the unhappy path as well. @sfc-gh-pfus said it provided by the server in such a case as well.

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -182,6 +185,7 @@ private SFBaseResultSet BuildResultSet(QueryExecResponse response, CancellationT
         {
             if (response.success)
```
sfc-gh-mhofman, line 186: QueryId is also for failure responses and imho it makes sense to provide it in any case if only server has supplied it in its response. 
That's why I would ask for this change @sfc-gh-ext-simba-hx but let's wait for @sfc-gh-igarish to confirm.

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/SFStatement.cs
sfc-gh-mhofman: I checked on JDBC, ODBC, NodeJS and queryId is provided in those when there is a result set returned. So I would do a similar thing here (resultset might contain just error message) and provide queryId not only for reponse.success

sfc-gh-ext-simba-hx: Posted more details through slack. For my understanding JDBC/ODBC throw exception when query fails (by checking "success" field in query response) therefore only set queryId on success case.

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-hx: done

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-hx: done

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-hx: fixed

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -321,17 +321,17 @@ public override Task OpenAsync(CancellationToken cancellationToken)
                 cancellationToken);
         }
 
-        public Mutex GetArrayBindingMutex()
+        internal Mutex GetArrayBindingMutex()
```
sfc-gh-mhofman, line 324: Technically, it's a breaking change @sfc-gh-ext-simba-hx 
I would consider grouping all such noticed things into a separate PR and merge it once we will be having more unavoidable BCR changes. Otherwise let's revert it. If we need it for some reason  we must start a BCR process.
cc @sfc-gh-igarish

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -180,6 +183,10 @@ private void CleanUpCancellationTokenSources()
 
         private SFBaseResultSet BuildResultSet(QueryExecResponse response, CancellationToken cancellationToken)
         {
+            if (response.data.queryId != null)
```
sfc-gh-mhofman, line 186: Response.data should be there but I would still check if it's not null

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-hx: done

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-ext-simba-hx: done

sfc-gh-mhofman (approved):

PR #776, opened 2023-09-13
SNOW-914342: Log the test status

sfc-gh-pbulawa:
### Description
Log the test status to GH Actions.

### Checklist
- [x] Code compiles correctly
- [x] Code is formatted according to [Coding Conventions](../CodingConventions.md)
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing (`dotnet test`)
- [x] Extended the README / documentation, if necessary
- [x] Provide JIRA issue id (if possible) or GitHub issue id in PR name

PR #778, opened 2023-09-21
SNOW-723810: Add coverage for SFFileTransferAgent

sfc-gh-ext-simba-lf:
### Description
Regarding issue 113

The PR adds tests for SFFileTransferAgent to increase code coverage

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
+++ b/Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
@@ -312,19 +312,22 @@ namespace Snowflake.Data.Tests.UnitTests
             // The file name used for creating multiple files
             string mockFileName = "testUploadWithMultipleFiles";
             string extension = "txt";
+            t_mockUploadRootDirectory = TestNameWithWorker + "t_mockUploadRootDirectory";
+
+            Directory.CreateDirectory($"{t_mockUploadRootDirectory}");
 
             // Create source location with wildcard in its filename
             _responseData.src_locations = new List<string>()
             {
                 // Add wildcard in the source location
-                $"{mockFileName}*.{extension}",
+                $"{t_mockUploadRootDirectory}/{mockFileName}*.{extension}",
             };
 
             // Write the mock files
             int numberOfFiles = 3;
             for (int i = 0; i < numberOfFiles; i++)
             {
-                File.WriteAllText($"{mockFileName}{i}.{extension}", FileContent);
+                File.WriteAllText($"{t_mockUploadRootDirectory}/{mockFileName}{i}.{extension}", FileContent);
             }
 
             // Set command to upload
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
```diff
@@ -0,0 +1,551 @@
...
+            // Delete local directory recursively
+            if (Directory.Exists(t_localLocation))
+            {
+                Directory.Delete(t_localLocation, true);
+            }
+        }
+
+        [Test]
+        [Ignore("FileTransferAgentTest")]
+        public void FileTransferAgentTestDone()
```
sfc-gh-dstempniak, line 159: We don't need this since https://github.com/snowflakedb/snowflake-connector-net/pull/776

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -785,6 +794,12 @@ private static IEnumerable<string> ExpandDirectories(string directoryPath)
             {
                 if (ContainsWildcard(part))
                 {
+                    // Directory containing the wildcard is the first one in the path
+                    if (resolvedPaths.Count == 0)
+                    {
+                        resolvedPaths.Add("./");
```
sfc-gh-dstempniak, line 800: If we have directoryPath like A*/B*/... and there is no directory A*, resolvedPaths will be cleared in 811 line. So we will enter here also for second part B* and I think it's not ok.

sfc-gh-dstempniak: Question: Does "./" work on Windows?

Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
```diff
@@ -0,0 +1,551 @@
...
+            File.WriteAllText(_srcLocations[0], FileContent);
+            _sourceFileSize = new FileInfo(_srcLocations[0]).Length;
+
+            // Download setup
+            // Write mock file in the local location to download
+            if (!Directory.Exists(_location))
+            {
+                Directory.CreateDirectory(_location);
+            }
+            File.WriteAllText(_location + t_realSourceFilePath, FileContent);
```
sfc-gh-dstempniak, line 130: As I understand these steps aren't needed for Upload tests. Maybe split somehow Upload and Download setup so that there is no unnecessary writing on the disk.

Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
sfc-gh-ext-simba-lf: Ok, I'll remove it

Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
sfc-gh-ext-simba-lf: Sounds good, will do

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
sfc-gh-ext-simba-lf: > If we have directoryPath like A*/B*/... and there is no directory A*, resolvedPaths will be cleared in 811 line. So we will enter here also for second part B* and I think it's not ok.

I see, I'll make changes and try to make it work with that case

> Question: Does "./" work on Windows?

It does, would it be safer to use something like `Directory.GetCurrentDirectory()` so it can work with all OS?

sfc-gh-ext-simba-lf: I've updated the PR and made a test for the A*/B* case

sfc-gh-dstempniak (approved):
LGTM

PR #782, opened 2023-10-04
SNOW-929914 Update README.md with example how to bind an array as variable

sfc-gh-dszmolka:
Co-authored-by: @sfc-gh-hachouraria

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,14 +657,14 @@ using Newtonsoft.Json;
                     {
 
                         int[] vals = new int[] { 1, 2, 3 };
-                        string array = JsonConvert.SerializeObject(vals); 
+                        string array = JsonConvert.SerializeObject(vals); // alternatively you can do `vals.ToArray()` when passing it to `p1.Value`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -642,6 +642,41 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+
+                        int[] vals = new int[] { 1, 2, 3 };
+                        string array = JsonConvert.SerializeObject(vals); 
+                        string sql = "CALL test_db.public.test(parse_json(?))"; // test SP, returns a single value
+                        // execute this sql with bind variable 'array'
+                        cmd.CommandText = sql;
+
+                        var p1 = cmd.CreateParameter();
+                        p1.ParameterName = "1";
+                        p1.Value = array; // passing the array in the bind variable
```
sfc-gh-mhofman, line 667: Simple array would do as well: https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs#L567
Should we mention about since it doesn't require dependency nor formatting to a JSON string?

PR #783, opened 2023-10-05
SNOW-902611 move control over creation of session to session pool

sfc-gh-mhofman:
### Description
Moving a code responsible for establishing a new session to the session pool class as a prerequisite step to introduction of a new connection pool implementation.
Fixed some flaky tests which were unrelated to the pool but were failing when mocked connection (with some intentional throw exception behavior) was put back to the pool by another test. 

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -28,6 +28,7 @@ public void TestBasicConnection()
         {
             using (IDbConnection conn = new SnowflakeDbConnection())
             {
+                bool pooling = SnowflakeDbConnectionPool.GetPooling();
```
sfc-gh-mhofman, line 31: will do a setup/teardown for that

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -349,15 +333,15 @@ public void SetArrayBindStageCreated()
         /// Create a new SFsession with the connection string settings.
         /// </summary>
```
sfc-gh-knozderko, line 334: This description is no longer valid.

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -349,15 +333,15 @@ public void SetArrayBindStageCreated()
         /// Create a new SFsession with the connection string settings.
         /// </summary>
         /// <exception cref="SnowflakeDbException">If the connection string can't be processed</exception>
-        private void SetSession()
+        private void OnSessionConnecting()
         {
-            SfSession = new SFSession(ConnectionString, Password);
-            _connectionTimeout = (int)SfSession.connectionTimeout.TotalSeconds;
+            // SfSession = new SFSession(ConnectionString, Password);
```
sfc-gh-knozderko, line 338: remove if not important

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -97,8 +101,93 @@ internal SFSession GetSession(string connStr)
...
-            return null;
+            return NewSession(connStr, password);
         }
+        
+        internal Task<SFSession> GetSessionAsync(string connStr, SecureString password, CancellationToken cancellationToken)
+        {
+            s_logger.Debug("SessionPool::GetSession");
+            if (!_pooling)
+                return NewSessionAsync(connStr, password, cancellationToken);
+            lock (s_sessionPoolLock)
```
sfc-gh-knozderko, line 112: GetSession and GetSessionAsync have a very large common part. Please extract it to a separate method.

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -97,8 +101,93 @@ internal SFSession GetSession(string connStr)
...
+        internal Task<SFSession> NewSessionAsync(String connectionString, SecureString password, CancellationToken cancellationToken)
+        {
+            try
+            {
+                var session = new SFSession(connectionString, password);
+                return session
+                    .OpenAsync(cancellationToken)
+                    .ContinueWith(previousTask =>
+                    {
+                        if (previousTask.IsFaulted && previousTask.Exception != null)
```
sfc-gh-knozderko, line 167: isn't this logic the same as in line 181?
If we would remove this ContinueWith code wouldn't the result be the same?
Maybe it is fine, but I'm just asking.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -46,6 +47,7 @@ public void TestBasicConnection()
 
                 conn.Close();
                 Assert.AreEqual(ConnectionState.Closed, conn.State);
+                SnowflakeDbConnectionPool.SetPooling(pooling);
```
sfc-gh-knozderko, line 50: here you clean up the changes in pooling, but in the subsequent tests you don't...

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1753,6 +1755,7 @@ public void TestAsyncLoginTimeout()
         [Test, NonParallelizable]
         public void TestAsyncDefaultLoginTimeout()
         {
+            SnowflakeDbConnectionPool.SetPooling(false);
```
sfc-gh-knozderko, line 1758: here you don't clean up changes to pooling.
Maybe it would be good to add TearDown method to be executed after each of tests to set the default value of pooling.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -18,7 +18,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
     using Snowflake.Data.Tests.Mock;
     using System.Runtime.InteropServices;
 
-    [TestFixture]
+    [TestFixture, NonParallelizable]
     class SFConnectionIT : SFBaseTest
```
sfc-gh-knozderko, line 22: what do you think about applying arrange/act/assert pattern in this file?

sfc-gh-knozderko (changes requested):

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-mhofman: cause not all tests disable pooling, that's why it was so flaky

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-mhofman: ContinueWith is crucial as we need to return session out of async call.
OpenAsync returns future task with just a state. 
So catch handles any exception from constructor of a session.
ContinueWith needs to handle exceptions from OpenAsync as well as provide final opened session as a result.

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-mhofman: thx

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-mhofman: I think once we deal with those flaky tests we can approach this. For now I didn't want this PR to grow and this class is huge. It should be even split into 2 separate tests. And those who change pooling toggle should be separately treated not to affect biggest bunch of tests

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-mhofman: I was fixing just necessary things as we encountered flaky tests

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -99,6 +119,63 @@ internal SFSession GetSession(string connStr)
...
+                        if (previousTask.IsFaulted)
+                            throw new SnowflakeDbException(
+                                SnowflakeDbException.CONNECTION_FAILURE_SSTATE,
+                                SFError.INTERNAL_ERROR,
+                                "Failure while opening session async");
+
+                        return session;
+                    }, TaskContinuationOptions.NotOnCanceled);
+            }
+            catch (Exception e)
```
sfc-gh-dstempniak, line 167: I think we don't need try .. catch here. The only exception catched here can be from "new SFSession", but why to cover it?

sfc-gh-mhofman: It's was to ensure that anything from new will be thrown as SnowflakeDbException but you're right - new Session done in SnowflakeDbConnection was not covered with try catch so to keep same logic I'll remove it

Snowflake.Data.Tests/IntegrationTests/PoolConfigRestorer.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void Reset()
+        {
+            SnowflakeDbConnectionPool.SetMaxPoolSize(_maxPoolSize);
+            SnowflakeDbConnectionPool.SetTimeout(_timeout);
+            SnowflakeDbConnectionPool.SetPooling(_pooling);
+        }
+    }
+
+
+}
```
sfc-gh-knozderko, line 26: add new line at the end

sfc-gh-mhofman: done

Snowflake.Data.Tests/IntegrationTests/PoolConfigRestorer.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public void Reset()
+        {
+            SnowflakeDbConnectionPool.SetMaxPoolSize(_maxPoolSize);
+            SnowflakeDbConnectionPool.SetTimeout(_timeout);
+            SnowflakeDbConnectionPool.SetPooling(_pooling);
+        }
+    }
+
+
```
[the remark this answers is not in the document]
sfc-gh-mhofman: done

Snowflake.Data.Tests/App.config
```diff
@@ -31,7 +31,7 @@ Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
       <appender-ref ref="RollingFileAppender" />
     </root>    
     <root>
-      <level value="WARN" />
+      <level value="ALL" />
```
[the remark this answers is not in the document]
sfc-gh-mhofman: reverted

PR #787, opened 2023-10-07
SNOW-933246: adding missed fix for HTAP parameter removal

sfc-gh-ext-simba-hx:
### Description
sdk issue 693
Some cases for HTAP metadata/parameter removal was missed in September release. Fix them and add test case as well.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
@@ -182,21 +182,26 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 conn.ConnectionString = ConnectionString;
                 conn.Open();
 
-                IDbCommand cmd = conn.CreateCommand();
-                cmd.CommandText = "alter session set DATE_OUTPUT_FORMAT='MM/DD/YYYY'";
-                cmd.ExecuteNonQuery();
+                try
+                {
+                    IDbCommand cmd = conn.CreateCommand();
+                    cmd.CommandText = "alter session set DATE_OUTPUT_FORMAT='MM/DD/YYYY'";
+                    cmd.ExecuteNonQuery();
 
-                cmd.CommandText = $"select TO_DATE('2013-05-17')";
-                IDataReader reader = cmd.ExecuteReader();
+                    cmd.CommandText = $"select TO_DATE('2013-05-17')";
+                    IDataReader reader = cmd.ExecuteReader();
 
-                Assert.IsTrue(reader.Read());
-                Assert.AreEqual("05/17/2013", reader.GetString(0));
+                    Assert.IsTrue(reader.Read());
+                    Assert.AreEqual("05/17/2013", reader.GetString(0));
 
-                reader.Close();
-
-                // set format back to default to avoid impact other test cases
-                cmd.CommandText = "alter session set DATE_OUTPUT_FORMAT='YYYY-MM-DD'";
-                cmd.ExecuteNonQuery();
+                    reader.Close();
+                }
+                finally
+                {
+                    // set format back to default to avoid impact other test cases
+                    cmd.CommandText = "alter session set DATE_OUTPUT_FORMAT='YYYY-MM-DD'";
+                    cmd.ExecuteNonQuery();
+                }
 
                 conn.Close();
             }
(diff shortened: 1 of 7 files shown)


Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
```diff
@@ -174,6 +174,33 @@ public void TestGetDate(string inputTimeStr)
...
+                cmd.CommandText = $"select TO_DATE('2013-05-17')";
+                IDataReader reader = cmd.ExecuteReader();
+
+                Assert.IsTrue(reader.Read());
+                Assert.AreEqual("05/17/2013", reader.GetString(0));
+
+                reader.Close();
+
+                // set format back to default to avoid impact other test cases
+                cmd.CommandText = "alter session set DATE_OUTPUT_FORMAT='YYYY-MM-DD'";
```
sfc-gh-pfus, line 198: I believe we should perform this in `finally` block. Otherwise, if assertion fails, dirty connection returns to the pool.

Snowflake.Data/Core/SFResultSetMetaData.cs
```diff
@@ -37,24 +37,20 @@ class SFResultSetMetaData
...
+        internal SFResultSetMetaData(QueryExecResponseData queryExecResponseData, SFSession session)
         {
             rowTypes = queryExecResponseData.rowType;
             columnCount = rowTypes.Count;
             statementType = findStatementTypeById(queryExecResponseData.statementTypeId);
             columnTypes = InitColumnTypes();
-            
-            foreach (NameValueParameter parameter in queryExecResponseData.parameters)
+
+            if (session.ParameterMap.ContainsKey(SFSessionParameter.DATE_OUTPUT_FORMAT))
```
sfc-gh-pfus, line 47: If we have only these two params involved, maybe we should have tests for both cases?

sfc-gh-pfus, line 47: Session should always have these parameters as they are returned with the first response, am I right?

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
sfc-gh-mhofman: We have other tests like this without this cleanup. Imho to avoid what you suggest we could disable pooling for this test. Still that would require making this test run non parallel as rest of the tests would be affected with pooling disabled.

Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -432,8 +439,16 @@ internal RequestQueryContext GetQueryContextRequest()
 
         internal void UpdateDatabaseAndSchema(string databaseName, string schemaName)
         {
-            this.database = databaseName;
-            this.schema = schemaName;
+            // with HTAP session metadata removal database/schema
+            // might be not returened in query result
+            if (databaseName != null)
```
sfc-gh-mhofman, line 444: I would use String.IsNullOrEmpty even we know empty string it's less likely

Snowflake.Data/Core/SFResultSetMetaData.cs
sfc-gh-ext-simba-hx: Currently TIME_OUTPUT_FORMAT is not used in the driver.
https://github.com/snowflakedb/snowflake-connector-net/blob/v2.1.2/Snowflake.Data/Core/SFBaseResultSet.cs#L55
Not sure whether that's intentional but I think if that needs to be fixed it should be made in a separated PR. In my opinion better not to change it until we get customer issue about that. Time (or maybe timestamp as well) is pretty common data type so likely the existing customer applications already adapted with the existing output format.

sfc-gh-ext-simba-hx: Yes but for some unit test cases the session is made from a mocked login response which doesn't have these parameters.

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
sfc-gh-ext-simba-hx: Done. Put the clean up into finally block, while I think we can keep parallel and pooling enabled with this test case. The parameter is set on session level and cleared before it back to the pool and can be reused by other test case.

Snowflake.Data/Core/Session/SFSession.cs
sfc-gh-ext-simba-hx: Done. That was my initial thoughts but then I thought maybe empty string could be a valid value returned from server in the future? Since it's not currently (there is no way to clear the database/schema set in the session without change it to a new one) I think better to make the code more straight forward.

sfc-gh-mhofman (approved):

PR #798, opened 2023-10-20
SNOW-893834 tests fix for Arrow

sfc-gh-dstempniak:
### Description
SNOW-893834 tests fix for Arrow

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
@@ -186,7 +186,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
         [Test]
         public void TestDateOutputFormat()
         {
-            using (IDbConnection conn = new SnowflakeDbConnection())
+            using (IDbConnection conn = CreateAndOpenConnection())
             {
                 conn.ConnectionString = ConnectionString;
                 conn.Open();
@@ -199,6 +199,8 @@ namespace Snowflake.Data.Tests.IntegrationTests
 
                     cmd.CommandText = $"select TO_DATE('2013-05-17')";
                     IDataReader reader = cmd.ExecuteReader();
+
+                    ValidateResultFormat(reader);
 
                     Assert.IsTrue(reader.Read());
                     Assert.AreEqual("05/17/2013", reader.GetString(0));
@@ -607,7 +609,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
         [Test]
         public void TestGetByte()
         {
-            using (IDbConnection conn = new SnowflakeDbConnection())
+            using (IDbConnection conn = CreateAndOpenConnection())
             {
                 // Arrange
                 conn.ConnectionString = ConnectionString;
@@ -635,6 +637,8 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 // Act
                 using (IDataReader reader = cmd.ExecuteReader())
                 {
+                    ValidateResultFormat(reader);
+
                     int index = 0;
                     while (reader.Read())
                     {
@@ -790,7 +794,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
         [Test]
         public void TestGetChar()
         {
-            using (IDbConnection conn = new SnowflakeDbConnection())
+            using (IDbConnection conn = CreateAndOpenConnection())
             {
                 // Arrange
                 conn.ConnectionString = ConnectionString;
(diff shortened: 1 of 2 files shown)


Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
```diff
@@ -997,6 +1003,8 @@ public void TestGetDataTypeName()
                 // Act
                 using (DbDataReader reader = (DbDataReader)cmd.ExecuteReader())
                 {
+                    ValidateResultFormat(reader);
```
sfc-gh-knozderko, line 1006: this validation should be part of assert block

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
```diff
@@ -811,6 +815,8 @@ public void TestGetChar()
                 // Act
                 using (IDataReader reader = cmd.ExecuteReader())
                 {
+                    ValidateResultFormat(reader);
```
sfc-gh-knozderko, line 818: this validation should be part of assert block

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
```diff
@@ -200,6 +200,8 @@ public void TestDateOutputFormat()
                     cmd.CommandText = $"select TO_DATE('2013-05-17')";
                     IDataReader reader = cmd.ExecuteReader();
 
+                    ValidateResultFormat(reader);
+
```
sfc-gh-knozderko, line 204: you could even remove this empty line since it is one of assertions

Snowflake.Data.Tests/UnitTests/Logger/EasyLoggerManagerTest.cs
```diff
@@ -22,9 +22,7 @@ public class EasyLoggerManagerTest
         private const string ErrorMessage = "Easy logging Error message";
         private const string FatalMessage = "Easy logging Fatal message";
         private static readonly string s_logsDirectory = Path.GetTempPath();
-
-        [ThreadStatic]
-        private static string t_directoryLogPath;
+        private string _directoryLogPath;
```
sfc-gh-knozderko, line 25: what was wrong with the previous form of thread static? It didn't work?

sfc-gh-dstempniak: There is no need to use static field so I removed it. Its because of: https://github.com/snowflakedb/snowflake-connector-net/actions/runs/6585609942/job/17892373465#step:8:4582

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
sfc-gh-dstempniak: What do you think about Assert.IsTrue(reader.Read()) ?
I think it should be in Act not Assert block. But if so, should we leave Assert.IsTrue(...) or just call reader.Read() ?

sfc-gh-knozderko: ValidateResultFormat is a wraper for:
```
Assert.AreEqual(_resultFormat, ((SnowflakeDbDataReader)reader).ResultFormat);
```

which is definitely an assert.

If you want you could split `Assert.IsTrue(reader.Read())` into something like :
```
// act
var read = reader.Read();

// assert
Assert.IsTrue(read);
```

you can split that. However I would not do it, because for me "act" block should be "execute query on the server and get the result" and that's `cmd.ExecuteReader()` and checking the result is rather a part of assert for me.

Snowflake.Data.Tests/UnitTests/Logger/EasyLoggerManagerTest.cs
sfc-gh-knozderko: This means that this test might be flaky. How your change is going to improve the situation? I think that the cause of the error is that we try to remove working directory after the test while logger has not flushed its logs to the file yet - or something like that. I don't see how your change will improve the situation.

sfc-gh-dstempniak: _The process cannot access the file 'C:\Users\runneradmin\AppData\Local\Temp\easy_logging_logs_dkdsyynl.bgd\dotnet\snowflake_dotnet_k0201fjm.log' because it is being used by another process._
The problem might be connected with static field. Since there is no need to use static I changed it to private.

sfc-gh-pbulawa (approved):

sfc-gh-pbulawa (approved):

PR #801, opened 2023-10-25
SNOW-948463 Removed error information for PUT command for GCP

sfc-gh-dstempniak:
### Description
SNOW-948463 Removed error information for PUT command for GCP

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/FileTransfer/StorageClient/SFGCSClient.cs
```diff
@@ -472,19 +471,18 @@ private SFFileMetadata HandleFileHeaderErrForGeneratedUrls(WebException ex, SFFi
         {
             Logger.Error("Failed to get file header for non-presigned url: " + ex.Message);
 
-            // If file doesn't exist, GET request fails
-            fileMetadata.lastError = ex;
```
sfc-gh-knozderko, line 476: are we sure that it never can be an error worth of showing to the client?
How it is if we download a file and GetFileHeader() does not find the file on the server?

sfc-gh-dstempniak: checked and for GET command if there is no file to download reader.Read() returns false, lastError isn't important in this scenario

PR #803, opened 2023-10-26
SNOW-878067: Retry Strategy

sfc-gh-ext-simba-lf:
### Description
Regarding issue 575

The PR adds a new retry strategy for login requests with a new base time and adding jitter. It also adds the id and name to the header for login requests

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -73,8 +74,18 @@ public class HttpClientConfig
     public sealed class HttpUtil
     {
         static internal readonly int MAX_BACKOFF = 16;
+        private static readonly int s_baseBackOffTime = 1;
+        private static readonly int s_baseLoginBackOffTime = 4;
```
sfc-gh-dstempniak, line 78: Why 4? As I see in the doc and in python implementation, first sleep time is 1s?

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -438,20 +450,29 @@ internal RetryHandler(HttpMessageHandler innerHandler, bool disableRetry, bool f
 
                     requestMessage.RequestUri = updater.Update(errorReason);
 
+                    // Add jitter for login requests
+                    if (isLoginRequest)
+                    {
+                        backOffInSec += (int)GetJitter(backOffInSec);
```
sfc-gh-dstempniak, line 456: What about taking random value between (backOffInSec + jitter) and (2^retryCount + jitter) ?

Snowflake.Data.Tests/UnitTests/HttpUtilTest.cs
```diff
@@ -39,6 +40,49 @@ public async Task TestIsRetryableHTTPCode(HttpStatusCode statusCode, bool forceR
...
+        {
+            // given
+            var lowerBound = -(seconds / 2);
+            var upperBound = seconds / 2;
+
+            double jitter;
+            for (var i = 0; i < 10; i++)
+            {
+                // when
+                jitter = HttpUtil.GetJitter(seconds);
```
sfc-gh-dstempniak, line 79: "seconds" dooesn't change in the loop, is it ok?

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-dstempniak: As I understand from doc and python implementation, we should have something like this:

- 1 retry 
    sleep = 1s
- 2 retry
    jitter = 0.5 * 1s * chooseRandom(-1,1) -> (-0.5s, +0.5s)
    sleep = chooseRandom(1s + jitter, 4s + jitter) -> (0.5s, 4.5s)
- 3 retry
    jitter = 0.5 * prev.sleep * chooseRandom(-1,1) -> (-2.25s, 2.25s)
    sleep = chooseRandom(prev.sleep + jitter, 8s + jitter) -> (0.25s, 10.25s)

Or maybe there is different idea for .NET driver.
@sfc-gh-igarish could you confirm?

sfc-gh-dstempniak (changes requested):

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -36,7 +36,7 @@ internal enum SFSessionProperty
         USER,
         [SFSessionPropertyAttr(required = false)]
         WAREHOUSE,
-        [SFSessionPropertyAttr(required = false, defaultValue = "120")]
+        [SFSessionPropertyAttr(required = false, defaultValue = "300")]
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Ok, will do

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: It's based from this [discussion ](https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/575#issuecomment-1760362338) where base time for login requests is 4

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: > What about taking random value between (backOffInSec + jitter) and (2^retryCount + jitter) ?

The random value is the range of the -50% / +50% of the current sleep time which is increased exponentially by 2 after each retry

- 1st retry
```
sleep = 4s + (0.5 * 4s * ChooseRandom(-1, 1))
// after timeout
sleep *= 2;
```
- 2nd retry
```
sleep = prevSleep + (0.5 * prevSleep  * ChooseRandom(-1, 1))
// after timeout
sleep *= 2;
```

This is from the same discussion as [above](https://github.com/snowflakedb/snowflake-connector-net/pull/803#discussion_r1378041316)

Snowflake.Data.Tests/UnitTests/HttpUtilTest.cs
sfc-gh-ext-simba-lf: It's ok since the generated jitter is random, as long as it's within range of -50%/+50% of "seconds" then it's the expected behaviour

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1735,15 +1743,15 @@ public void TestAsyncLoginTimeout()
...
                     stopwatch.Stop();
-                    int detla = 10; //in case server time slower.
+                    int delta = 10; // in case server time slower.
 
-                    // Should timeout after 5sec
-                    Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 5 * 1000 - detla);
-                    // But never more than 1 sec (max backoff) after the default timeout
-                    Assert.LessOrEqual(stopwatch.ElapsedMilliseconds, (6) * 1000);
+                    // Should timeout after the minimum possible timeout with jitter
+                    Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 1 * 1000 - delta);
```
sfc-gh-dstempniak, line 1749: why 1 * 1000 ? should be timeoutSec * 1000

sfc-gh-dstempniak: Could you remove the restriction `if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) `from all tests you modified? It should be tested on all operating systems.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -90,6 +92,44 @@ public void TestCancelExecuteAsync()
...
+                catch (Exception e)
+                {
+                    Assert.IsInstanceOf<TaskCanceledException>(e.InnerException);
+                }
+                stopwatch.Stop();
+
+                // retry 5 times with backoff 1, 2, 4, 8, 16 seconds
+                // but should not delay more than another 16 seconds
+                Assert.Less(stopwatch.ElapsedMilliseconds, 51 * 1000);
+                Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 30 * 1000);
```
sfc-gh-dstempniak, line 130: Are you sure 30s isn't too much?

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -423,36 +444,52 @@ internal RetryHandler(HttpMessageHandler innerHandler, bool disableRetry, bool f
...
+                        throw new OperationCanceledException($"http request failed and max retry {maxRetryCount} reached");
                     }
 
                     // Disposing of the response if not null now that we don't need it anymore
                     response?.Dispose();
 
                     requestMessage.RequestUri = updater.Update(errorReason);
 
-                    logger.Debug($"Sleep {backOffInSec} seconds and then retry the request, retryCount: {retryCount}");
+                    if (retryCount > 1)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: retryCount can be < 1 which is the first retry and the first retry should use 1 second with no jitter So it's not included in the if clause which adds jitter and chooses randomly between previous sleep + jitter and retryCount^2 + jitter

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -391,21 +391,23 @@ public void TestLoginTimeout()
...
 
-                    //Should timeout before the default timeout (120 sec) * 1000
-                    Assert.Less(stopwatch.ElapsedMilliseconds, 120 * 1000);
-                    // Should timeout after the defined connection timeout
-                    Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, timeoutSec * 1000 - detla);
-                    Assert.AreEqual(5, conn.ConnectionTimeout);
+                    // Should timeout before the defined timeout plus 1 (buffer time)
+                    Assert.LessOrEqual(stopwatch.ElapsedMilliseconds, (timeoutSec + 1) * 1000);
+                    // Should timeout after the minimum possible timeout with jitter
+                    Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 1 * 1000 - delta);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: If timeoutSec was used, the test could fail if the jitter for each retry was the lowest possible value (1, 0.5, 0.25, 0.125, 0.0625, 0.03125)

```  
Expected: greater than or equal to 5000
But was:  1059
```

Should the timeout be changed to a minimum of 1?

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-ext-simba-lf: >why 1 * 1000 ? should be timeoutSec * 1000

Same as above where the test could fail if the jitter was the lowest possible value (1, 0.5, 0.25, 0.125, 0.0625, 0.03125)

If the sleep time is made to be at minimum 1 second for each retry then using timeoutSec should work

> Could you remove the restriction if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) from all tests you modified? It should be tested on all operating systems.

Ok

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: That's the number used for the original retry test case: https://github.com/snowflakedb/snowflake-connector-net/pull/803/files#diff-1bb87bf600c092adcc18c664d65701025fe5a30d0d99722aa1fbc3d1e3c9836aL437

This is for non-login requests which doesn't use the new retry strategy and still uses the original back off pattern of 1, 2, 4, 8, and 16

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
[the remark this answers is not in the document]
sfc-gh-dstempniak: maybe set maxHttpRetries=0 to check real timeout?

Snowflake.Data/Core/HttpUtil.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Putting the sleep back before the next timeout is set seems to have removed the need for the if clause. It's now removed

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Yes, that should work

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -341,7 +358,10 @@ internal RetryHandler(HttpMessageHandler innerHandler, bool disableRetry, bool f
                 p.ConnectionLimit = 20;      // Default value is 2, we need more connections for performing multiple parallel queries
 
                 TimeSpan httpTimeout = (TimeSpan)requestMessage.Properties[SFRestRequest.HTTP_REQUEST_TIMEOUT_KEY];
-                TimeSpan restTimeout = (TimeSpan)requestMessage.Properties[SFRestRequest.REST_REQUEST_TIMEOUT_KEY];
+                // Use the lower timeout value between rest timeout and max retry timeout
+                TimeSpan restTimeout = ((TimeSpan)requestMessage.Properties[SFRestRequest.REST_REQUEST_TIMEOUT_KEY]).TotalSeconds < retryTimeout ?
```
sfc-gh-ext-simba-hx, line 362: Check 0 as unlimited I think the existng rest timeout support that as well.

sfc-gh-dstempniak (approved):
LGTM, please remember about `(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))`

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-ext-simba-lf: Set maxHttpRetries to 0 and replaced assertion with timeoutSec * 1000

Also removed the Windows only restriction

sfc-gh-ext-simba-lf:
> LGTM, please remember about `(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))`

The OS check is now removed

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: It checks for 0 in [here](https://github.com/snowflakedb/snowflake-connector-net/pull/803/files#diff-28552eb7d57bcb8b7f43eb5706654e031aaac0cf623630bbdfb8f7c9b61a93cfR483) to determine whether to limit the timeout or not. And if the timeout from `GetRestTimeout` is 0 when sending the [request](https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/RestRequester.cs#L97), then it will also be unlimited.

However, if the retry timeout is shorter than the connection timeout, it won't be used when the request is sent since it's currently only checked in HttpUtil. The PR is now updated so the shorter timeout will be set in the properties instead so it will be used in both the request and the retry

Also added test case to cover retry timeout being shorter than connection timeout

sfc-gh-ext-simba-hx (approved):
LGTM

PR #805, opened 2023-10-30
SNOW-856231 easy logging - not fail on directory serch error

sfc-gh-knozderko:
### Description
Prevent to fail if client config file could not be found because of an unexpected error during the phase of searching in predefined directory. It could happen if for example there is no home folder in the operating system.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
@@ -137,6 +137,7 @@ namespace Snowflake.Data.Tests.UnitTests.Configuration
             
             // assert
             Assert.IsNull(filePath);
+            t_environmentOperations.Verify(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile), Times.Once);
         }
         
         [Test]
@@ -150,6 +151,7 @@ namespace Snowflake.Data.Tests.UnitTests.Configuration
             
             // assert
             Assert.IsNull(filePath);
+            t_environmentOperations.Verify(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile), Times.Once);
         }
 
         private static void MockHomeDirectory()
--- a/Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
@@ -127,13 +127,24 @@ namespace Snowflake.Data.Tests.UnitTests.Configuration
         }
 
         [Test]
-        public void TestThatReturnsNullIfDirectoryBasedSearchFailsWithUnexpectedError()
+        public void TestThatDoesNotFailWhenSearchForOneOfDirectoriesFails()
         {
             // arrange
-            t_environmentOperations
-                .Setup(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile))
-                .Throws(() => new Exception("No home directory"));
+            MockHomeDirectoryFails();
+
+            // act
+            var filePath = t_finder.FindConfigFilePath(null);
             
+            // assert
+            Assert.IsNull(filePath);
+        }
+        
+        [Test]
+        public void TestThatDoesNotFailWhenOneOfDirectoriesNotDefined()
+        {
+            // arrange
+            MockHomeDirectoryReturnsNull();
+
             // act
             var filePath = t_finder.FindConfigFilePath(null);
             
@@ -146,6 +157,20 @@ namespace Snowflake.Data.Tests.UnitTests.Configuration
             t_environmentOperations
                 .Setup(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile))
                 .Returns(HomeDirectory);
+        }
+
+        private static void MockHomeDirectoryFails()
+        {
+            t_environmentOperations
+                .Setup(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile))
+                .Throws(() => new Exception("No home directory"));
+        }
+
+        private static void MockHomeDirectoryReturnsNull()
+        {
+            t_environmentOperations
+                .Setup(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile))
+                .Returns((string) null);
         }
 
         private static void MockFileFromEnvironmentalVariable()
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
```diff
@@ -43,24 +46,34 @@ private string GetFilePathEnvironmentVariable()
...
         
         private string GetFilePathFromInputParameter(string filePath) => string.IsNullOrEmpty(filePath) ? null : filePath;
 
         private string GetHomeDirectory() =>_environmentOperations.GetFolderPath(Environment.SpecialFolder.UserProfile);
 
-        private string GetFilePathFromDriverLocation() => SearchForConfigInDirectory(".");
+        private string GetFilePathFromDriverLocation() => SearchForConfigInDirectory(() => ".", "driver");
 
-        private string SearchForConfigInDirectory(string directory)
+        private string SearchForConfigInDirectory(Func<string> directoryProvider, string directoryDescription)
```
sfc-gh-pfus, line 59: Do we really need this `directoryDescription`? Maybe we can move `var directory = ...` outside of `try` and just use it in `catch`? It would be more even meaningful.

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
```diff
@@ -126,6 +126,21 @@ public void TestThatReturnsNullIfNoWayOfGettingTheFile()
             Assert.IsNull(filePath);
         }
 
+        [Test]
+        public void TestThatReturnsNullIfDirectoryBasedSearchFailsWithUnexpectedError()
```
sfc-gh-pfus, line 130: Is it a correct level of testing? I would expect more like integration test.

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
sfc-gh-knozderko: and what if the operation of getting the directory fails?
This way I did it I handle all sort of errors from different reasons with little code changes so I like it better.

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
sfc-gh-knozderko: I think there is no need for that since all I changed is covered only in this class.

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
```diff
@@ -126,6 +126,21 @@ public void TestThatReturnsNullIfNoWayOfGettingTheFile()
...
+        {
+            // arrange
+            t_environmentOperations
```
sfc-gh-mhofman, line 133: For other mocks there are methods providing a descriptive name for the mock. There is one method used once so having than, maybe this mock here, even used once could have  such a wrapper method?

sfc-gh-mhofman (changes requested):

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
```diff
@@ -126,6 +126,21 @@ public void TestThatReturnsNullIfNoWayOfGettingTheFile()
...
+                .Setup(e => e.GetFolderPath(Environment.SpecialFolder.UserProfile))
+                .Throws(() => new Exception("No home directory"));
+            
+            // act
+            var filePath = t_finder.FindConfigFilePath(null);
```
sfc-gh-dstempniak, line 138: Are we sure the exception has been thrown?
Maybe add `t_environmentOperations.VerifyNoOtherCalls()`

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
sfc-gh-knozderko: done

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
```diff
@@ -43,24 +46,34 @@ private string GetFilePathEnvironmentVariable()
...
+            try
+            {
+                var directory = directoryProvider.Invoke();
+                if (string.IsNullOrEmpty(directory))
+                {
+                    return null;
+                }
+
+                var filePath = Path.Combine(directory, ClientConfigFileName);
+                return OnlyIfFileExists(filePath);
```
sfc-gh-dstempniak, line 70: What if (3 use cases):
1. file exists but there is no read access?
2. configuration points to folder, which doesn't exist? (eg. cannot create directory in EasyLoggingStarter.Init -> GetLogPath) ?
3. folder exists, but there is no write access for the log file?

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
```diff
@@ -43,24 +46,34 @@ private string GetFilePathEnvironmentVariable()
...
-        private string SearchForConfigInDirectory(string directory)
+        private string SearchForConfigInDirectory(Func<string> directoryProvider, string directoryDescription)
         {
-            if (string.IsNullOrEmpty(directory))
+            try
+            {
+                var directory = directoryProvider.Invoke();
+                if (string.IsNullOrEmpty(directory))
+                {
+                    return null;
```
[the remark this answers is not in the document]
sfc-gh-knozderko: done

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
sfc-gh-knozderko: we don't care about the cases you write about - it will be an error.
If someone makes dummy configuration it is his fault.
I would like just to prevent the driver to fail on on the stage of choosing the configuration file. And only on this stage.

Snowflake.Data.Tests/UnitTests/Configuration/EasyLoggingConfigFinderTest.cs
sfc-gh-knozderko: done

sfc-gh-mhofman (approved):

PR #807, opened 2023-11-08
SNOW-957736: Fix SnowflakeDbConnection.Dispose behaviour

sfc-gh-pbulawa:
### Description
The connection which is not closed properly by the user should not enter the connection pool.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbConnection.cs
+++ b/Snowflake.Data/Client/SnowflakeDbConnection.cs
@@ -26,7 +26,7 @@ namespace Snowflake.Data.Client
         
         internal int _connectionTimeout;
 
-        private bool disposed = false;
+        private bool _disposed = false;
 
         private static Mutex _arraybindingMutex = new Mutex();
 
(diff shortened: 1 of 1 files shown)


Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -361,21 +361,30 @@ protected override DbCommand CreateDbCommand()
...
+                    }
+                    catch (Exception ex)
+                    {
+                        // Prevent an exception from being thrown when disposing of this object
+                        logger.Error("Unable to close connection", ex);
+                    }
+                }
+                else
+                {
+                    SfSession?.close();
```
sfc-gh-dstempniak, line 380: Don't we want exception handling here?

sfc-gh-pbulawa: I don't think this is necessary as the session Close method should be safe in the usage.

sfc-gh-dstempniak (approved):

PR #818, opened 2023-11-21
Pool/SNOW-937189 busy sessions in pool (3/4)

sfc-gh-knozderko:
### Description
Connection Pool v2:
- introduction of a counter of busy sessions created by Pool Manager
- pool size counts both: Idle and Busy sessions

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/Session/ICounter.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    internal interface ICounter
+    {
+        int Count();
+
+        void Increase();
+
+        void Decrease();
+    }
+}
```
sfc-gh-mhofman, line 11: new line

Snowflake.Data.Tests/UnitTests/Session/NonNegativeCounterTest.cs
```diff
@@ -0,0 +1,69 @@
...
+            // act
+            counter.Decrease();
+            
+            // assert
+            Assert.AreEqual(1, counter.Count());
+            
+            // act
+            counter.Decrease();
+            
+            // assert
```
sfc-gh-mhofman, line 59: If I were pedantic 2nd part could have been a new TestDecreaseDoesNotGoBelowZero

Snowflake.Data/Core/Session/ConnectionPoolManager.cs
```diff
@@ -102,6 +102,9 @@ public int GetCurrentPoolSize()
 
         public bool SetPooling(bool poolingEnabled)
         {
+            if (!poolingEnabled)
+                throw new Exception(
```
sfc-gh-mhofman, line 106: why? if an app has several pools and wants to switch them off in one shot - why not allow this?

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
```diff
@@ -160,30 +161,16 @@ public void TestSetTimeoutForAllPools()
         }         
         
         [Test]
-        public void TestSetPoolingDisabledForAllPools()
+        public void TestSetPoolingDisabledForAllPoolsNotPossible()
```
sfc-gh-mhofman, line 164: I've commented on possibility of this in the production code

Snowflake.Data.Tests/IntegrationTests/ConnectionSinglePoolCacheIT.cs
```diff
@@ -71,7 +92,54 @@ public void TestConnectionPoolFull()
             Assert.AreEqual(ConnectionState.Closed, conn4.State);
             SnowflakeDbConnectionPool.ClearAllPools();
         }
+
+        [Test]
+        public void TestConnectionPoolDisableFromHighLevel()
```
sfc-gh-mhofman, line 97: more precisely it would be from PoolManagerLevel but I don't push to change, WDYT?

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -36,14 +37,22 @@ private SessionPool()
...
             }
         }
 
-        private SessionPool(string connectionString, SecureString password) : this()
+        private SessionPool(string connectionString, SecureString password)
         {
-            ConnectionString = connectionString;
-            Password = password;
-            _allowExceedMaxPoolSize = false; // TODO: SNOW-937190
+            lock (s_sessionPoolLock)
```
sfc-gh-mhofman, line 46: Such a lock is at the pool manager layer. Here we assume problems which are not entirely for this class. I would remove. Moreover, just one constructor has this lock. Seems weird.

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -143,6 +153,8 @@ private SFSession NewSession(String connectionString, SecureString password)
             try
             {
                 var session = s_sessionFactory.NewSession(connectionString, password);
+                if (_pooling)
+                    _busySessionsCounter.Increase();
```
sfc-gh-mhofman, line 157: I would increase the counter after we open a session (for the sake of any error occurred)

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -163,6 +175,8 @@ private Task<SFSession> NewSessionAsync(String connectionString, SecureString pa
         {
             s_logger.Debug("SessionPool::NewSessionAsync");
             var session = s_sessionFactory.NewSession(connectionString, password);
+            if (_pooling)
+                _busySessionsCounter.Increase();
```
sfc-gh-mhofman, line 179: lets here move this counter increase to the ContinueWith phase, makes sense?

sfc-gh-mhofman (changes requested):

Snowflake.Data.Tests/UnitTests/Session/NonNegativeCounterTest.cs
```diff
@@ -0,0 +1,69 @@
...
+            Assert.AreEqual(0, counter.Count());
+            
+            // act
+            counter.Decrease();
+
+            // assert
+            Assert.AreEqual(0, counter.Count());
+        }
+    }
+}
```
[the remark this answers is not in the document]
sfc-gh-knozderko: done

Snowflake.Data/Core/Session/ICounter.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: done

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -129,6 +138,7 @@ private SFSession GetIdleSession(string connStr)
                         else
                         {
                             s_logger.Debug($"reuse pooled session with sid {session.sessionId}");
+                            _busySessionsCounter.Increase();
```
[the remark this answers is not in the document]
sfc-gh-knozderko: it depends on _pooling variable. GetIdleSession() is a private method. It is only used in GetSession() and GetSessionAsync(). It is called only if _pooling is true.

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -187,15 +201,19 @@ internal bool AddSession(SFSession session)
...
             lock (s_sessionPoolLock)
             {
-                if (_idleSessions.Count >= _maxPoolSize)
-                {
-                    CleanExpiredSessions();
-                }
-                if (_idleSessions.Count >= _maxPoolSize)
+                _busySessionsCounter.Decrease();
+                CleanExpiredSessions();
+                if (GetCurrentPoolSize() >= _maxPoolSize)
```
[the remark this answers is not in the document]
sfc-gh-knozderko: for me this method could be renamed to ClearIdleSessions() to avoid confusion.
So the question is if session is opened and is executing a query and we clear its pool, what should happen to the busy session.
For now I would do nothing about it. If in future we make the session aware of if it comes from a pool or not then we can think about a different behaviour.

Snowflake.Data.Tests/UnitTests/Session/NonNegativeCounterTest.cs
sfc-gh-knozderko: done

Snowflake.Data/Core/Session/ConnectionPoolManager.cs
sfc-gh-knozderko: due to our conversation. You wanted in the next steps to introduce parameter in the connection string to control if the pool is to use. It would be the way to control the pool usage by the user. We can discuss it.

Snowflake.Data.Tests/IntegrationTests/ConnectionSinglePoolCacheIT.cs
sfc-gh-knozderko: done

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: I believe this lock is to avoid creating at the same time pool for the same connection string in two separate threads.
You are right that ConnectionPoolManager also creates a lock so it could be removed here. I'm removing.
ConnectionCacheManager does not have such a lock so in SessionPool constructor dedicated to it the lock is required.

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: sessionPool offers in their API:
- GetSession() - here we increment busySessions count
- AddSession() - which is rather "return session to the pool as an idle session" which is decreasing busySessions count
There is no place for open() in this api. Remember that for both: idle and busy sessions, the session is already opened.

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: done

Snowflake.Data/Core/Session/FixedZeroCounter.cs
```diff
@@ -0,0 +1,15 @@
+namespace Snowflake.Data.Core.Session
+{
+    internal class FixedZeroCounter: ICounter
```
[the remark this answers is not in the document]
sfc-gh-knozderko: we discussed to leave it as it is

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: tl; dr; removed

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
sfc-gh-knozderko: done

sfc-gh-mhofman (approved):

Issue #817, opened 2023-11-20
SNOW-974746: Cannot use bind parameters with batch/multi statements


paulverbeke:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of .NET driver are you using?
2.1.2.0
   
2. What operating system and processor architecture are you using?
W10 64bits
   
3. What version of .NET framework are you using?
   .net framework 4.8
   
4. What did you do?
I constructed a query of 2 statements INSERT INTO.
`INSERT INTO MY_TABLE VALUES (:1,:2,:3,:4,:5);INSERT INTO MY_TABLE VALUES (:6,:7,:8,:9,:10);`
They both have bind parameters that are (all 10 of them, I checked) correctly added to the ParametersCollection.
Then I added the last parameter "MULTI_STATEMENT_COUNT" with a value of 0 for enabling batch statements.

5. What did you expect to see?

   10 new rows should be inserted correctly but instead at execution I got an error:
```Error: JavaScript execution error: Uncaught Execution of multiple statements failed on statement "INSERT INTO IMPORT_PRC_DATA_PN..." (at line 1, position 0).
SQL compilation error: error line 1 at position 45
Bind variable :1 not set. in SYSTEM$MULTISTMT at '    throw `Execution of multiple statements failed on statement {0} (at line {1}, position {2}).`.replace('{1}', LINES[i])' position 4
stackstrace: 
SYSTEM$MULTISTMT line: 10 SqlState: P0000, VendorCode: 100132, QueryId: 01b0751e-0103-1ca4-0000-9ba90328a4be
```

I can send batch statements INSERT INTO without any problems - OK
I can send one INSERT INTO statement with bind parameters without any problems - OK
But as soon as there is severals statements in a batch AND bind parameters I receive an error - KO

I finally found this in the documentation of Snowflake, small note. "Snowflake does not currently support variable binding in multi-statement SQL requests."
https://docs.snowflake.com/en/developer-guide/sql-api/submitting-multiple-statements

Is this true ? Then why is it not documented in the readme of this connector ?

sfc-gh-dszmolka:
hi and thank you for submitting this issue! We observed the exactly same behaviour in other Snowflake drivers as well, which are absolutely independent from each other and even use different programming languages. See https://github.com/snowflakedb/snowflake-connector-nodejs/issues/704 for reference

[...]

PR #819, opened 2023-11-22
SNOW-974746 document that we currently do not support bind variables in multi-statement queries

sfc-gh-dszmolka:
re: SNOW-974746 / #817 
document that we currently do not support bind variables in multi-statement queries, we're getting issues related to this

document-only change, based on https://docs.snowflake.com/en/developer-guide/sql-api/submitting-multiple-statements#specifying-multiple-sql-statements-in-the-request which sems to be valid for all the drivers

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,6 +488,8 @@ With version 2.0.18 and later of the .NET connector, you can send
 a batch of SQL statements, separated by semicolons,
 to be executed in a single request.
 
+**Note**: Snowflake does not currently support variable binding in multi-statement SQL requests.
+
 ---
 **Note**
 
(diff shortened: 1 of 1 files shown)


sfc-gh-ext-simba-hx:
Hi @sfc-gh-dszmolka
https://github.com/snowflakedb/snowflake-connector-net/issues/817 I think this one is for parameter binding, which for my understanding is different from variable binding https://docs.snowflake.com/en/developer-guide/snowflake-scripting/variables
We have test cases cover multiple statements with parameter binding, both select and insert query, so the issue more sounds like customer usage issue to me.
https://github.com/snowflakedb/snowflake-connector-net/blob/v2.1.3/Snowflake.Data.Tests/IntegrationTests/SFMultiStatementsIT.cs#L129

Thanks,
Harry

PR #825, opened 2023-11-30
SNOW-950923 fix to provide QueryID for failures during GET/PUT file operations

sfc-gh-mhofman:
### Description
QueryID is super helpful when solving issues but it was not reachable when PUT/GET operation got failed at the command/exception level.
It was not available for successful scenario for PUT/GET commands as well.
This change actually introduces a new type of exception prior to the one used previously - FileNotFoundException/DirectoryNotFoundException and this previous exception is now an inner exception of SnowflakeDbException which we had to use to provide QueryId.
In rare cases (for instance when a connection was not opened) request to PUT/GET file on stage might still end up with a base Exception since QueryId won't be created in such a case.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -701,7 +701,7 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 
     }
 ```
-In case of a failure a SnowflakeDbException will be thrown and will provide QueryId.
+In case of a failure a SnowflakeDbException will be thrown with the affected QueryId.
 Inner exception (if applicable) will provide some details on the failure cause and
 it will be for example: FileNotFoundException, DirectoryNotFoundException.
 
--- a/README.md
+++ b/README.md
@@ -685,11 +685,11 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 {
     try
     {
-	    conn.ConnectionString = "<connecting parameter>";
+	    conn.ConnectionString = "<connection parameters>";
 	    conn.Open();
 	    var cmd = (SnowflakeDbCommand)conn.CreateCommand(); // cast allows get QueryId from the command
 	
-	    cmd.CommandText = $"PUT file://some_data.csv @my_schema.my_stage AUTO_COMPRESS=TRUE";
+	    cmd.CommandText = "PUT file://some_data.csv @my_schema.my_stage AUTO_COMPRESS=TRUE";
 	    var reader = cmd.ExecuteReader();
 	    Assert.IsTrue(reader.read()); // on success
         Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
@@ -698,10 +698,15 @@ using (IDbConnection conn = new SnowflakeDbConnection())
     {
         Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // when failed
         Assert.That(e.InnerException.GetType(), Is.EqualTo(typeof(FileNotFoundException)));
-
+    }
+    catch (Exception e)
+    {
+        // for instance on a query execution if a connection was not opened
     }
 ```
-In case of a failure a SnowflakeDbException will be thrown with the affected QueryId.
+In case of a failure an exception will be thrown. 
+If it was after the query got executed this exception will be a SnowflakeDbException containing affected QueryId.
+In case of the initial phase of execution QueryId might not be provided.
 Inner exception (if applicable) will provide some details on the failure cause and
 it will be for example: FileNotFoundException, DirectoryNotFoundException.
 
@@ -715,11 +720,11 @@ To use the command in a driver similar code can be executed in a client app:
 ```cs
     try
     {
-	    conn.ConnectionString = "<connecting parameter>";
+	    conn.ConnectionString = "<connection parameters>";
 	    conn.Open();
 	    var cmd = (SnowflakeDbCommand)conn.CreateCommand(); // cast allows get QueryId from the command
 	
-	    cmd.CommandText = $"GET @my_schema.my_stage/stage_file.csv file://local_file.csv AUTO_COMPRESS=TRUE";
+	    cmd.CommandText = "GET @my_schema.my_stage/stage_file.csv file://local_file.csv AUTO_COMPRESS=TRUE";
 	    var reader = cmd.ExecuteReader();
 	    Assert.IsTrue(reader.read()); // if succeeded
         Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
@@ -728,9 +733,15 @@ To use the command in a driver similar code can be executed in a client app:
     {
         Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // on failure
     }
+    catch (Exception e)
+    {
+        // some other failure has occurred before query got started
+    }
 ```
 In case of a failure a SnowflakeDbException will be thrown or DBDataReader will return a False response.
 In any case QueryId should be available from the exception or command property.
+Similarly to PUT command in some rare cases QueryID might not be possible to provide and a generic Exception
+can be thrown.
 
 Close the Connection
 --------------------
--- a/README.md
+++ b/README.md
@@ -672,6 +672,66 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+PUT local files to stage
+------------------------
+
+PUT command can be used to upload files of a local directory or a single local file to the Snowflake stages (named, internal table stage or internal user stage).
+Such staging files can be used to load data into a table. 
+More on this topic: [File staging with PUT](https://docs.snowflake.com/en/sql-reference/sql/put).
+
+In the driver the command can be executed in a bellow way:
+```cs
+using (IDbConnection conn = new SnowflakeDbConnection())
+{
+    try
+    {
+	    conn.ConnectionString = "<connecting parameter>";
+	    conn.Open();
+	    var cmd = (SnowflakeDbCommand)conn.CreateCommand(); // cast allows get QueryId from the command
+	
+	    cmd.CommandText = $"PUT file://some_data.csv @my_schema.my_stage AUTO_COMPRESS=TRUE";
+	    var reader = cmd.ExecuteReader();
+	    Assert.IsTrue(reader.read()); // on success
+        Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
+    }
+    catch (SnowflakeDbException e)
+    {
+        Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // when failed
+        Assert.That(e.InnerException.GetType(), Is.EqualTo(typeof(FileNotFoundException)));
+
+    }
+```
+In case of a failure a SnowflakeDbException will be thrown and will provide QueryId.
+Inner exception (if applicable) will provide some details on the failure cause and
+it will be for example: FileNotFoundException, DirectoryNotFoundException.
+
+GET stage files
+---------------
+GET command allows to download stage directories or files to a local directory.
+It can be used in connection with named stage, table internal stage or user stage. 
+Detailed information on the command: [Downloading files with GET](https://docs.snowflake.com/en/sql-reference/sql/get).
+
+To use the command in a driver similar code can be executed in a client app:
+```cs
+    try
+    {
+	    conn.ConnectionString = "<connecting parameter>";
+	    conn.Open();
+	    var cmd = (SnowflakeDbCommand)conn.CreateCommand(); // cast allows get QueryId from the command
+	
+	    cmd.CommandText = $"GET @my_schema.my_stage/stage_file.csv file://local_file.csv AUTO_COMPRESS=TRUE";
+	    var reader = cmd.ExecuteReader();
+	    Assert.IsTrue(reader.read()); // if succeeded
+        Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
+    }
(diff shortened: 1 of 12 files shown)


Snowflake.Data/Core/SFStatement.cs
```diff
@@ -356,12 +356,18 @@ internal SFBaseResultSet Execute(int timeout, string sql, Dictionary<string, Bin
                              bindings,
                              describeOnly);
 
+                    if (logger.IsDebugEnabled())
```
sfc-gh-knozderko, line 359: is it a kind of optimisation? Debug operation has this check in itself, so I would rather not duplicate it here. Especially that then the debug log could not be covered by the tests so it may decrease the code coverage.

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -185,46 +185,60 @@ class SFFileTransferAgent
...
+                }
+                else if (CommandTypes.DOWNLOAD == CommandType)
+                {
+                    download();
                 }
             }
-
-            // Check command type
-            if (CommandTypes.UPLOAD == CommandType)
+            catch (FileNotFoundException e)
```
sfc-gh-knozderko, line 231: why do we handle only FileNotFound and IOException? how did we select these 2? what about others?

Snowflake.Data/Core/SFStatement.cs
sfc-gh-mhofman: I was thinking about not loosing a sec if debug is not enabled but the string being built is not that huge, I'll change it

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -185,46 +185,63 @@ class SFFileTransferAgent
...
+                Logger.Error("Error while transferring file(s): " + e.Message);
+                if (e is SnowflakeDbException snowflakeException)
+                {
+                    if (snowflakeException.QueryId == null)
+                    {
+                        snowflakeException.QueryId = TransferMetadata.queryId;
+                    }
+                    throw snowflakeException;
+                }
+                throw new SnowflakeDbException(SFError.IO_ERROR_ON_GETPUT_COMMAND, TransferMetadata.queryId, e);
```
sfc-gh-dstempniak, line 242: Do we really need to override exceptions? The problem was with GetQueryId, not exception.
From user perspective isn't it better to see FileNotFoundException?

sfc-gh-mhofman: Yes, but in JDBC @sfc-gh-dprzybysz implemented fixed the same problem in a way that both - exception and statement provide query id. I had change addressing exactly what you point out here (https://github.com/snowflakedb/snowflake-connector-net/pull/828/files) but I think drivers should be consistent.

Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
```diff
@@ -474,29 +517,41 @@ private static bool IsCompressedByTheDriver()
...
+                    {
+                        Assert.AreEqual(SFFileCompressionTypes.NONE.Name,
+                            reader.GetString((int)SFResultSet.PutGetResponseRowTypeInfo.SourceCompressionType));
+                        Assert.AreEqual(SFFileCompressionTypes.NONE.Name,
+                            reader.GetString((int)SFResultSet.PutGetResponseRowTypeInfo.DestinationCompressionType));
+                    }
+                    Assert.IsNull(reader.GetString((int)SFResultSet.PutGetResponseRowTypeInfo.ErrorDetails));
                 }
-                else
+                catch (SnowflakeDbException e)
```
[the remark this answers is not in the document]
sfc-gh-mhofman: We need this catch to verify if in case of any failure the command regardless of the outcome contains query id but I will move this catch up and not consume but rethrow

Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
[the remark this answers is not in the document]
sfc-gh-mhofman: base methods are static and those overrides still need those base run. Hiding was not intended, I'll see what we can do about it.

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
sfc-gh-mhofman: Actually we also need it to transport queryId to the SFStatement in case of a failure. Since this is transfer agent starting entire job with a query, to provide it on a statement layer we either have to return result.But in case of an exception there's the easiest way to achieve itwith  without wrapping the original exception and providing along with started and failed queryId. WDYT?

Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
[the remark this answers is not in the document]
sfc-gh-mhofman: Also when PUT failed it either didn't throw or if it did some or all asserts wouldn't take place but I'll reorganize it so it is less prone to hiding issues.

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -185,46 +185,63 @@
...
             }
+
```
sfc-gh-pbulawa, line 244: Empty line

sfc-gh-pbulawa (approved):

Snowflake.Data/Client/SnowflakeDbException.cs
```diff
@@ -23,11 +23,11 @@ public sealed class SnowflakeDbException : DbException
         static private ResourceManager rm = new ResourceManager("Snowflake.Data.Core.ErrorMessages",
             typeof(SnowflakeDbException).Assembly);
 
-        public string SqlState { get; private set; }
+        public new string SqlState { get; private set; }
```
sfc-gh-dstempniak, line 26: _The member 'SnowflakeDbException.SqlState' does not hide an accessible member. The new keyword is not required._
😁

Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
```diff
@@ -208,7 +208,50 @@ public void TestPutFileRelativePathWithoutDirectory()
...
+            // Arrange
+            // Set the PUT query variables but do not create a file
+            t_inputFilePath = "unexisting_file.csv";
+            t_internalStagePath = $"@{t_schemaName}.{t_stageName}";
+
+            // Act
+            using (var conn = new SnowflakeDbConnection(ConnectionString))
+            {
+                conn.Open();
+                var snowflakeDbException = Assert.Throws<SnowflakeDbException>(()=>PutFile(conn));
```
sfc-gh-dstempniak, line 224: Can we add check if this is IO_ERROR_ON_GETPUT_COMMAND ?

Snowflake.Data.Tests/UnitTests/SFFileTransferAgentTests.cs
```diff
@@ -487,15 +490,17 @@ public void TestUploadThrowsArgumentExceptionForMissingRootDirectoryWithWildcard
...
                 _session,
                 _responseData,
                 _cancellationToken);
 
             // Act
-            Exception ex = Assert.Throws<ArgumentException>(() => _fileTransferAgent.execute());
+            SnowflakeDbException ex = Assert.Throws<SnowflakeDbException>(() => _fileTransferAgent.execute());
 
             // Assert
+            Assert.AreEqual(_responseData.queryId, ex.QueryId);
```
sfc-gh-dstempniak, line 503: Maybe add IO_ERROR_ON_GETPUT_COMMAND check (like in TestDownloadThrowsErrorFileNotFound)?
And maybe rename the test.

Snowflake.Data/Client/SnowflakeDbException.cs
sfc-gh-mhofman: Fixed

Snowflake.Data/Core/FileTransfer/SFFileTransferAgent.cs
```diff
@@ -185,46 +185,63 @@ class SFFileTransferAgent
...
-            else if (CommandTypes.DOWNLOAD == CommandType)
+            catch (Exception e)
             {
-                download();
+                Logger.Error("Error while transferring file(s): " + e.Message);
+                if (e is SnowflakeDbException snowflakeException)
+                {
+                    if (snowflakeException.QueryId == null)
+                    {
+                        snowflakeException.QueryId = TransferMetadata.queryId;
```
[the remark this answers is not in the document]
sfc-gh-mhofman: It is. I provided the test when not a FileNotFoundException but SnowflakeDbException is thrown and in such case it not necessarily will contain QueryId from TransferMetadata.

Snowflake.Data/Client/SnowflakeDbException.cs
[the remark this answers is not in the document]
sfc-gh-mhofman: Cannot override base class property as it is nullable and to have it we would need to change language version of entire project. I will defer for now as agreed.

README.md
```diff
@@ -672,6 +672,66 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+        Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
+    }
+    catch (SnowflakeDbException e)
+    {
+        Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // when failed
+        Assert.That(e.InnerException.GetType(), Is.EqualTo(typeof(FileNotFoundException)));
+
+    }
+```
+In case of a failure a SnowflakeDbException will be thrown and will provide QueryId.
```
sfc-gh-pbulawa, line 704: Maybe "In case of a failure a SnowflakeDbException will be thrown with the affected QueryId."

README.md
```diff
@@ -672,6 +672,66 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+	
+	    cmd.CommandText = $"PUT file://some_data.csv @my_schema.my_stage AUTO_COMPRESS=TRUE";
+	    var reader = cmd.ExecuteReader();
+	    Assert.IsTrue(reader.read()); // on success
+        Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
+    }
+    catch (SnowflakeDbException e)
+    {
+        Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // when failed
+        Assert.That(e.InnerException.GetType(), Is.EqualTo(typeof(FileNotFoundException)));
```
sfc-gh-knozderko, line 700: I'm thinking if all this asserts should be in documentation... Someone could be confused and get wrong conclusions like that always FileNotFound is the inner exception. Maybe we need to add a comment clarifying that it is just an example? Is it possible to have SnowflakeDbException without QueryId, for example if some prevalidation fails?

README.md
```diff
@@ -672,6 +672,66 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+	    var reader = cmd.ExecuteReader();
+	    Assert.IsTrue(reader.read()); // if succeeded
+        Assert.DoesNotThrow(() => Guid.Parse(cmd.GetQueryId()));
+    }
+    catch (SnowflakeDbException e)
+    {
+        Assert.DoesNotThrow(() => Guid.Parse(e.QueryId)); // on failure
+    }
+```
+In case of a failure a SnowflakeDbException will be thrown or DBDataReader will return a False response.
```
sfc-gh-knozderko, line 732: I belive that in case of exception there will be no reader at all.

Snowflake.Data.Tests/Util/SnowflakeDbExceptionAssert.cs
```diff
@@ -0,0 +1,87 @@
...
+        private static List<Exception> CollectExceptions(Exception exception)
+        {
+            var collected = new List<Exception>();
+            if (exception is null)
+                return collected;
+            switch (exception)
+            {
+                case AggregateException aggregate:
+                    var inner = aggregate.Flatten().InnerExceptions;
+                    // collected.AddRange(inner.OfType<SnowflakeDbException>());
```
sfc-gh-knozderko, line 73: comment?

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1841,7 +1851,7 @@ public void TestAsyncLoginTimeout()
         {
             using (var conn = new MockSnowflakeDbConnection())
             {
-                int timeoutSec = 5;
+                int timeoutSec = 15;
```
sfc-gh-knozderko, line 1854: is it for purpose or maybe it was a temporary change for testing?

Snowflake.Data.Tests/Util/SnowflakeDbExceptionAssert.cs
sfc-gh-mhofman: Fixed

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-mhofman: I've revert

sfc-gh-knozderko (approved):

PR #828, opened 2023-12-06
SNOW-950923 queryid for put/get statement

sfc-gh-mhofman:
### Description
Please explain the changes you made here.

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
@@ -208,7 +208,49 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 VerifyFilesAreUploaded(conn, new List<string> { t_inputFilePath }, t_internalStagePath);
             }
         }
+                
+        [Test]
+        public void TestPutFileProvidesQueryIdOnFailure()
+        {
+            // Arrange
+            // Set the PUT query variables but do not create a file
+            t_inputFilePath = "unexisting_file.csv";
+            t_internalStagePath = $"@{t_schemaName}.{t_stageName}";
+
+            // Act
+            using (var conn = new SnowflakeDbConnection(ConnectionString))
+            {
+                conn.Open();
+                var queryId = PutFile(conn);
+                
+                // Assert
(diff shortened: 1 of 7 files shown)


Snowflake.Data.Tests/IntegrationTests/SFPutGetTest.cs
```diff
@@ -474,29 +517,40 @@ private static bool IsCompressedByTheDriver()
...
-                    Assert.AreEqual(t_destCompressionType,
-                        reader.GetString((int)SFResultSet.PutGetResponseRowTypeInfo.DestinationCompressionType));
+                    var reader = command.ExecuteReader();
+                    Assert.IsTrue(reader.Read());
+                    // Checking query id when reader succeeded
+                    queryId = ((SnowflakeDbDataReader)reader).GetQueryId();
+                    // Checking if query Id is provided on the command level as well
+                    Assert.AreEqual(queryId, ((SnowflakeDbCommand)command).GetQueryId());
+                    // Check file status
+                    Assert.AreEqual(expectedStatus.ToString(),
```
sfc-gh-knozderko, line 529: now all these assert don't work, because exception is caught and not rethrown

sfc-gh-mhofman: if you refer to the exception from reader.Read how not rethrowing exception would affect asserts which are skipped in case of this exception?
Those asserts take place if there is a successful scenario.
Also all the tests pass right now. If there was an issue we would see some failures

sfc-gh-knozderko: command.ExecuteReader() does not throw any exception, but reader.Read() return false.
In that case previously the test would fail. Currently the test will pass.
I would prefer to add try/catch only around execute and return the function immediately and not putting these asserts into try/catch block, because it makes some scenarios untested.

sfc-gh-mhofman:
Closing as this change does not provide query id on the exception level but only at the statement level.
It is different than what is implemented in the JDBC Driver.
cc @sfc-gh-knozderko @sfc-gh-dszmolka

PR #832, opened 2023-12-13
SNOW-961434 Fix usage of insecure mode in connection parameters

sfc-gh-knozderko:
### Description
SNOW-961434 Fix usage of insecure mode in connection parameters

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/Session/SFHttpClientPropertiesTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/Session/SFHttpClientPropertiesTest.cs
@@ -15,11 +15,11 @@ namespace Snowflake.Data.Tests.UnitTests.Session
     public class SFHttpClientPropertiesTest
     {
         [Test]
-        public void ShouldConvertToMapOnly2Properties(
+        public void TestConvertToMapOnly2Properties(
             [Values(true, false)] bool validateDefaultParameters,
             [Values(true, false)] bool clientSessionKeepAlive)
         {
-            // given
+            // arrange
             var proxyProperties = new SFSessionHttpClientProxyProperties()
             {
                 proxyHost = "localhost",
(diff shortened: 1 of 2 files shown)


sfc-gh-fochnik (approved):

Snowflake.Data/Core/Session/SFSessionHttpClientProperties.cs
```diff
@@ -75,7 +75,7 @@ internal TimeSpan TimeoutDuration()
         internal HttpClientConfig BuildHttpClientConfig()
         {
             return new HttpClientConfig(
-                insecureMode,
+                !insecureMode,
```
sfc-gh-jkaluzny, line 78: Why would we even want to allow insecureMode?

sfc-gh-dszmolka: we allow it in _every_ Snowflake driver. the default setting is (except for this bug) to use certificate validation (OCSP / CRL) but we need to give the user the option , that knowing the risk and accepting it on their full own responsibility, to turn it off.
We never recommend to do it though, other than very well outlined, short test purposes.

sfc-gh-mhofman: more on this here https://community.snowflake.com/s/article/How-to-turn-off-OCSP-checking-in-Snowflake-client-drivers

Snowflake.Data.Tests/UnitTests/Session/SFHttpClientPropertiesTest.cs
```diff
@@ -79,7 +79,7 @@ public void ShouldBuildHttpClientConfig()
             var config = properties.BuildHttpClientConfig();
 
             // then
-            Assert.AreEqual(properties.insecureMode, config.CrlCheckEnabled);
+            Assert.AreEqual(!properties.insecureMode, config.CrlCheckEnabled);
```
sfc-gh-mhofman, line 82: one test run checks only true or false for this flag. Are we sure the other will succeed and is actually also in reverse of insecureMode?

sfc-gh-knozderko: I added more tests.

sfc-gh-mhofman (approved):

PR #834, opened 2023-12-14
fix dns names with underscore

sfc-gh-bculberson:
### Description
DNS names with underscores are not valid in the .net framework. This will construct a working HOST name by replacing with - which is how Snowflake solves for this use case, but users had to manually set HOST.

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Core/Session/SFSessionProperty.cs
+++ b/Snowflake.Data/Core/Session/SFSessionProperty.cs
@@ -260,7 +260,8 @@ namespace Snowflake.Data.Core
             if (!properties.ContainsKey(SFSessionProperty.HOST) ||
                 (0 == properties[SFSessionProperty.HOST].Length))
             {
-                string hostName = String.Format("{0}.snowflakecomputing.com", properties[SFSessionProperty.ACCOUNT]);
+                string compliantAccountName = properties[SFSessionProperty.ACCOUNT].Replace('_','-');
+                string hostName = String.Format("{0}.snowflakecomputing.com", compliantAccountName);
                 // Remove in case it's here but empty
                 properties.Remove(SFSessionProperty.HOST);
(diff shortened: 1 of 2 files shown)


gboucher90:
Hey,

We just encountered this issue in Production. Indeed it seems OpenSSL which is used by the .Net Core runtime implementation doesn't handle well underscores:
https://github.com/dotnet/runtime/issues/58722
https://mta.openssl.org/pipermail/openssl-dev/2015-August/002411.html

Would it be possible to have this code released?

Thank you!

sfc-gh-knozderko:
We are planning to release it in the nearest release.

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -260,7 +260,8 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
             if (!properties.ContainsKey(SFSessionProperty.HOST) ||
                 (0 == properties[SFSessionProperty.HOST].Length))
             {
-                string hostName = String.Format("{0}.snowflakecomputing.com", properties[SFSessionProperty.ACCOUNT]);
+                string compliantAccountName = properties[SFSessionProperty.ACCOUNT].Replace('_','-');
```
sfc-gh-pbulawa, line 263: We require a possibility to disable the replacement by the parameter in a connection string similar to JDBC to mitigate the risk of privatelink connections not working: [Reasoning for JDBC](https://community.snowflake.com/s/article/Behaviour-Change-Release-information-associated-with-Snowflake-JDBC-driver-version-3-13-25)

sfc-gh-pbulawa (changes requested):

sfc-gh-pbulawa:
PR #848 implements the feature. Thanks for the contribution.

PR #848, opened 2024-01-11
SNOW-993556: Replace underscores with hyphen

sfc-gh-pbulawa:
### Description
Enable automatic replacement of the underscores with hyphen with a possibility to disable it in similar way as JDBC.

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -90,7 +90,9 @@ internal enum SFSessionProperty
         [SFSessionPropertyAttr(required = false, defaultValue = "false")]
         DISABLEQUERYCONTEXTCACHE,
         [SFSessionPropertyAttr(required = false)]
-        CLIENT_CONFIG_FILE
+        CLIENT_CONFIG_FILE,
+        [SFSessionPropertyAttr(required = false, defaultValue = "false")]
+        ALLOWUNDERSCORESINHOST
```
sfc-gh-mhofman, line 95: Do we need to document this in the readme?

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -255,12 +257,30 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
...
             
             // compose host value if not specified
             if (!properties.ContainsKey(SFSessionProperty.HOST) ||
                 (0 == properties[SFSessionProperty.HOST].Length))
             {
-                string hostName = String.Format("{0}.snowflakecomputing.com", properties[SFSessionProperty.ACCOUNT]);
+                var compliantAccountName = properties[SFSessionProperty.ACCOUNT];
+                if (!allowUnderscoresInHost)
+                {
+                    compliantAccountName = compliantAccountName.Replace('_', '-');
```
sfc-gh-mhofman, line 281: is there a point in logging this fact that the account had been changed according to the flag?

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-pbulawa: It has been documented in the readme.

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-pbulawa: Good idea, added the log.

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-mhofman: my bad, didn't notice that file

sfc-gh-mhofman (approved):

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -255,12 +257,31 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
             checkSessionProperties(properties);
             ValidateFileTransferMaxBytesInMemoryProperty(properties);
             ValidateAccountDomain(properties);
+
+            var allowUnderscoresInHost = bool.Parse(SFSessionProperty.ALLOWUNDERSCORESINHOST.GetAttribute<SFSessionPropertyAttr>().defaultValue);
```
sfc-gh-knozderko, line 261: I would prefer to extract this retrieving of the flag value into a separate method. And then use here something like 
```
allowUnderscoresInHost = ParseAllowUnderscoresInHost(properties)
```

because this method parseConnectionString is about 150 lines long. I think we should not continue this style.

sfc-gh-pbulawa: Done

sfc-gh-knozderko (approved):

PR #867, opened 2024-02-15
SNOW-798828: Target .NET Standard 2.0

sfc-gh-ext-simba-lf:
### Description
Change the target framework to .NET Standard 2.0

### Checklist
- [ ] Code compiles correctly


.github/workflows/main.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dotnet: ['net6.0', 'net472', 'net471']
+        dotnet: ['net5.0', 'net6.0', 'net7.0', 'net8.0', 'net471', 'net472', 'net481']
```
sfc-gh-knozderko, line 37: In line 44 there is:
 ```
        uses: actions/setup-dotnet@v3
        with:
          dotnet-version: '6.0.x'
```
so we are using .NET 6 to build and run tests for .NET version 8. It looks suspicious to me. I would expect to have dotnet 8 version installed to test against dotnet 8.
If you look at .NET 8 build time executions they are only 2 or 3 minutes instead of 13 minutes. This is showing that the test weren't executed and we have false positive results.

**Review by sfc-gh-knozderko — CHANGES_REQUESTED** [review]

File: .github/workflows/main.yml (hunk 1 above)
Comment (sfc-gh-ext-simba-lf) [reply .github/workflows/main.yml:37]: Good catch, the .net versions 5, 7, and 8 are only running the tests for 1 second while the .NET Framework versions are running for at least 10 minutes

I'll add the missing net versions

Comment (sfc-gh-ext-simba-lf) [reply .github/workflows/main.yml:37]: Looks like the tests are running now after changing the `RuntimeFrameworkVersion` to 8

**Comment by sfc-gh-mhofman** [comment]
Let's plan to make a note in the readme and with the docs team that although we target Standard 2.0 we do not confirm nor run the tests on Mono/Xamarin/UWP/Unity? @sfc-gh-knozderko @sfc-gh-ext-simba-lf @sfc-gh-ext-simba-hx

File: Snowflake.Data/Snowflake.Data.csproj [hunk 2]
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net471;net472</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net6.0</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
sfc-gh-knozderko, line 3: we have in this file: `<ItemGroup Condition="'$(TargetFramework)' == 'net471' Or '$(TargetFramework)' == 'net472'">` Shouldn't we modify it?

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net471;net472</TargetFrameworks>
```
sfc-gh-knozderko, line 3: @sfc-gh-ext-simba-lf maybe update your PR to the current master, and then prepare also a modification of README.md. We want to write a disclaimar that we tested the library only on particular environments and do not guarantee how it will work with other environements. As soon as its ready, ask @sfc-gh-mheavin for review of the documentation.

sfc-gh-ext-simba-lf: Will do, for now the PR is updated to master

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net471;net472</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net6.0</TargetFrameworks>
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: The [commit from 2018](https://github.com/snowflakedb/snowflake-connector-net/commit/cd42190b7c9e409b5cb328fcc372b73f4aeb88a3#diff-8c6b62aeb3ce0a430174920a2eaa2eb9a628d9f6881f5ddba4dc8bb546a2021aR31) that added that line mentioned it was to `Allow user to config between different json parser and chunk downloader`

However, the part of the code that used `ConfigurationManager` was commented and I don't see it in master anymore

Snowflake.Data/Snowflake.Data.csproj
sfc-gh-ext-simba-lf: It can be removed since only netstandard2.0 is targeted

Snowflake.Data/Snowflake.Data.csproj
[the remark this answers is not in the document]
sfc-gh-knozderko: currently we do not target netstandard2.0 so this conditional reference was not applied. If everything works without it I would remove it.

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -21,7 +20,7 @@
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="14.0.2" />
     <PackageReference Include="AWSSDK.S3" Version="3.7.0.4" />
-    <PackageReference Include="Google.Cloud.Storage.V1" Version="4.6.0" />
+    <PackageReference Include="Google.Cloud.Storage.V1" Version="3.7.0" />
```
sfc-gh-knozderko, line 23: I don't like downgrading this dependency. I think we should resign from .netstandard.

sfc-gh-ext-simba-lf: It's possible to keep the original package version but there would be a build warning when its set higher than `3.7.0`

sfc-gh-ext-simba-lf: Set to the latest 4.10.0 which supports netstandard2.0

Snowflake.Data/Snowflake.Data.csproj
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Snowflake.Data.csproj
sfc-gh-ext-simba-lf: Removed

Snowflake.Data.Tests/UnitTests/ArrowResultSetTest.cs
```diff
@@ -275,12 +275,20 @@ public void TestGetText()
...
+            charArr.Add(TestDataGenarator.SnowflakeUnicode);
+            testValues = charArr.ToArray();
+#else
+            testValues =
+            TestDataGenarator.AsciiCodes.ToCharArray()
+                .Append(TestDataGenarator.SnowflakeUnicode)
+                .ToArray();
+#endif
+
+PrepareTestCase(SFDataType.TEXT, 0, testValues);
```
sfc-gh-knozderko, line 291: missing idents

Snowflake.Data.Tests/Util/SnowflakeDbExceptionAssert.cs
```diff
@@ -41,7 +41,7 @@ public static void HasHttpErrorCodeInExceptionChain(Exception exception, HttpSta
                         case SnowflakeDbException se:
                             return se.ErrorCode == (int)expected;
                         case HttpRequestException he:
-#if NETFRAMEWORK
+#if NETFRAMEWORK || NETCOREAPP
```
sfc-gh-knozderko, line 44: doesn't it cover all the cases? If so the `#if` is not needed

.github/workflows/main.yml
```diff
@@ -25,7 +25,7 @@ env:
   DOTNET_VERSION: 6.0
```
sfc-gh-knozderko, line 25: please have a look at:
- SFBaseTest line 175 and 179
- Snowflake.Data.Tests.csproj line 35
- ExternalBrowserAuthenticator line 207
- RestParams line 61

sfc-gh-knozderko (changes requested):

Snowflake.Data/Core/Authenticator/OktaAuthenticator.cs
```diff
@@ -80,7 +80,7 @@ async Task IAuthenticator.AuthenticateAsync(CancellationToken cancellationToken)
                     s_logger.Debug("step 4: Get SAML response from SSO");
                     var samlRestRequest = BuildSamlRestRequest(ssoUrl, onetimeToken);
                     samlRawResponse = await session.restRequester.GetAsync(samlRestRequest, cancellationToken).ConfigureAwait(false);
-#if NETFRAMEWORK                    
+#if NETFRAMEWORK || NETSTANDARD2_0
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: You're correct, only netstandard is being targeted now so it's not needed anymore. This is now removed

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -1,8 +1,8 @@
...
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net471;net472</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net6.0</TargetFrameworks>
-    <RuntimeFrameworkVersion>6.0.0</RuntimeFrameworkVersion>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0;net7.0;net8.0;net462;net471;net472;net48;net481</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net5.0;net6.0;net7.0;net8.0;</TargetFrameworks>
+    <RuntimeFrameworkVersion>8.0.0</RuntimeFrameworkVersion>
     <Title>Snowflake.Data.Tests</Title>
     <Description>Snowflake Connector for .NET</Description>
     <Company>Snowflake Computing, Inc</Company>
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Synced with master

Snowflake.Data.Tests/UnitTests/ArrowResultSetTest.cs
sfc-gh-ext-simba-lf: Added

.github/workflows/main.yml
```diff
@@ -25,7 +25,7 @@ env:
...
   DOTNET_LEGACY_VERSION: 4.7.1
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

.github/workflows/main.yml
sfc-gh-ext-simba-lf: The `#if` directives are removed for `ExternalBrowserAuthenticator` and `RestParams`

But the `Snowflake.Data.Tests` and `SFBaseTest` files need the `#if` directives since the project targets specific implementations (e.g. net462, net471, netcoreapp3.1, net5.0, …)

Snowflake.Data.Tests/Util/SnowflakeDbExceptionAssert.cs
sfc-gh-ext-simba-lf: The `if` directive is now removed

.github/workflows/main.yml
```diff
@@ -34,15 +34,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dotnet: ['net6.0', 'net472', 'net471']
+        dotnet: ['netcoreapp3.1', 'net5.0', 'net6.0', 'net7.0', 'net8.0', 'net462', 'net471', 'net472', 'net48', 'net481']
```
mungojam, line 37: .net core 3.1, .net 5 and .net 7 are no longer supported by Microsoft so I would personally drop support for them yourself to reduce maintenance requirements. .net 6 goes out of support in November

sfc-gh-ext-simba-lf: Thanks for the info, I'll let the team know about this and add it to consideration

sfc-gh-ext-simba-lf: Both `netcoreapp3.1` and `net5.0` are now removed

.github/workflows/main.yml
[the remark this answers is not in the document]
sfc-gh-knozderko: I see that you removed one assertion form ArrowResultChunkTest - but this was not what I meant.
My comment was about removing this line: https://github.com/snowflakedb/snowflake-connector-net/blob/20216e4b3eb3c31499924d2742cf98327793d085/Snowflake.Data/Core/ArrowResultChunk.cs#L141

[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: My mistake, I re-added the assertion and removed the Obsolete line that you specified

Snowflake.Data/Core/RestParams.cs
```diff
@@ -60,13 +60,8 @@ static SFEnvironment()
             {
                 application = System.Diagnostics.Process.GetCurrentProcess().ProcessName,
                 osVersion = System.Environment.OSVersion.VersionString,
-#if NETFRAMEWORK
-                netRuntime = "NETFramework",
```
sfc-gh-knozderko, line 64: I think you should extract this from runtime configuration.
See: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=net-6.0
The same applies to netVersion.
It would be good to extract it only once and then use it for every request.
It would be good if you wrote a test for proper netRuntime and netVersion extraction for given test environments.

sfc-gh-knozderko (changes requested):

Snowflake.Data/Snowflake.Data.csproj
mungojam: We have generally found it best to target both netstandard2.0 and net6.0,  net8.0 etc. I think it then means that all the extra dlls that get included when bringing in netstandard packages don't get included in the output as they're all included with the runtimes.

I might be out of date though

Snowflake.Data/Core/RestParams.cs
sfc-gh-ext-simba-lf: I was able to use `RuntimeInformation.FrameworkDescription` for the netRuntime property

However the version that comes from `RuntimeInformation.FrameworkDescription` seems different from what's expected:
- For .NET Framework it returns ".NET Framework 4.8.9246.0" for all the versions instead of ".NET Framework 471", ".NET Framework 472", or ".NET Framework 481"
- For .NET 6.0 and greater it returns ".NET 8.0.6" instead of ".NET 6.0", ".NET 7.0", or ".NET 8.0"

Is that the expected behavior?

sfc-gh-knozderko: not really... maybe try to get the version somehow differently from the assembly.

sfc-gh-ext-simba-lf: I tried to use the version from the assembly but it returned "netstandard2.0" since that's the target framework of the project . Should the `netVersion` value be set to `2.0` or `netstandard2.0` since that is the target version?

sfc-gh-ext-simba-lf: It's fixed now for .NET 6.0 and above

I'm not sure why the .NET Framework versions still returns "4.8" but it might be because it installs the latest .NET Framework version that is [compatible with previous versions](https://support.microsoft.com/en-us/topic/microsoft-net-framework-4-8-offline-installer-for-windows-9d23f658-3b97-68ab-d013-aa3c3e7495e0)
> The Microsoft .NET Framework 4.8 is a highly compatible, in-place update to the Microsoft .NET Framework 4, 4.5, 4.5.1, 4.5.2, 4.6, 4.6.1, 4.6.2, 4.7, 4.7.1 and 4.7.2.

sfc-gh-ext-simba-lf: The `Assembly.GetExecutingAssembly().Location` contains the accurate .NET Framework version and is now used for the `netVersion` value

Just for future reference, here's a list of what was previously tried and their results when running on .NET Framework:
1. `TargetFrameworkAttribute` of the executing assembly returns "netstandard2.0"
2. `TargetFrameworkAttribute` of the calling assembly returns "netstandard2.0"
3. The entry assembly returns null
4. `System.Environment` property returns "4.0.30319.42000"
5. `RuntimeInformation.FrameworkDescription` returns "4.8.4724.0"

sfc-gh-ext-simba-lf: It's fixed for both .NET 6.0+ and .NET Framework versions now

Snowflake.Data.Tests/UnitTests/SFEnvironmentTest.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Snowflake.Data.Tests.UnitTests
+{
+    [TestFixture]
+    class SFEnvironmentTest
+    {
+        [Test]
+        public void TestRuntimeExtraction()
+        {
+            // Arrange
+            var targetFrameworkAttribute = Assembly.GetExecutingAssembly().GetCustomAttributes(true).ToList().OfType<TargetFrameworkAttribute>().First();
```
sfc-gh-knozderko, line 16: the previous version:
```
...
#elif NET6_0
            expectedVersion = "6.0";
#elif NET7_0
            expectedVersion = "7.0";
#elif NET8_0
            expectedVersion = "8.0";
...
```

was very readable. Current version of tests does not give us so much readability.
Looking at this I'm not sure what the result will be under framework471 or net6.0.
Maybe you could restore the previous way of checking it in tests?
I'm not sure what is the problem here - if it is about that the version name has too much details e.g. 6.0.1 instead of 6.0 we could compare in tests only.
The problem is that if all these complicated methods like `Assembly.GetExecutingAssembly()` or `Runtime...` would return something which does not make sense (for example an empty string) then the test would not discover the problem since similar code is in 2 places: the production code and the test code. But if in test code we would compare the result with simple values like (".net", "6.0") and the method of getting the .net details would change, then the test would fail.

Snowflake.Data/Core/RestParams.cs
sfc-gh-knozderko: as I see you extract the location of the assembly and get the information from the directory name.
I don't like this idea since the library can be located anywhere - we don't have a guarantee that customers would store their dlls in similar form as it is when building the project.

Try maybe something else. If you would not find anything better I think using RuntimeInformation.FrameworkDescription would be better than using location of the dll.

Snowflake.Data.Tests/UnitTests/SFEnvironmentTest.cs
sfc-gh-ext-simba-lf: I agree, it was much easier to understand before

Reverted the change to the previous version

Snowflake.Data/Core/RestParams.cs
sfc-gh-ext-simba-lf: It's reverted to use the `RuntimeInformation.FrameworkDescription` again. I was not able to find a way to get framework version from other methods

sfc-gh-knozderko (approved):

sfc-gh-ext-simba-lf:
Updated readme with the new supported versions

PR #887, opened 2024-03-09
SNOW-817091: Async execution

sfc-gh-ext-simba-lf:
### Description
- Adds the option to execute queries in async mode
- Adds the capability to asynchronously wait for the query to finish and get the results using the query ID
- Adds checking the query status
- Adds checking if query is still running or encountered an error

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
@@ -1606,6 +1606,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
             {
                 conn.ConnectionString = ConnectionString;
                 conn.Open();
+
                 using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
                 {
                     // Arrange
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
@@ -516,7 +516,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
             }
         }
 
-        [Test, NonParallelizable]
+        [Test]
         public void TestGetResultsOfUnknownQueryIdAsync()
         {
             string unknownQueryId = "ab123fed-1abc-987f-987f-1234a56b789c";
@@ -1669,7 +1669,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
             }
         }
 
-        [Test, NonParallelizable]
+        [Test]
         public void TestGetResultsOfUnknownQueryId()
         {
             string unknownQueryId = "ba987def-1abc-987f-987f-1234a56b789c";
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
@@ -295,20 +295,11 @@ namespace Snowflake.Data.Tests.IntegrationTests
 
                     // Act
                     cancelToken.Cancel();
-#if NET471
                     var thrown = Assert.ThrowsAsync<OperationCanceledException>(async () =>
                         await cmd.GetResultsFromQueryIdAsync(queryId, cancelToken.Token).ConfigureAwait(false));
-#else
-                    var thrown = Assert.ThrowsAsync<TaskCanceledException>(async () =>
-                        await cmd.GetResultsFromQueryIdAsync(queryId, cancelToken.Token).ConfigureAwait(false));
-#endif
 
                     // Assert
-#if NET472
-                    Assert.IsTrue(thrown.Message.Contains("A task was canceled"));
-#else
                     Assert.IsTrue(thrown.Message.Contains("The operation was canceled"));
-#endif
                 }
 
                 await conn.CloseAsync(CancellationToken.None).ConfigureAwait(false);
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
@@ -206,7 +206,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     // Assert
                     Assert.IsTrue(reader.Read());
                     Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
-                    Assert.AreEqual(QueryStatus.SUCCESS, queryStatus);
+                    Assert.AreEqual(QueryStatus.Success, queryStatus);
                 }
 
                 await conn.CloseAsync(CancellationToken.None).ConfigureAwait(false);
@@ -277,7 +277,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     // Assert
                     Assert.IsTrue(reader.Read());
                     Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
-                    Assert.AreEqual(QueryStatus.SUCCESS, queryStatus);
+                    Assert.AreEqual(QueryStatus.Success, queryStatus);
                 }
 
                 await conn.CloseAsync(CancellationToken.None).ConfigureAwait(false);
@@ -403,7 +403,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     }
 
                     // Assert
-                    Assert.AreEqual(QueryStatus.FAILED_WITH_ERROR, queryStatus);
+                    Assert.AreEqual(QueryStatus.FailedWithError, queryStatus);
 
                     // Act
                     var thrown = Assert.ThrowsAsync<SnowflakeDbException>(async () =>
@@ -481,7 +481,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     var queryStatus = await cmd.GetQueryStatusAsync(unknownQueryId, CancellationToken.None).ConfigureAwait(false);
 
                     // Assert
-                    Assert.AreEqual(QueryStatus.NO_DATA, queryStatus);
+                    Assert.AreEqual(QueryStatus.NoData, queryStatus);
                 }
 
                 await conn.CloseAsync(CancellationToken.None).ConfigureAwait(false);
@@ -1405,7 +1405,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     // Assert
                     Assert.IsTrue(reader.Read());
                     Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
-                    Assert.AreEqual(QueryStatus.SUCCESS, cmd.GetQueryStatus(queryId));
+                    Assert.AreEqual(QueryStatus.Success, cmd.GetQueryStatus(queryId));
                 }
 
                 conn.Close();
@@ -1474,7 +1474,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     // Assert
                     Assert.IsTrue(reader.Read());
                     Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
-                    Assert.AreEqual(QueryStatus.SUCCESS, cmd.GetQueryStatus(queryId));
+                    Assert.AreEqual(QueryStatus.Success, cmd.GetQueryStatus(queryId));
                 }
 
                 conn.Close();
@@ -1523,7 +1523,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
 
                     // Assert
                     Assert.IsFalse(QueryStatuses.IsStillRunning(cmd.GetQueryStatus(queryIdTwo)));
-                    Assert.AreEqual(QueryStatus.SUCCESS, cmd.GetQueryStatus(queryIdTwo));
+                    Assert.AreEqual(QueryStatus.Success, cmd.GetQueryStatus(queryIdTwo));
                     Assert.IsFalse(conn.SfSession.StillRunningAsyncQueries());
                 }
 
@@ -1554,7 +1554,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     }
 
                     // Assert
-                    Assert.AreEqual(QueryStatus.FAILED_WITH_ERROR, cmd.GetQueryStatus(queryId));
+                    Assert.AreEqual(QueryStatus.FailedWithError, cmd.GetQueryStatus(queryId));
 
                     // Act
                     var thrown = Assert.Throws<SnowflakeDbException>(() => cmd.GetResultsFromQueryId(queryId));
@@ -1662,7 +1662,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                     var queryStatus = cmd.GetQueryStatus(unknownQueryId);
 
                     // Assert
-                    Assert.AreEqual(QueryStatus.NO_DATA, queryStatus);
+                    Assert.AreEqual(QueryStatus.NoData, queryStatus);
                 }
 
                 conn.Close();
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
@@ -175,6 +175,375 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 30 * 1000);
             }
         }
+
+        [Test]
+        public void TestAsyncExecQueryAsync()
+        {
+            string queryId;
+            var expectedWaitTime = 5;
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"CALL SYSTEM$WAIT({expectedWaitTime}, \'SECONDS\');";
+
+                    // Act
+                    var queryTask = cmd.ExecuteAsyncInAsyncMode(CancellationToken.None);
+                    queryTask.Wait();
+                    queryId = queryTask.Result;
+
+                    var statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                    statusTask.Wait();
+                    var queryStatus = statusTask.Result;
+
+                    // Assert
+                    Assert.IsTrue(QueryStatuses.IsStillRunning(queryStatus));
+
+                    // Act
+                    var readerTask = cmd.GetResultsFromQueryIdAsync(queryId, CancellationToken.None);
+                    readerTask.Wait();
+                    DbDataReader reader = readerTask.Result;
+
+                    statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                    statusTask.Wait();
+                    queryStatus = statusTask.Result;
+
+                    // Assert
+                    Assert.IsTrue(reader.Read());
+                    Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
+                    Assert.AreEqual(QueryStatus.SUCCESS, queryStatus);
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+        }
+
+        [Test]
+        public void TestMixedSyncAndAsyncQueryAsync()
+        {
+            string queryId;
+            var expectedWaitTime = 5;
+            Task task;
+
+            // Start the async exec query
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"CALL SYSTEM$WAIT({expectedWaitTime}, \'SECONDS\');";
+
+                    // Act
+                    var queryTask = cmd.ExecuteAsyncInAsyncMode(CancellationToken.None);
+                    queryTask.Wait();
+                    queryId = queryTask.Result;
+
+                    var statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                    statusTask.Wait();
+                    var queryStatus = statusTask.Result;
+
+                    // Assert
+                    Assert.IsTrue(QueryStatuses.IsStillRunning(queryStatus));
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+
+            // Execute a normal query
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                conn.Open();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"select 1;";
+
+                    // Act
+                    var row = cmd.ExecuteScalar();
+
+                    // Assert
+                    Assert.AreEqual(1, row);
+                }
+
+                conn.Close();
+            }
+
+            // Get results of the async exec query
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Act
+                    var readerTask = cmd.GetResultsFromQueryIdAsync(queryId, CancellationToken.None);
+                    readerTask.Wait();
+                    DbDataReader reader = readerTask.Result;
+
+                    var statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                    statusTask.Wait();
+                    var queryStatus = statusTask.Result;
+
+                    // Assert
+                    Assert.IsTrue(reader.Read());
+                    Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
+                    Assert.AreEqual(QueryStatus.SUCCESS, queryStatus);
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+        }
+
+        [Test]
+        public void TestStillRunningAsyncQueriesAsync()
+        {
+            string queryIdOne, queryIdTwo;
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"CALL SYSTEM$WAIT({3}, \'SECONDS\');";
+
+                    // Act
+                    var queryTask = cmd.ExecuteAsyncInAsyncMode(CancellationToken.None);
+                    queryTask.Wait();
+                    queryIdOne = queryTask.Result;
+
+                    // Arrange
+                    cmd.CommandText = $"CALL SYSTEM$WAIT({10}, \'SECONDS\');";
+
+                    // Act
+                    queryTask = cmd.ExecuteAsyncInAsyncMode(CancellationToken.None);
+                    queryTask.Wait();
+                    queryIdTwo = queryTask.Result;
+                    var stillRunningTask = conn.SfSession.StillRunningAsyncQueriesAsync(CancellationToken.None);
+                    stillRunningTask.Wait();
+                    var statusTask = cmd.GetQueryStatusAsync(queryIdOne, CancellationToken.None);
+                    statusTask.Wait();
+
+                    // Assert
+                    Assert.IsTrue(QueryStatuses.IsStillRunning(statusTask.Result));
+                    Assert.IsTrue(stillRunningTask.Result);
+
+                    // Act
+                    var readerTask = cmd.GetResultsFromQueryIdAsync(queryIdOne, CancellationToken.None);
+                    readerTask.Wait();
+                    statusTask = cmd.GetQueryStatusAsync(queryIdOne, CancellationToken.None);
+                    statusTask.Wait();
+                    stillRunningTask = conn.SfSession.StillRunningAsyncQueriesAsync(CancellationToken.None);
+                    stillRunningTask.Wait();
+
+                    // Assert
+                    Assert.IsFalse(QueryStatuses.IsStillRunning(statusTask.Result));
+                    Assert.IsTrue(QueryStatuses.IsStillRunning(cmd.GetQueryStatus(queryIdTwo)));
+                    Assert.IsTrue(stillRunningTask.Result);
+
+                    // Act
+                    readerTask = cmd.GetResultsFromQueryIdAsync(queryIdTwo, CancellationToken.None);
+                    readerTask.Wait();
+                    statusTask = cmd.GetQueryStatusAsync(queryIdTwo, CancellationToken.None);
+                    statusTask.Wait();
+                    stillRunningTask = conn.SfSession.StillRunningAsyncQueriesAsync(CancellationToken.None);
+                    stillRunningTask.Wait();
+
+                    // Assert
+                    Assert.IsFalse(QueryStatuses.IsStillRunning(statusTask.Result));
+                    Assert.IsFalse(stillRunningTask.Result);
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+        }
+
+        [Test]
+        public void TestFailedAsyncExecQueryThrowsErrorAsync()
+        {
+            string queryId;
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"SELECT * FROM FAKE_TABLE;";
+
+                    // Act
+                    var queryTask = cmd.ExecuteAsyncInAsyncMode(CancellationToken.None);
+                    queryTask.Wait();
+                    queryId = queryTask.Result;
+
+                    var statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                    statusTask.Wait();
+                    var queryStatus = statusTask.Result;
+
+                    while (QueryStatuses.IsStillRunning(queryStatus))
+                    {
+                        Thread.Sleep(1000);
+                        statusTask = cmd.GetQueryStatusAsync(queryId, CancellationToken.None);
+                        statusTask.Wait();
+                        queryStatus = statusTask.Result;
+                    }
+
+                    // Assert
+                    Assert.AreEqual(QueryStatus.FAILED_WITH_ERROR, queryStatus);
+
+                    // Act
+                    var readerTask = cmd.GetResultsFromQueryIdAsync(queryId, CancellationToken.None);
+                    var thrown = Assert.Throws<AggregateException>(() => readerTask.Wait());
+
+                    // Assert
+                    Assert.IsTrue(thrown.InnerException.Message.Contains("'FAKE_TABLE' does not exist"));
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+        }
+
+        [Test]
+        public void TestGetStatusOfInvalidQueryIdAsync()
+        {
+            string fakeQueryId = "fakeQueryId";
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    var statusTask = cmd.GetQueryStatusAsync(fakeQueryId, CancellationToken.None);
+
+                    // Act
+                    var thrown = Assert.Throws<AggregateException>(() => statusTask.Wait());
+
+                    // Assert
+                    Assert.IsTrue(thrown.InnerException.Message.Contains($"The given query id {fakeQueryId} is not valid uuid"));
+                }
+
+                task = conn.CloseAsync(CancellationToken.None);
+                task.Wait();
+            }
+        }
+
+        [Test]
+        public void TestGetResultsOfInvalidQueryIdAsync()
+        {
+            string fakeQueryId = "fakeQueryId";
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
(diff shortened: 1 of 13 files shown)


Snowflake.Data/Core/SFStatement.cs
```diff
@@ -309,19 +366,26 @@ private void registerQueryCancellationCallback(int timeout, CancellationToken ex
...
-                    var req = BuildResultRequest(lastResultUrl);
-                    response = await _restRequester.GetAsync<QueryExecResponse>(req, cancellationToken).ConfigureAwait(false);
-
-                    if (SessionExpired(response))
-                    {
-                        logger.Info("Ping pong request failed with session expired, trying to renew the session.");
-                        await SfSession.renewSessionAsync(cancellationToken).ConfigureAwait(false);
-                    }
-                    else
+                    SfSession.AsyncQueries.Add(response.data.queryId, QueryStatus.RUNNING);
```
sfc-gh-knozderko, line 371: is it possible that even though we send asyncExec = "false" we are getting the result immediately?

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -432,12 +496,13 @@ private SFBaseResultSet ExecuteSqlOtherThanPutGet(int timeout, string sql, Dicti
                     }
                 }
 
-                QueryExecResponse response =
-                    ExecuteHelper<QueryExecResponse, QueryExecResponseData>(
-                        timeout,
-                        sql,
-                        bindings,
-                        describeOnly);
+                    QueryExecResponse response =
```
sfc-gh-knozderko, line 499: too many idents

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -680,6 +763,138 @@ CancellationToken cancellationToken
...
+            var queryUri = SfSession.BuildUri(RestPath.SF_MONITOR_QUERY_PATH + queryId);
+
+            return new SFRestRequest
+            {
+                Url = queryUri,
+                authorizationToken = string.Format(SF_AUTHORIZATION_SNOWFLAKE_FMT, SfSession.sessionToken),
+                serviceName = SfSession.ParameterMap.ContainsKey(SFSessionParameter.SERVICE_NAME)
+                                ? (String)SfSession.ParameterMap[SFSessionParameter.SERVICE_NAME] : null,
+                HttpTimeout = Timeout.InfiniteTimeSpan,
+                RestTimeout = Timeout.InfiniteTimeSpan,
```
sfc-gh-knozderko, line 782: I'm just thinking if infinitive waiting for getting the query status is good... But from the other hand we do the same when executing a normal query.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,379 @@ public void TestExecuteAsyncWithMaxRetryReached()
...
+        {
+            string queryId;
+            var expectedWaitTime = 5;
+            Task task;
+
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString = ConnectionString;
+                task = conn.OpenAsync(CancellationToken.None);
+                task.Wait();
```
sfc-gh-knozderko, line 190: if you would change the signature of the test method to `public async Task TestAsyncExecQueryAsync()` then you could write just await conn.OpenAsync(CancellationToken.None).ConfigureAwait(false);
Maybe it would be a little bit shorter.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -1040,5 +1413,287 @@ public void TestGetQueryId()
...
+                    // Assert
+                    Assert.AreEqual(QueryStatus.NO_DATA, queryStatus);
+                }
+
+                conn.Close();
+            }
+        }
+
+        [Test]
+        public void TestGetResultsOfUnknownQueryId()
```
sfc-gh-knozderko, line 1677: when I run this test it is hanging

sfc-gh-knozderko (changes requested):

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
...
+                var sfStatement = new SFStatement(connection.SfSession);
+                return sfStatement.GetQueryStatus(queryId);
+            }
+            else
+            {
+                var errorMessage = $"The given query id {queryId} is not valid uuid";
+                logger.Error(errorMessage);
+                throw new Exception(errorMessage);
+            }
+
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -11,20 +11,27 @@
...
 
         private SnowflakeDbParameterCollection parameterCollection;
 
         private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
 
+        // Async max retry and retry pattern
+        private const int AsyncNoDataMaxRetry = 24;
+        private readonly int[] _asyncRetryPattern = { 1, 1, 2, 3, 4, 8, 10 };
+
+        private static readonly Regex UuidRegex = new Regex("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$");
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Based on [uuid docs](https://www.uuidtools.com/what-is-uuid#:~:text=Are%20UUIDs%20Case%2Dsensitive%3F,UUIDs%20in%20upper%20and%20lowercase.), it's not case sensitive so it can contain capital letters. But at the same time, it says UUID generators should only output in lowercase letters:
```No, UUIDs are written in base 16 which uses numbers 0-9 and characters a-f. There is no distinction between upper and lowercase letters. However, [RCF 4122 section 3](https://tools.ietf.org/html/rfc4122#section-3) requires that UUID generators output in lowercase and systems accept UUIDs in upper and lowercase.```

I'll modify the regex to include uppercase letters

File: Snowflake.Data/Client/SnowflakeDbCommand.cs [hunk 8]
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
             }
         }
 
+        /// <summary>
+        /// Execute a query in async mode.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: I added a short description of what async mode does in the summary. 

For the readme, I can write a similar thing and also give sample codes of how to run queries in async mode. I'm not sure if I can update the readme in this PR because usually we would ping someone from the docs team and then they would update the readme

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
...
         }
 
+        /// <summary>
+        /// Execute a query in async mode.
+        /// </summary>
+        /// <returns>The query id.</returns>
+        public string ExecuteInAsyncMode()
+        {
+            logger.Debug($"ExecuteInAsyncMode");
+            SFBaseResultSet resultSet = ExecuteInternal(asyncExec: true);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Good catch, I'll add a check for PUT/GET

I don't know the reason as well but python connector also doesn't support PUT/GET: https://github.com/snowflakedb/snowflake-connector-python/blob/23eda6e8e52e7f0b07997078b99e866f5e29be7f/src/snowflake/connector/cursor.py#L1082

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -17,6 +17,61 @@
...
+        RUNNING,
+        ABORTING,
+        SUCCESS,
+        FAILED_WITH_ERROR,
+        ABORTED,
+        QUEUED,
+        FAILED_WITH_INCIDENT,
+        DISCONNECTED,
+        RESUMING_WAREHOUSE,
+        // purposeful typo.Is present in QueryDTO.java
```
sfc-gh-knozderko, line 34: .java?

Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -577,6 +577,54 @@ internal bool IsExpired(long timeoutInSeconds, long utcTimeInSeconds)
         {
             return _startTime + timeoutInSeconds <= utcTimeInSeconds;
         }
+
+        /// <summary>
+        /// Contains the ID and status of queries executed in async mode
+        /// </summary>
+        internal Dictionary<string, QueryStatus> AsyncQueries = new Dictionary<string, QueryStatus>();
```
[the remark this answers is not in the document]
sfc-gh-knozderko: Actually I found the reason for such a logic. The reason is when we close the original http session then all running queries started in this session are becoming no longer available.

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -153,7 +153,10 @@ public override void Close()
             {
                 var transactionRollbackStatus = SnowflakeDbConnectionPool.GetPooling() ? TerminateTransactionForDirtyConnectionReturningToPool() : TransactionRollbackStatus.Undefined;
                 
-                if (CanReuseSession(transactionRollbackStatus) && SnowflakeDbConnectionPool.AddSession(SfSession))
+                if ((CanReuseSession(transactionRollbackStatus) ||
+                    SfSession.StillRunningAsyncQueries()) &&
```
sfc-gh-knozderko, line 157: why || ? shouldn't it be `CanReuseSession(transactionRollbackStatus) && !SfSession.StillRunningAsyncQueries() && SnowflakeDbConnectionPool.AddSession(SfSession)`?

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,379 @@ public void TestExecuteAsyncWithMaxRetryReached()
                 Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 30 * 1000);
             }
         }
+
```
sfc-gh-knozderko, line 178: it would be great to have a test showing how big problem would be revealing a queryId...
The scenario to test would be that in one connection we create an async long running query and in another one for different user/role/schema we try to get the result for given queryId. It would be good if we weren't allowed to get the result.

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
...
+        /// <returns>The query results.</returns>
+        public async Task<DbDataReader> GetResultsFromQueryIdAsync(string queryId, CancellationToken cancellationToken)
+        {
+            logger.Debug($"GetResultsFromQueryIdAsync");
+
+            int retryPatternPos = 0;
+            int noDataCounter = 0;
+
+            QueryStatus status;
+            while (true)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added cancelling when requested

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -11,20 +11,27 @@
...
 
         private SFStatement sfStatement;
 
         private SnowflakeDbParameterCollection parameterCollection;
 
         private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
 
+        // Async max retry and retry pattern
+        private const int AsyncNoDataMaxRetry = 24;
+        private readonly int[] _asyncRetryPattern = { 1, 1, 2, 3, 4, 8, 10 };
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: The values are from the python connector: https://github.com/snowflakedb/snowflake-connector-python/blob/23eda6e8e52e7f0b07997078b99e866f5e29be7f/src/snowflake/connector/cursor.py#L126

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
...
+        {
+            logger.Debug($"GetResultsFromQueryIdAsync");
+
+            int retryPatternPos = 0;
+            int noDataCounter = 0;
+
+            QueryStatus status;
+            while (true)
+            {
+                status = GetQueryStatus(queryId);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: You're right, it's replaced now

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -274,6 +281,179 @@ protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBeha
...
+        /// Gets the query results based on query ID.
+        /// </summary>
+        /// <param name="queryId"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>The query results.</returns>
+        public async Task<DbDataReader> GetResultsFromQueryIdAsync(string queryId, CancellationToken cancellationToken)
+        {
+            logger.Debug($"GetResultsFromQueryIdAsync");
+
+            int retryPatternPos = 0;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It's extracted into a function. I'll check about moving it into a different class

Snowflake.Data/Client/SnowflakeDbConnection.cs
```diff
@@ -176,7 +179,7 @@ public override async Task CloseAsync()
         }
 #endif
 
-        public virtual Task CloseAsync(CancellationToken cancellationToken)
+        public virtual async Task CloseAsync(CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: This existing test checks the CloseAsync() function: https://github.com/snowflakedb/snowflake-connector-net/blob/4d855888608e984be43f1005ce18a15aa55012de/Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs#L2023

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: I don't believe so, if it's set to false then it uses the original code and waits for the query to finish executing before returning

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: Fixed

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: Done. I'll do the same for other async tests

I was following how other async tests did it so I wasn't sure which one was preferred. I'll do it this way from now on

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: It's a typo from other connectors:
- Java: https://github.com/snowflakedb/snowflake-jdbc/blob/066e25a8718a26b39983b99e0731f3999b906ca6/src/main/java/net/snowflake/client/core/QueryStatus.java#L14
- Python: https://github.com/snowflakedb/snowflake-connector-python/blob/23eda6e8e52e7f0b07997078b99e866f5e29be7f/src/snowflake/connector/constants.py#L345

I'm not sure why but I think the server sends back a misspelled query status

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -17,6 +17,61 @@
...
+    /// </summary>
+    public enum QueryStatus
+    {
+        RUNNING,
+        ABORTING,
+        SUCCESS,
+        FAILED_WITH_ERROR,
+        ABORTED,
+        QUEUED,
+        FAILED_WITH_INCIDENT,
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

Since the enum values are based on the server, I modified the unit test to use the original values to confirm that they still work with PascalCase

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-ext-simba-lf: With `CanReuseSession(transactionRollbackStatus) && !SfSession.StillRunningAsyncQueries()`, is it possible that the session gets closed even if there are still async queries running? Like if the session can't be reused and there are still async queries being executed?

Snowflake.Data/Core/Session/SFSession.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: I reviewed the test I wrote once again and I found the bug in my testing. The original session can be closed and the queries are available. So there is no need to remember async queries on session. Let's remove them.

Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -577,6 +577,54 @@ internal bool IsExpired(long timeoutInSeconds, long utcTimeInSeconds)
...
+        /// <summary>
+        /// Contains the ID and status of queries executed in async mode
+        /// </summary>
+        internal Dictionary<string, QueryStatus> AsyncQueries = new Dictionary<string, QueryStatus>();
+
+        /// <summary>
+        /// Checks if the session contains async queries that are still executing
+        /// </summary>
+        /// <returns>True if session can be closed, false otherwise</returns>
+        internal bool StillRunningAsyncQueries()
```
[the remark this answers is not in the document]
sfc-gh-knozderko: Let's remove it because it found out that storing async execution on a session is useless.

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: The typo is `REPARING` from `QUEUED_REPARING_WAREHOUSE`

The part about QueryDTO.java is now removed

Snowflake.Data/Core/Session/SFSession.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Client/SnowflakeDbConnection.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Core/Session/SFSession.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: I tested this using two different accounts:
 - On the original account, the result was retrieved using the query id
 - On a different account, the query status was `NO_DATA` using the same query id even when the query already finished executing in the original account

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: Ok, let's not change README.md for now.

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: currently this check is not present in the code

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: But let's try to refactor more. Would it be possible to remove isAsync parameter from the method signature and instead add to different parameters:
- `Func<string, QueryStatus> queryStatusVerifier`
- Action<int, CancellationToken> waitingAction

Then we could create 2 functions for sync and async scenario. One would provide functions:
- queryId => GetQueryStatus(queryId)
- (seonds, _) => Thread.Sleep(...);
And the other method would have the async versions.

Check if such a refactor is possible.

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -433,11 +497,12 @@ private SFBaseResultSet ExecuteSqlOtherThanPutGet(int timeout, string sql, Dicti
                 }
 
                 QueryExecResponse response =
-                    ExecuteHelper<QueryExecResponse, QueryExecResponseData>(
-                        timeout,
-                        sql,
-                        bindings,
-                        describeOnly);
+                ExecuteHelper<QueryExecResponse, QueryExecResponseData>(
```
sfc-gh-knozderko, line 500: missing ident

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,289 @@ public void TestExecuteAsyncWithMaxRetryReached()
...
+        }
+
+        [Test]
+        public async Task TestMixedSyncAndAsyncQueryAsync()
+        {
+            string queryId;
+            var expectedWaitTime = 5;
+
+            // Start the async exec query
+            using (SnowflakeDbConnection conn = new SnowflakeDbConnection())
```
sfc-gh-knozderko, line 223: you are starting 3 connections one by one and closing them. But when you close the connection the session is returned to the pool so the next connection probably uses the same session. Do we want to demonstrate that they are separate sessions? If so then we should open 3 connections at the begining.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,289 @@ public void TestExecuteAsyncWithMaxRetryReached()
...
+                    Assert.AreEqual($"waited {expectedWaitTime} seconds", reader.GetString(0));
+                    Assert.AreEqual(QueryStatus.Success, queryStatus);
+                }
+
+                await conn.CloseAsync(CancellationToken.None).ConfigureAwait(false);
+            }
+        }
+
+        [Test]
+        public async Task TestMixedSyncAndAsyncQueryAsync()
```
sfc-gh-knozderko, line 217: maybe let's think about the better name for the test or changing what we test.
When I read  the name I thought that we test query executions in async mode and then we would demonstrate that  GetResultsFromQueryId(Async) - for both versions work properly. But in the test only GetResultsFromQueryIdAsync and GetQueryStatusAsync are tested and GetResultsFromQueryId and GetQueryStatus are not tested.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,289 @@ public void TestExecuteAsyncWithMaxRetryReached()
...
+                    // Act
+                    var queryId = await cmd.ExecuteAsyncInAsyncMode(CancellationToken.None).ConfigureAwait(false);
+                    var queryStatus = await cmd.GetQueryStatusAsync(queryId, CancellationToken.None).ConfigureAwait(false);
+
+                    // Assert
+                    Assert.IsTrue(QueryStatuses.IsStillRunning(queryStatus));
+
+                    // Act
+                    cancelToken.Cancel();
+#if NET471
```
sfc-gh-knozderko, line 310: what about NET472?
what about targeting different dotnet framewrks (I'm thinking about .netstandard 2.0)?

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,289 @@ public void TestExecuteAsyncWithMaxRetryReached()
...
+                    var thrown = Assert.ThrowsAsync<OperationCanceledException>(async () =>
+                        await cmd.GetResultsFromQueryIdAsync(queryId, cancelToken.Token).ConfigureAwait(false));
+#else
+                    var thrown = Assert.ThrowsAsync<TaskCanceledException>(async () =>
+                        await cmd.GetResultsFromQueryIdAsync(queryId, cancelToken.Token).ConfigureAwait(false));
+#endif
+
+                    // Assert
+#if NET472
```
sfc-gh-knozderko, line 319: what about NET471?
what about targeting different dotnet framewrks (I'm thinking about .netstandard 2.0)?

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: NET472 has a `TaskCanceledException` covered in the else statement

I'll check the other NET versions

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: NET471 has a `The operation was canceled` message covered in the else statement

I'll check the other NET versions

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -1040,5 +1323,270 @@ public void TestGetQueryId()
...
+                conn.Open();
+
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    cmd.CommandText = $"SELECT * FROM FAKE_TABLE;";
+
+                    // Act
+                    queryId = cmd.ExecuteInAsyncMode();
+                    while (QueryStatuses.IsStillRunning(cmd.GetQueryStatus(queryId)))
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Sure, added a break to the while loop after a certain amount of retries

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: I understand the test name can be confusing, it's now renamed to `TestExecuteNormalQueryWhileAsyncExecQueryIsRunning`

The purpose of the test was to show that it's possible to run a normal query while an async exec query is running

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: It's more about showing a normal query can be ran while an async exec query is already running. But using different sessions can be included in the test as well just to show that it works

Modified test to use different sessions

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -175,6 +175,279 @@ public void TestExecuteAsyncWithMaxRetryReached()
                 Assert.GreaterOrEqual(stopwatch.ElapsedMilliseconds, 30 * 1000);
             }
         }
+
```
sfc-gh-knozderko, line 178: we don't use in any test methods:
- GetResultsFromQueryId
- GetQueryStatus

It would be good to write a test which uses these methods as well.

Snowflake.Data/Core/SFStatement.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: You did it by ignoring case. I think that the enum with underscores (e. g. FAILED_WITH_INCIDENT) would not be parsed properly, because ignoring case doesn't solve the issue with uderscores.

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: Those methods are used in the non-async tests of `SFDbCommandIT`. One of the tests uses [both ](https://github.com/snowflakedb/snowflake-connector-net/blob/912f4a250eb7600ce2e04f0dd235d7770afbce9a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs#L1340) [methods](https://github.com/snowflakedb/snowflake-connector-net/blob/912f4a250eb7600ce2e04f0dd235d7770afbce9a/Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs#L1345)

Snowflake.Data/Core/SFStatement.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Currently, the underscores are removed before parsing

The enum is now assigned with a string value instead

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: My mistake, it got overwritten. It's re-added now

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: This is now fixed because the check for a cancelled token is re-added. Tested on all netstandard 2.0 supported versions and confirmed they all throw the same error

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: This is now fixed because the check for a cancelled token is re-added. Tested on all netstandard 2.0 supported versions and confirmed they all throw the same error

Snowflake.Data/Core/SFStatement.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: It would be good to add a unit test for all of the cases. You added the string values as a description.
This introduces a risk that someone would modify it since it may be regarded as a description only.
We could do a unit test for the values.
Also method GetQueryStatusByDescription could be renamed to GetQueryStatusByStringValue since it is a string value rather than a description.
I would also propose to rename class QueryStatuses to QueryStatusExtensions.

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -17,6 +17,61 @@
 
 namespace Snowflake.Data.Core
 {
+    /// <summary>
```
sfc-gh-knozderko, line 21: let's remove unused imports from this file

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
```diff
@@ -1040,5 +1304,260 @@ public void TestGetQueryId()
...
+                using (SnowflakeDbCommand cmd = (SnowflakeDbCommand)conn.CreateCommand())
+                {
+                    // Arrange
+                    var statusRetryCount = 0;
+                    cmd.CommandText = $"SELECT * FROM FAKE_TABLE;";
+
+                    // Act
+                    queryId = cmd.ExecuteInAsyncMode();
+                    while (statusRetryCount < 5 && QueryStatuses.IsStillRunning(cmd.GetQueryStatus(queryId)))
```
sfc-gh-knozderko, line 1418: let's create `var statusMaxRetryCount = 5` in the arrange section and use the variable here

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Core/RestParams.cs
```diff
@@ -39,6 +39,8 @@ internal static class RestPath
 
         internal const string SF_QUERY_PATH = "/queries/v1/query-request";
 
+        internal const string SF_MONITOR_QUERY_PATH = "/monitoring/queries/";
+
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Updated to master

Snowflake.Data/Core/SFStatement.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: > It would be good to add a unit test for all of the cases.

There's a unit test for all the statuses [here](https://github.com/snowflakedb/snowflake-connector-net/blob/d4994d2ac8e035bb573daf09e0166e3be4a4dfc0/Snowflake.Data.Tests/UnitTests/SFStatementTest.cs#L141)

Or by case did you mean like lowercase and uppercase?

> You added the string values as a description.
This introduces a risk that someone would modify it since it may be regarded as a description only

It can be changed to something like `StringValue` or `ServerValue` so it's clearer that's the value the server sends back

> We could do a unit test for the values.

Did you mean something like use the original string value and check if it's correctly parsed to the query status?

> Also method GetQueryStatusByDescription could be renamed to GetQueryStatusByStringValue since it is a string value rather than a description.

Done

> I would also propose to rename class QueryStatuses to QueryStatusExtensions.

Done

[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: > You used FirstOrDefault() method so in case of unknown string value in the response it would be converted to NO_DATA. Is it for purpose? If so, there should be a test for that. But maybe in case of unknown query status we should rather throw an exception because we actually don't now how to interpret given state?

I used `NO_DATA` as the default because if `RUNNING` was the default value then it would've kept retrying over and over

Adding an extra status to indicate it's unknown would be better since it's clearer if there's actually no data or if the status is just unknown

> Maybe let's make the comparison condition case insensitive.

Done

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: I tried it and passing async functions has an error: `Cannot convert async lambda expression to delegate type 'Func<string, QueryStatus>'. An async lambda expression may return void, Task or Task<T>, none of which are convertible to 'Func<string, QueryStatus>'`
```
RetryUntilQueryResultIsAvailable(
    qId,
    cancellationToken,
    async queryId => await GetQueryStatusAsync(queryId, cancellationToken).ConfigureAwait(false),
    async (seconds, _) => await Task.Delay(TimeSpan.FromSeconds(seconds), cancellationToken).ConfigureAwait(false));
```

It's possible to turn the `Func` param to return a Task instead to make it work `Func<Task<QueryStatus>> queryStatusVerifier`

But afterwards I noticed some of the async tests finish much faster than they should because it wasn't actually waiting for the timeout. So I had to convert the `Action` param into `Func<int, CancellationToken, Task>` to call Wait() and that seems to fix it

I haven't committed these changes yet, I'll let you know when they're in

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: Could you try again and check if it still happens? The test runs for ~3.5 minutes because it retries until the max data retry is reached

Snowflake.Data/Core/SFStatement.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Other than the above, these changes have been committed as well:
- Added Unknown status for queries and throw exception if unknown
- Added test for GetQueryStatusByStringValue in lowercase and uppercase
- Replaced `Description` attribute with `StringAttr` attribute

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -3,20 +3,107 @@
...
+using System.ComponentModel;
 
 namespace Snowflake.Data.Core
 {
+    /// <summary>
+    /// The status types of the query.
+    /// </summary>
+    public enum QueryStatus
+    {
+        [StringAttr(value = "UNKNOWN")]
```
sfc-gh-knozderko, line 23: let's not introduce "unknown" enum value just to throw an exception for case of an unknown value.
Someone could think that "unknown" is one of legal values.
Instead you can write it such a way:
```
            var statuses = Enum.GetValues(typeof(QueryStatus))
                .Cast<QueryStatus>()
                .Where(v => v.GetAttribute<StringAttr>().value.Equals(description, StringComparison.OrdinalIgnoreCase));
            return statuses.Any() ? statuses.First() : throw new Exception("The query status returned by the server is not recognized");
```

or maybe just something like that would work:
```
Enum.Parse(typeof(QueryStatus, description, true))
```

Snowflake.Data.Tests/UnitTests/SFStatementTest.cs
```diff
@@ -118,9 +119,93 @@ public void TestTrimSqlLineCommentWithClosingNewline()
...
+        [TestCase("ABORTED", QueryStatus.Aborted)]
+        [TestCase("failed_with_incident", QueryStatus.FailedWithIncident)]
+        [TestCase("FAILED_WITH_INCIDENT", QueryStatus.FailedWithIncident)]
+        [TestCase("disconnected", QueryStatus.Disconnected)]
+        [TestCase("DISCONNECTED", QueryStatus.Disconnected)]
+        [TestCase("restarted", QueryStatus.Restarted)]
+        [TestCase("RESTARTED", QueryStatus.Restarted)]
+        [TestCase("blocked", QueryStatus.Blocked)]
+        [TestCase("BLOCKED", QueryStatus.Blocked)]
+        public void TestGetQueryStatusByStringValue(string status, QueryStatus expectedStatus)
```
sfc-gh-knozderko, line 157: let's split it into 2 tests: one for positive cases and one for negative ones (errors)

Snowflake.Data/Core/SFStatement.cs
sfc-gh-knozderko: what about `using System.ComponentModel;`?

Snowflake.Data/Core/SFStatement.cs
```diff
@@ -3,20 +3,107 @@
...
+    }
+
+    class StringAttr : Attribute
+    {
+        public string value { get; set; }
+    }
+
+    internal static class QueryStatusExtensions
+    {
+        internal static QueryStatus GetQueryStatusByStringValue(string description)
```
sfc-gh-knozderko, line 61: rename description to value or stringValue

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: I see that it gets a little complicated, so maybe let's not do this refactoring I suggested - we can leave it as it is now

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: Removed as well. Sorry, I posted that before replacing the `Description` attribute

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-knozderko: 3,5 min is a lot of time. Let's exclude the retry logic into a separate class e.g. QueryResultsAwaiter.
This class would have the default configuration (you have to move it):
```
        private const int AsyncNoDataMaxRetry = 24;
        private readonly int[] _asyncRetryPattern = { 1, 1, 2, 3, 4, 8, 10 };
```

but would have also an internal constructor allowing to provide another configuration.
Then you could write a test for some shorter waiting scenario.
And we would not test it for the original configuration, because 3,5 min is too much of time.

The class SnowflakeDbCommand would hold QueryResultsAwaiter instance and use it.

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: Renamed

sfc-gh-knozderko (approved):
But first fix all the comments.

Snowflake.Data.Tests/UnitTests/SFStatementTest.cs
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Client/SnowflakeDbCommand.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Ok

Snowflake.Data/Core/SFStatement.cs
sfc-gh-ext-simba-lf: Done, removed the `Unknown` enum and refactored using the above

Snowflake.Data.Tests/IntegrationTests/SFDbCommandIT.cs
sfc-gh-ext-simba-lf: Done. The PR will already be merged but please feel free to let me know if there's any follow up changes you'd like to see

sfc-gh-ext-simba-lf:
I added utility functions `IsStillRunning` and `IsAnError` to `SnowflakeDbConnection` for parity with other connectors: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#is_still_running

PR #889, opened 2024-03-13
SNOW-979288: Add explicit DbType Parameter assignment

sfc-gh-ext-simba-lf:
### Description
Add explicit DbType Parameter assignment

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbParameter.cs
+++ b/Snowflake.Data/Client/SnowflakeDbParameter.cs
@@ -15,6 +15,8 @@ namespace Snowflake.Data.Client
 
         private SFDataType OriginType;
 
+        private DbType _dbType;
+
         public SnowflakeDbParameter()
         {
             SFDataType = SFDataType.None;
@@ -34,7 +36,12 @@ namespace Snowflake.Data.Client
             this.SFDataType = SFDataType;
         }
 
-        public override DbType DbType { get; set; }
+        public override DbType DbType
+        {
+            get => _dbType == DbType.AnsiString ? SFDataConverter.TypeToDbTypeMap[Value.GetType()] : _dbType;
+
+            set => _dbType = value;
+        }
 
         public override ParameterDirection Direction
         {
--- a/Snowflake.Data/Core/SFDataConverter.cs
+++ b/Snowflake.Data/Core/SFDataConverter.cs
@@ -3,6 +3,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Globalization;
 using System.Text;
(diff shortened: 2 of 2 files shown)


Snowflake.Data/Client/SnowflakeDbParameter.cs
```diff
@@ -34,7 +36,28 @@ public SnowflakeDbParameter(int ParameterIndex, SFDataType SFDataType)
             this.SFDataType = SFDataType;
         }
 
-        public override DbType DbType { get; set; }
+        public override DbType DbType
+        {
+            get
+            {
+                if (_dbType != DbType.AnsiString || Value == null || Value is DBNull)
```
sfc-gh-mhofman, line 43: what happens when _dbType and Value is null?

Snowflake.Data.Tests/UnitTests/SFDbParameterTest.cs
```diff
@@ -125,5 +127,75 @@ public void TestDbParameterResetDbType([Values] SFDataType expectedSFDataType)
             _parameter.ResetDbType();
             Assert.AreEqual(SFDataType.None, _parameter.SFDataType);
         }
+
+        [Test]
+        public void TestDbTypeExplicitAssignment([Values] DbType expectedDbType)
```
sfc-gh-mhofman, line 132: what if DbType is not provided?
I would suggest to have a special edge case tests for:
- null value and (just one) non null DbType
- non null value and null DbType
- both - DbType and Value nulls

Snowflake.Data/Client/SnowflakeDbParameter.cs
sfc-gh-ext-simba-lf: I tried to test this but only `Value` can be set to null. The `_dbType` can't be set to null because of this error: `Cannot convert null to 'type' because it is a non-nullable value type`

It looks like `_dbType` will always default to `DbType.AnsiString` and the code would return the default `DbType`

Snowflake.Data.Tests/UnitTests/SFDbParameterTest.cs
sfc-gh-ext-simba-lf: I added a test case for the first item on the list. And I also added a test case for null value and default DbType

But I'm not able to add a test case for the other two because `DbType` is a non-nullable value type. It always defaults to `DbType.AnsiString`

sfc-gh-mhofman: Thanks @sfc-gh-ext-simba-lf !

sfc-gh-mhofman (approved):

PR #890, opened 2024-03-13
SNOW-1232318 fix account name validation

sfc-gh-knozderko:
### Description
SNOW-1232318 Fix account name validation to allow account name be even 1 or 2 chars long

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
@@ -27,11 +27,34 @@ namespace Snowflake.Data.Tests.UnitTests
         }
 
         [Test]
+        [TestCase("a", "a", "a.snowflakecomputing.com")]
+        [TestCase("ab", "ab", "ab.snowflakecomputing.com")]
+        [TestCase("a.b", "a", "a.b.snowflakecomputing.com")]
+        [TestCase("a-b", "a-b", "a-b.snowflakecomputing.com")]
+        [TestCase("a_b", "a_b", "a-b.snowflakecomputing.com")]
+        [TestCase("abc", "abc", "abc.snowflakecomputing.com")]
+        public void TestValidateCorrectAccountNames(string accountName, string expectedAccountName, string expectedHost)
+        {
+            // arrange
+            var connectionString = $"ACCOUNT={accountName};USER=test;PASSWORD=test;";
+            
+            // act
+            var properties = SFSessionProperties.parseConnectionString(connectionString, null);
+            
+            // assert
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -117,8 +117,13 @@ class SFSessionProperties : Dictionary<SFSessionProperty, String>
                 SFSessionProperty.PRIVATE_KEY_PWD,
                 SFSessionProperty.PROXYPASSWORD,
             };
-        
-        private const string AccountRegexString = "^\\w[\\w.-]+\\w$";
+
+        private static readonly List<string> s_accountRegexStrings = new List<string>
```
sfc-gh-mhofman, line 121: why not just single
`^[\w.-]+$`

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-pmotacki: it is too general- works for example 'ss_dd_dd_dd_dd.d..d' .  I'm thinking about ^\w+(.|-)?\w+$` but still underscore is in grup represented by \w. The underscore is a correct separator?

sfc-gh-knozderko: @sfc-gh-mhofman it would allow the account name to start or finish with `.` or `-`

sfc-gh-knozderko: @sfc-gh-pmotacki what you are thinking doesn't solve the problem. For example 1 letter account would match to the regex

sfc-gh-mhofman: Agreed on this version

sfc-gh-mhofman (approved):

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -26,12 +26,35 @@ public void TestThatPropertiesAreParsed(TestCase testcase)
             CollectionAssert.AreEquivalent(testcase.ExpectedProperties, properties);
         }
 
+        [Test]
+        [TestCase("a", "a", "a.snowflakecomputing.com")]
```
sfc-gh-mhofman, line 30: lets use real case scenarios from the docs:
https://docs.snowflake.com/en/user-guide/admin-account-identifier

sfc-gh-knozderko: added an example from doc

sfc-gh-mhofman (approved):

PR #891, opened 2024-03-18
SNOW-1230345 Fix to support quoted values in object connection properties

sfc-gh-jmartinez:
### Description
 Fix to support quoted values in object connection properties.
Supported object properties with scaped quotes:
- DB
- SCHEMA
- ROLE
- WAREHOUSE

### Checklist
- [ ] Code compiles correctly


sfc-gh-jmartinez (commented):

sfc-gh-jmartinez:

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -11,14 +11,16 @@
 
 namespace Snowflake.Data.Tests.UnitTests
 {
+    using System;
```
sfc-gh-erojaslizano, line 14: move using outside namespace

sfc-gh-jmartinez: Fixed, it was removed from the code

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -134,6 +135,74 @@ public static IEnumerable<TestCase> ConnectionStringTestCases()
                     { SFSessionProperty.ALLOWUNDERSCORESINHOST, defAllowUnderscoresInHost }
                 }
             };
+
+            var testCaseWithEmptyUserAndPassword = new TestCase()
```
sfc-gh-knozderko, line 139: these tests are following the the pattern as it was written earlier. It is OK, but maybe we could write them this way: https://github.com/snowflakedb/snowflake-connector-net/blob/4d855888608e984be43f1005ce18a15aa55012de/Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs#L37 The problem with earlier approach was that it tested many properties at once even though we would like to test only one aspect (e. g. that the user and password can be empty). So my proposal would be to try to write it another way.

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -175,14 +175,14 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
                                 SFError.INVALID_CONNECTION_STRING,
                                 e.Message);
             }
-            SFSessionProperties properties = new SFSessionProperties();
+            var properties = new SFSessionProperties();
 
-            string[] keys = new string[builder.Keys.Count];
-            string[] values = new string[builder.Values.Count];
+            var keys = new string[builder.Keys.Count];
```
sfc-gh-knozderko, line 180: I went through the code but I only noticed a refactoring. Were there any behavioural changes?
Do I understand correctly that the code was working well and all you did was refactoring (without behavioural changes) and writing tests?

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -292,6 +265,49 @@ internal static SFSessionProperties parseConnectionString(String connectionStrin
             return properties;
         }
 
+        private static void ProcessPropertySpecialCases(string connectionString, SFSessionProperties properties)
```
sfc-gh-knozderko, line 268: I like that you thought of extracting some block of code into a separate method.
However this method does a tricky thing - it updates its parameter - properties values.
Normally we do not expect that a method updates its parameters - in such case a method usually returns the modified object.
So it would be good to give it a more descriptive name at least. What would say about `UpdatePropertiesForSpecialCases` ?

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-jmartinez: The issue is in this line https://github.com/snowflakedb/snowflake-connector-net/blob/master/Snowflake.Data/Core/Session/SFSessionProperty.cs#L207, the condition was value.ToLower() == "WAREHOUSE" so it will always return false. So that was the fixed behavior.

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-jmartinez: I renamed the method with the suggested name. Thank you

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
sfc-gh-jmartinez: The new tests are refactored using the new approach.

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -75,13 +76,46 @@ public void TestThatItFailsIfNoAccountSpecified(string connectionString)
...
+        {
+            // arrange
+            var connectionString = $"ACCOUNT=test;USER=;PASSWORD=;";
+            
+            // act
+            var properties = SFSessionProperties.ParseConnectionString(connectionString, null);
+            
+            // assert
+            Assert.AreEqual(string.Empty, properties[SFSessionProperty.USER]);
+            Assert.AreEqual(string.Empty, properties[SFSessionProperty.USER]);
```
sfc-gh-knozderko, line 116: it should be password not user

sfc-gh-knozderko (approved):

PR #897, opened 2024-03-26
SNOW-1271212 Fixed values uploaded to stage for bindings exceeding CLIENT_STAGE_ARRAY_BINDING_THRESHOLD

sfc-gh-mhofman:
### Description
When number of binded values during query execution exceeds the threshold of a session parameter CLIENT_STAGE_ARRAY_BINDING_THRESHOLD then values are written as a CSV file to a stage and it get's picked during query excecution.
Improper values (or values truncating fractions of ms) has been uploaded for date and time related columns of type: DATE, TIME, TIMESTAMP_LTZ, TIMESTAMP_NTZ, TIMESTAMP_TZ. 

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,283 @@ public void testPutArrayBind1()
...
+        [TestCase(ResultFormat.ARROW, SFTableType.Iceberg, SFDataType.TIME, 6, DbType.Time, FormatHmsf, null)]
+        [TestCase(ResultFormat.ARROW, SFTableType.Iceberg, SFDataType.TIMESTAMP_NTZ, 6, DbType.DateTime, FormatYmdHms, null)]
+        // [TestCase(ResultFormat.ARROW, SFTableType.Iceberg, SFDataType.TIMESTAMP_TZ, 6, DbType.DateTime, FormatYmdHmsZ, null)] // Unsupported data type 'TIMESTAMP_TZ(6)' for iceberg tables
+        [TestCase(ResultFormat.ARROW, SFTableType.Iceberg, SFDataType.TIMESTAMP_LTZ, 6, DbType.DateTimeOffset, FormatYmdHmsZ, null)]
+        */
+        // Session TimeZone cases
+        [TestCase(ResultFormat.ARROW, SFTableType.Standard, SFDataType.TIMESTAMP_LTZ, 6, DbType.DateTimeOffset, FormatYmdHmsZ, "Europe/Warsaw")]
+        [TestCase(ResultFormat.JSON, SFTableType.Hybrid, SFDataType.TIMESTAMP_LTZ, 6, DbType.DateTimeOffset, FormatYmdHmsZ, "Asia/Tokyo")]
+        public void TestDateTimeBindingWithValue(ResultFormat resultFormat, SFTableType tableType, SFDataType columnType, Int32? columnPrecision, DbType bindingType, string comparisonFormat, string timeZone)
```
sfc-gh-knozderko, line 828: maybe let's write an unit test for all these timestamps. So a similar test as this, but without inserting to a real table.
This test should be located in unit tests scope.
The test should verify that when we call:
- `SFDataConverter.csharpTypeValToSfTypeVal(DbType srcType, object srcVal)`
- `SFBindUploader.GetCSVData(string sType, string sValue)` - we could make this method internal

as a result we are getting what we put as a parameter.

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,283 @@ public void testPutArrayBind1()
...
+                    new[] { 
+                        "id number(10,0) not null primary key", // necessary only for HYBRID tables
+                        $"ts {columnWithPrecision}" 
+                    }, 
+                    tableType.TableCreationFlags());
+
+                // Act
+                var sqlInsert = $"insert into {TableName} (id, ts) values (?, ?)";
+                
+                // insert 1 row with value binding
```
sfc-gh-knozderko, line 860: maybe lets try to extract each of the blocks into a method. If it turned out very complicated it may stay as is. But you could try.

Snowflake.Data.Tests/UnitTests/SFBindUploaderTest.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void TestCsvDataConversionForTime(SFDataType dbType, string input, string expected)
+        {
+            DateTime timeExpected = DateTime.Parse(expected);
+            DateTime timeActual = DateTime.Parse(_bindUploader.GetCSVData(dbType.ToString(), input));
+            Assert.AreEqual(timeExpected, timeActual);
+        }
+        
+        [TestCase(SFDataType.TIMESTAMP_LTZ, "0", "2012-07-01T12:00:00.0000000+02:00")]
+        [TestCase(SFDataType.TIMESTAMP_LTZ, "1341144000000000000", "2012-07-01T12:00:00.0000000+02:00")]
+        [TestCase(SFDataType.TIMESTAMP_LTZ, "352245599987654000", "1981-02-28T23:59:59.9876540+02:00")]
```
sfc-gh-knozderko, line 40: It will work in CET timezone only? What if someone will run the tests in another timezone?

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Done - SFBindUploaderTest.

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Done.

Snowflake.Data.Tests/UnitTests/SFBindUploaderTest.cs
sfc-gh-mhofman: Fixed, upload happens after converting to local tz, during check expected value is also in local tz.

sfc-gh-mhofman: Checked manually in 3 different timezones (we can work it out later to swap tz during the test) - US/NY, JPN/Tokyo, PL/WAW

Snowflake.Data/Core/SFBindUploader.cs
```diff
@@ -246,33 +246,36 @@ private string GetCSVData(string sType, string sValue)
...
                     long dateLong = long.Parse(sValue);
-                    DateTime date = dateTime.AddMilliseconds(dateLong).ToUniversalTime();
+                    DateTime date = epoch.AddMilliseconds(dateLong);
                     return date.ToShortDateString();
                 case "TIME":
+                    // SFDateConverter.csharpValToSfVal provides in [ns]
                     long timeLong = long.Parse(sValue);
-                    DateTime time = dateTime.AddMilliseconds(timeLong).ToUniversalTime();
-                    return time.ToLongTimeString();
+                    DateTime time = epoch.AddTicks(timeLong/100);
```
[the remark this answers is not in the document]
sfc-gh-mhofman: I improved local var naming - should be clear now what happens here and why the changes needed

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,290 @@ public void testPutArrayBind1()
...
+        private ExpectedTimestampValueProvider(DateTimeOffset dateTimeOffset, SFDataType columnType)
+        {
+            _expectedDateTimeOffset = dateTimeOffset;
+            _expectedDateTime = null;
+            _columnType = columnType;
+        }
+
+        internal SFDataType ExpectedColumnType() => _columnType;
+
+        internal void IsEqual(object actual, string comparisonFormat, string faultMessage)
```
sfc-gh-knozderko, line 958: better name would be AssertEqual

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,290 @@ public void testPutArrayBind1()
...
+    class ExpectedTimestampValueProvider
+    {
+        private readonly SFDataType _columnType;
+        private readonly DateTime? _expectedDateTime;
+        private readonly DateTimeOffset? _expectedDateTimeOffset;
+
+        internal static ExpectedTimestampValueProvider From(string timestampWithTimeZone, SFDataType columnType)
+        {
+            if (IsOffsetType(columnType))
+                return new ExpectedTimestampValueProvider(DateTimeOffset.ParseExact(timestampWithTimeZone,
```
sfc-gh-knozderko, line 933: personally I wouldn't inline the ParseExact method - splitting it into 2 steps would be more readable I think

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,290 @@ public void testPutArrayBind1()
...
+        private readonly DateTimeOffset? _expectedDateTimeOffset;
+
+        internal static ExpectedTimestampValueProvider From(string timestampWithTimeZone, SFDataType columnType)
+        {
+            if (IsOffsetType(columnType))
+                return new ExpectedTimestampValueProvider(DateTimeOffset.ParseExact(timestampWithTimeZone, 
+                        "yyyy/MM/dd HH:mm:ss.fff zzz", 
+                        CultureInfo.InvariantCulture), columnType); 
+                    
+            return new ExpectedTimestampValueProvider(DateTime.ParseExact(timestampWithTimeZone,
```
sfc-gh-knozderko, line 937: the same here

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,290 @@ public void testPutArrayBind1()
...
+                        Assert.IsNull(((SnowflakeDbCommand)insert).GetBindStage());
+                }
+            }
+        }
+
+        private static string ColumnTypeWithPrecision(SFDataType columnType, Int32? columnPrecision) 
+            => columnPrecision != null ? $"{columnType}({columnPrecision})" : $"{columnType}";
+    }
+    
+    class ExpectedTimestampValueProvider
```
sfc-gh-knozderko, line 924: Provider should provide something, but this class does not provide anything, it is rather a test case. What do you think about the name TimeoutTestCase?

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -769,5 +774,290 @@ public void testPutArrayBind1()
...
+    }
+    
+    public enum SFTableType 
+    {
+        Standard,
+        Hybrid,
+        Iceberg
+    }
+    
+    static class TestExtensions
```
sfc-gh-knozderko, line 994: this is not used, the tests are using SFBaseTest. I'm writing a comment just to let you think if we should add it in this PR since there is no usage of this code. Another question would be why in this 1000 lines long SFBindTestID?

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: It does provide an expected value in the required by columnType format which can be either DateTime or DateTimeOffset. Doesn't it make sense?

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Code doesn't compile without this code. It's an integral part of this test.
I would maybe even consider in the future those Add, CreateCommand, ExecuteNonQuery methods become a production interface. For now if needed we can move it outside this test class although some of it work along just with the ExpectedTimeoutValueProvider class. So why to move it away from here?

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Fixed

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Agree, fixed here as well

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: And it is not a TestCase alone. I could get rid of this class and in the result some of the new test lines
require an if expression related to columnType, SFDataType etc

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: For now moved it so slim this class. Have a look @sfc-gh-knozderko

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
sfc-gh-mhofman: Changed as agreed

Snowflake.Data/Core/SFBindUploader.cs
```diff
@@ -224,13 +224,13 @@ internal async Task UploadStreamAsync(MemoryStream stream, string destFileName,
...
-        private string GetCSVData(string sType, string sValue)
+        
+        internal string GetCSVData(string sType, string sValue)
         {
             if (sValue == null)
                 return sValue;
 
-            DateTime dateTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Unspecified);
-            DateTimeOffset dateTimeOffset = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Unspecified);
+            DateTime epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Unspecified);
```
sfc-gh-pbulawa, line 233: Can we keep the epoch as a const somewhere, also wouldn't it be more predictable to keep it as `DateTimeKind.Utc`? In .Net standard 2.1, the `Utc` kind is used in `DateTime.UnixEpoch`.

Snowflake.Data.Tests/IntegrationTests/SFBindTestIT.cs
```diff
@@ -14,10 +17,13 @@ namespace Snowflake.Data.Tests.IntegrationTests
     using System.Text;
     using System.Globalization;
     using System.Collections.Generic;
+    using Snowflake.Data.Tests.Util;
```
sfc-gh-knozderko, line 20: move all this using commands to the top (before namespace)

Snowflake.Data.Tests/Util/DbCommandTestExtensions.cs
```diff
@@ -0,0 +1,18 @@
+using System.Data;
+
+namespace Snowflake.Data.Tests.Util
+{
+    public static class DbCommandTestExtensions
```
sfc-gh-knozderko, line 5: what do you think about the name DbCommandExtensions?

Snowflake.Data.Tests/Util/DbConnectionTestExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using System.Data;
+
+namespace Snowflake.Data.Tests.Util
+{
+    public static class DbConnectionTestExtensions
```
sfc-gh-knozderko, line 5: what do you think about the name DbConnectionExtensions?

sfc-gh-pbulawa (approved):

sfc-gh-knozderko (approved):

PR #899, opened 2024-03-26
SNOW-977565: Added start and end symbol to match full regex to bypass proxy server.

sfc-gh-jmartinez:
### Description
Added start and end symbol to match string with full regex to bypass proxy server.

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,57 @@ public void TestInvalidProxySettingFromConnectionString()
             }
         }
 
+        [Test]
+        [TestCase("*")]
+        [TestCase("*{0}*")]
+        [TestCase("^*{0}*$")]
```
sfc-gh-mhofman, line 1546: maybe add a testcase for ^ without $ and vice versa as well regarding the change in HttpUtil

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -187,6 +189,10 @@ internal HttpMessageHandler SetupCustomHttpHandler(HttpClientConfig config)
                         // * -> .*  because * is a quantifier and need a char or group to apply to
                         entry = entry.Replace("*", ".*");
 
+                        if (!entry.StartsWith("^") || !entry.EndsWith("$"))
```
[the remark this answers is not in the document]
sfc-gh-jmartinez: It works, the regex supports multiple ^ or $ and matches the expression, but I will change it to avoid extra symbols in the expression.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,57 @@ public void TestInvalidProxySettingFromConnectionString()
...
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var nonProxyHosts = string.Format(regexHost, $"{testConfig.host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
+                conn.ConnectionString =
+                    $"{ConnectionString}connection_timeout=5;USEPROXY=true;PROXYHOST={proxyHostForConnection};NONPROXYHOSTS={nonProxyHosts};PROXYPORT=3128;";
+                try
+                {
+                    conn.Open();
+                    //Assert.Fail();
```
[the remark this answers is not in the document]
sfc-gh-jmartinez: Changed to use Assert.Throws to verify exception type if an exception is not thrown the test will fail.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Added requested cases

sfc-gh-mhofman (approved):

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,51 @@ public void TestInvalidProxySettingFromConnectionString()
...
+        {
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var host = ResolveHostToUseFromTestConfig(ConnectionString);
+                var nonProxyHosts = string.Format(regexHost, $"{host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
+                conn.ConnectionString =
+                    $"{ConnectionString}USEPROXY=true;PROXYHOST={proxyHostForConnection};NONPROXYHOSTS={nonProxyHosts};PROXYPORT=3128;";
+                conn.Open();
+                Assert.AreEqual(ConnectionState.Open, conn.State);
```
sfc-gh-knozderko, line 1562: let's write a comment how successful connecting proves that we bypassed the proxy server because it could be not obvious to the reader. Also verify if it is possible to write a unit test instead of integration tests, because integration tests are much slower.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,51 @@ public void TestInvalidProxySettingFromConnectionString()
...
+        [TestCase("la", "la")]
+        public void TestNonProxyHostShouldNotBypassProxyServer(string regexHost, string proxyHost = null)
+        {
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var nonProxyHosts = string.Format(regexHost, $"{testConfig.host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
+                conn.ConnectionString =
+                    $"{ConnectionString}connection_timeout=5;USEPROXY=true;PROXYHOST={proxyHostForConnection};NONPROXYHOSTS={nonProxyHosts};PROXYPORT=3128;";
+                var exception = Assert.Throws<SnowflakeDbException>(() => conn.Open());
```
sfc-gh-knozderko, line 1581: let's write a comment how exception on connecting proves that we didn't bypass the proxy server because it could be not obvious to the reader. Also verify if it is possible to write a unit test instead of integration tests, because integration tests are much slower.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,51 @@ public void TestInvalidProxySettingFromConnectionString()
...
+        [TestCase("^invalid{0}$")]
+        [TestCase("*a.b")]
+        [TestCase("a", "a")]
+        [TestCase("la", "la")]
+        public void TestNonProxyHostShouldNotBypassProxyServer(string regexHost, string proxyHost = null)
+        {
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var nonProxyHosts = string.Format(regexHost, $"{testConfig.host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
```
sfc-gh-knozderko, line 1578: I would rather pass "proxyserverhost" directly as proxyHost without default null value. Maybe it would be a little code duplication, but the test would be a little bit more readable.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,51 @@ public void TestInvalidProxySettingFromConnectionString()
...
+        [TestCase("^nonmatch*{0}$|*")]
+        [TestCase("*a*", "a")]
+        [TestCase("*la*", "la")]
+        public void TestNonProxyHostShouldBypassProxyServer(string regexHost, string proxyHost = null)
+        {
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var host = ResolveHostToUseFromTestConfig(ConnectionString);
+                var nonProxyHosts = string.Format(regexHost, $"{host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
```
sfc-gh-knozderko, line 1558: I would rather pass "proxyserverhost" directly as proxyHost without default null value. Maybe it would be a little code duplication, but the test would be a little bit more readable.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1790,6 +1835,11 @@ public void TestKeepAlive()
                 Assert.Fail();
             }
         }
+        
+        private string ResolveHostToUseFromTestConfig(string connectionString)
```
sfc-gh-knozderko, line 1839: connectionString is not used, such a method should be rather in the base class where testConfig is defined

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,51 @@ public void TestInvalidProxySettingFromConnectionString()
...
+        public void TestNonProxyHostShouldBypassProxyServer(string regexHost, string proxyHost = null)
+        {
+            using (var conn = new SnowflakeDbConnection())
+            {
+                var host = ResolveHostToUseFromTestConfig(ConnectionString);
+                var nonProxyHosts = string.Format(regexHost, $"{host}");
+                var proxyHostForConnection = proxyHost ?? "proxyserverhost";
+                conn.ConnectionString =
+                    $"{ConnectionString}USEPROXY=true;PROXYHOST={proxyHostForConnection};NONPROXYHOSTS={nonProxyHosts};PROXYPORT=3128;";
+                conn.Open();
```
sfc-gh-knozderko, line 1561: you could apply arrange/act/assert pattern

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-knozderko, line 1581: you could apply arrange/act/assert pattern

sfc-gh-knozderko (changes requested):

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -186,7 +188,11 @@ internal HttpMessageHandler SetupCustomHttpHandler(HttpClientConfig config)
                         entry = entry.Replace(".", "[.]");
                         // * -> .*  because * is a quantifier and need a char or group to apply to
                         entry = entry.Replace("*", ".*");
-
+                        
+                        entry = entry.StartsWith("^") ? entry : $"^{entry}";
```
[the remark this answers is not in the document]
sfc-gh-jmartinez: In the case of the jdbc driver it has the same behavior, it uses string.matches function so the full string should match with the host value, so if we provide snowflakecomputing.com as the nonproxyhosts it will not be bypassed.

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -18,6 +18,8 @@
 
 namespace Snowflake.Data.Core
 {
+    using System.Text.RegularExpressions;
```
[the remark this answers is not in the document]
sfc-gh-jmartinez: Done

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: I change the default value of the method parameter to use "proxyserverhost" instead of null. Thank you!!

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Moved to base class and removed unused parameter.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: I change the default value of the method parameter to use "proxyserverhost" instead of null. Thank you!!

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Added comment to specify how the test is handled. This is an integration test because we are validating the external WebProxy which has the functionality to determine if the proxy server is bypassed or not.

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Added comment to specify how the test is handled. This is an integration test because we are validating the external WebProxy which has the functionality to determine if the proxy server is bypassed or not.

sfc-gh-jmartinez: Applied arrange/act/assert pattern. Thank you!!!

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Applied arrange/act/assert pattern. Thank you!!!

README.md
```diff
@@ -377,6 +377,22 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+
+Examples:
+
+- `*` (Bypassed all hosts from the proxy server)
+- `*.snowflakecomputing.com` ('Bypass all host that ends with `snowflakecomputing.com`')
+- `https:\\testaccount.snowflakecomputing.com` (Bypass proxy server using full host url).
+- `*.myserver.com | *testaccount*` (You can specify multiple regex for the property divided by `|`)
+
+
+> Note: If you only specify the hostname in the property if will not be bypass because should match the full url including the http or https section.
```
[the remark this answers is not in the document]
sfc-gh-jmartinez: Done

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-knozderko: you wrote: "If the connection is open, it indicates that the server proxy was bypassed."
This is still explaining WHAT is happening not WHY. We should only comment about WHY's.
Maybe something like: "The connection would fail to open if the web proxy would be used because the proxy is configured to a non-existent host".

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-knozderko: you wrote: "If the nonproxyhosts property fails to bypass the server, the proxy will be use but it will return a exception because is not configure in the testing environment so the exception is used to verify that was not bypass."
It contains the reason WHY - that is good. But maybe we could use the shorter form as I suggested in the previous case: "The connection would fail to open if the web proxy would be used because the proxy is configured to a non-existent host"

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-knozderko: now the builds failed because you are still passing the unneeded parameter to this method. You could also think about a shorter name like `ResolveHost()`

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1540,6 +1540,62 @@ public void TestInvalidProxySettingFromConnectionString()
...
+                conn.ConnectionString =
+                    $"{ConnectionString}connection_timeout=5;USEPROXY=true;PROXYHOST={proxyHost};NONPROXYHOSTS={nonProxyHosts};PROXYPORT=3128;";
+                
+                // Act/Assert
+                // If the nonproxyhosts property fails to bypass the server, the proxy will be use but it will return a exception
+                // because is not configure in the testing environment so the exception is used to verify that was not bypass.
+                var exception = Assert.Throws<SnowflakeDbException>(() => conn.Open());
+                
+                // Assert
+                s_logger.Debug("Failed opening connection ", exception);
```
sfc-gh-knozderko, line 1593: it is OK, but normally we don't add logging to tests because it doesn't facilitate anything.
This log will not be visible anywhere because we have WARN log level set by default.
If we would like to check what the exception was exactly we could add a breakpoint here, run the test in debug mode and see the exception.
So I personally would not add such a debug statements, but I see that it was used in this class a few times so it could stay, but personally I don't like such things.

sfc-gh-knozderko (changes requested):

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Done

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Done

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Removed logger, thank you!!

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-knozderko: you could skip `this` in `this.ResolveHost()`

sfc-gh-knozderko (approved):

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-jmartinez: Done

PR #931, opened 2024-04-24
SNOW-1344341 make new pool as a default one

sfc-gh-knozderko:
### Description
Make new pool as a default one.

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/Session/ConnectionPoolManager.cs
```diff
@@ -85,39 +91,39 @@ public long GetTimeout()
...
 
-        public bool GetPooling() 
+        public bool GetPooling()
         {
             s_logger.Debug("ConnectionPoolManager::GetPooling");
-            var values = _pools.Values.Select(it => it.GetPooling()).Distinct().ToList();
-            return values.Count == 1
-                ? values.First()
-                : throw new SnowflakeDbException(SFError.INCONSISTENT_RESULT_ERROR, "Multiple pools have different Pooling values");
+            return true; // in new pool pooling is always enabled by default, disabling only by connection string parameter
```
sfc-gh-mhofman, line 119: but this method is a proxy to an old pool so it will always say true even when a user disables pooling on a cache pool

Snowflake.Data.Tests/Util/PoolConfig.cs
```diff
@@ -16,15 +17,17 @@ class PoolConfig
...
+            _timeout = (long) SFSessionHttpClientProperties.DefaultExpirationTimeout.TotalSeconds;
+            _pooling = SFSessionHttpClientProperties.DefaultPoolingEnabled;
+            _connectionPoolType = SnowflakeDbConnectionPool.DefaultConnectionPoolType;
         }
 
         public void Reset()
         {
             SnowflakeDbConnectionPool.SetConnectionPoolVersion(_connectionPoolType);
+            if (_connectionPoolType == ConnectionPoolType.MultipleConnectionPool)
+                return; // for multiple connection pool setting parameters for all the pools doesn't work by design
```
sfc-gh-mhofman, line 30: minor but it could be visible as a debug information so to track improper use of the interface within tests

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
```diff
@@ -168,38 +168,25 @@ public void TestSetPoolingForAllPoolsDisabled()
...
-            sessionPool1.SetPooling(true);
-            sessionPool2.SetPooling(true);
-            
-            // Act/Assert
-            Assert.AreEqual(true,_connectionPoolManager.GetPooling());
+            // Act
+            var pooling = _connectionPoolManager.GetPooling();
+
+            // Assert
+            Assert.IsTrue(pooling);
```
sfc-gh-mhofman, line 189: check my comment on the ConnectionPoolManager, imho it's misleading. Imagine a developer logging pool enabled flag using this interface - I would rather throw an error than return true in such a case

Snowflake.Data.Tests/IntegrationTests/ConnectionMultiplePoolsAsyncIT.cs
```diff
@@ -88,5 +177,23 @@ public async Task TestReleaseConnectionWhenRollbackFailsAsync()
             // assert
             Assert.AreEqual(0, pool.GetCurrentPoolSize(), "Should not return connection to the pool");
         }
+
+        [Test(Description = "test connection pooling with concurrent connection using async calls")]
+        public void TestConcurrentConnectionPoolingAsync()
+        {
+            // add test case name in connection string to make in unique for each test case
+            // set short expiration timeout to cover the case that connection expired
+            string connStr = ConnectionString + ";application=TestConcurrentConnectionPoolingAsync2;ExpirationTimeout=3";
```
sfc-gh-mhofman, line 186: does it need to end with "2" - TestConcurrentConnectionPoolingAsync2?

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/Session/ConnectionPoolManager.cs
sfc-gh-knozderko: no this class ConnectionPoolManager is for new pool only. For old pool it is a class ConnectionCacheManager.
If the user disables pooling on old pool he will be using ConnectionCacheManager class.

Snowflake.Data.Tests/Util/PoolConfig.cs
sfc-gh-knozderko: but the log level in tests is warn so it won't be visible either way

Snowflake.Data.Tests/IntegrationTests/ConnectionMultiplePoolsAsyncIT.cs
sfc-gh-knozderko: this test is actually a performance tests. We have the same tests in ConnectionSinglePoolCacheAsyncIT - I wanted to differentiate application name for these tests.

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
sfc-gh-knozderko: We have several places in code when we call `SnowflakeDbConnectionPool.GetPooling()` for example in `SnowflakeDbConnection.CanReuseSession`. So if we would do such a change everything is getting more complicated. From the other side, returning always true for new pool makes sense - because generally pooling is always enabled, you can disable it only in particular pools by connection string parameter.

sfc-gh-mhofman: Right, now you need to take i-th session pool and call GetPooling on it

sfc-gh-mhofman (approved):

PR #934, opened 2024-04-26
SNOW-937188 pool mode where changed session gets destroyed [WIP]

sfc-gh-mhofman:
### Description
When ChangedSession=Destroy is set in a connection string and a session is affected with a USE DATABASE/SCHEMA/ROLE/WAREHOUSE it is treated as no-longer matching original pool and gets destroyed on closing.
Its a non default pool behavior.

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/IntegrationTests/ConnectionChangedSessionIT.cs
```diff
@@ -0,0 +1,160 @@
...
+using Snowflake.Data.Client;
+using Snowflake.Data.Core;
+using Snowflake.Data.Core.Session;
+using Snowflake.Data.Tests.Util;
+
+namespace Snowflake.Data.Tests.IntegrationTests
+{
+    [TestFixture]
+    [NonParallelizable]
+    public class ConnectionChangedSessionIT : SFBaseTest
```
sfc-gh-knozderko, line 11: maybe ConnectionMultiplePoolsChangedSessionIT? it would be more visible that it is a test for the new pool

Snowflake.Data.Tests/IntegrationTests/ConnectionChangedSessionIT.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        private readonly QueryExecResponseData _queryExecResponseChangedRole = new()
```
sfc-gh-knozderko, line 13: BTW do we want it to work for old pool as well? As I see the implementation - it would work for old pool as well. Don't we want to cover that in tests?

sfc-gh-knozderko (changes requested):

Snowflake.Data.Tests/IntegrationTests/ConnectionChangedSessionIT.cs
sfc-gh-knozderko, line 13: I think all of the parameters that can cause not returning session to the pool should be covered in tests (so making one of them excluded from list should make some test to fail). But currently I think the situation is different. For example if the role has changed - it is only checked in the test for default pool mode which is OriginalPool, and is not covered in Destroy test. So I would add more unit tests to SFSessionTests.

sfc-gh-knozderko (changes requested):

Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -46,8 +46,10 @@ public class SFSession
         internal SFSessionProperties properties;
 
         internal string database;
-
         internal string schema;
+        internal string role;
```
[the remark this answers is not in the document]
sfc-gh-mhofman: there's no such thing in login's response session data
https://github.com/snowflakedb/snowflake-connector-net/blob/15f4c556ddb8d2487a8c1612a174be841ee35e63/Snowflake.Data/Core/RestResponse.cs#L127

Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -460,20 +464,30 @@ internal RequestQueryContext GetQueryContextRequest()
...
+            UpdateSessionProperty(ref warehouse, responseData.finalWarehouseName);
+        }
+
+        private void UpdateSessionProperty(ref string initialSessionValue, string finalSessionValue)
+        {
+            // with HTAP session metadata removal database/schema might be not returned in query result
+            if (!String.IsNullOrEmpty(finalSessionValue))
             {
-                this.schema = schemaName;
+                if (!String.IsNullOrEmpty(initialSessionValue) && initialSessionValue != finalSessionValue)
```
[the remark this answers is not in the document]
sfc-gh-mhofman: but then different objects would be treated equally

Snowflake.Data.Tests/IntegrationTests/ConnectionChangedSessionIT.cs
sfc-gh-mhofman: Agree but on the other side any changed is handled equally
https://github.com/snowflakedb/snowflake-connector-net/pull/934/files#diff-c36410e4608c5d3a815bdfa7327569a75af2ead26f4e5e67ab0f5f421120f6bdR476
I'm not sure we want to run each test 16 times in this case?

sfc-gh-mhofman: lets have it only in latest pool (the least changes to prev the better)

PR #941, opened 2024-05-13
SNOW-1373131 disable connection pooling for external browser authentication

sfc-gh-dstempniak:
### Description
Disable connection pooling for external browser authentication.

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -914,6 +914,30 @@ public void TestSSOConnectionWithUser()
...
+        {
+            // Use external browser to log in using proper password for qa@snowflakecomputing.com
+            using (IDbConnection conn = new SnowflakeDbConnection())
+            {
+                conn.ConnectionString
+                    = ConnectionStringWithoutAuth
+                      + ";authenticator=externalbrowser;user=qa@snowflakecomputing.com;POOLINGENABLED=TRUE";
+                conn.Open();
+                Assert.AreEqual(ConnectionState.Open, conn.State);
+                Assert.AreEqual(true, SnowflakeDbConnectionPool.GetPool(conn.ConnectionString).GetPooling());
```
sfc-gh-knozderko, line 940: why true? the feature was about enforcing it to be false?

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -192,7 +192,41 @@ public static IEnumerable<TestCase> ConnectionStringTestCases()
...
+                    { SFSessionProperty.INCLUDERETRYREASON, defIncludeRetryReason },
+                    { SFSessionProperty.DISABLEQUERYCONTEXTCACHE, defDisableQueryContextCache },
+                    { SFSessionProperty.DISABLE_CONSOLE_LOGIN, defDisableConsoleLogin },
+                    { SFSessionProperty.ALLOWUNDERSCORESINHOST, defAllowUnderscoresInHost },
+                    { SFSessionProperty.MAXPOOLSIZE, DefaultValue(SFSessionProperty.MAXPOOLSIZE) },
+                    { SFSessionProperty.MINPOOLSIZE, DefaultValue(SFSessionProperty.MINPOOLSIZE) },
+                    { SFSessionProperty.CHANGEDSESSION, DefaultValue(SFSessionProperty.CHANGEDSESSION) },
+                    { SFSessionProperty.WAITINGFORIDLESESSIONTIMEOUT, DefaultValue(SFSessionProperty.WAITINGFORIDLESESSIONTIMEOUT) },
+                    { SFSessionProperty.EXPIRATIONTIMEOUT, DefaultValue(SFSessionProperty.EXPIRATIONTIMEOUT) },
+                    { SFSessionProperty.POOLINGENABLED, "true" }
```
sfc-gh-knozderko, line 229: why true?

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -192,7 +192,41 @@ public static IEnumerable<TestCase> ConnectionStringTestCases()
                     { SFSessionProperty.CHANGEDSESSION, DefaultValue(SFSessionProperty.CHANGEDSESSION) },
                     { SFSessionProperty.WAITINGFORIDLESESSIONTIMEOUT, DefaultValue(SFSessionProperty.WAITINGFORIDLESESSIONTIMEOUT) },
                     { SFSessionProperty.EXPIRATIONTIMEOUT, DefaultValue(SFSessionProperty.EXPIRATIONTIMEOUT) },
-                    { SFSessionProperty.POOLINGENABLED, DefaultValue(SFSessionProperty.POOLINGENABLED) }
+                    { SFSessionProperty.POOLINGENABLED, "false" } // by default pooling is disabled for externalbrowser authentication
+                }
+            };
+            var testCaseWithExternalBrowserAndPoolingEnabled = new TestCase()
```
sfc-gh-knozderko, line 198: it works, and we have some tests written in this manner, but I would prefer the way of writing test like TestValidateCorrectAccountNames() - so a dedicated test verifying only the properties which are important from given test point of view.

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -296,6 +296,19 @@ private static void ValidateAuthenticator(SFSessionProperties properties)
                     logger.Error(error);
                     throw new SnowflakeDbException(SFError.UNKNOWN_AUTHENTICATOR, authenticator);
                 }
+
+                if (authenticator.Equals(ExternalBrowserAuthenticator.AUTH_NAME))
```
sfc-gh-knozderko, line 300: I think this logic would fit better to SFSessionHttpClientProperties class.
In SFSessionHttpClientProperties we interpret string values as proper types values and we do a logic of changing the values.
SFSessionProperty is operates only on string values rather without interpreting them.

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -296,6 +296,19 @@ private static void ValidateAuthenticator(SFSessionProperties properties)
                     logger.Error(error);
                     throw new SnowflakeDbException(SFError.UNKNOWN_AUTHENTICATOR, authenticator);
                 }
+
+                if (authenticator.Equals(ExternalBrowserAuthenticator.AUTH_NAME))
+                {
+                    if (!properties.TryGetValue(SFSessionProperty.POOLINGENABLED, out var poolingEnabledStr))
```
sfc-gh-knozderko, line 302: should all this logic be for old pool as well?

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -296,6 +296,19 @@ private static void ValidateAuthenticator(SFSessionProperties properties)
...
+                if (authenticator.Equals(ExternalBrowserAuthenticator.AUTH_NAME))
+                {
+                    if (!properties.TryGetValue(SFSessionProperty.POOLINGENABLED, out var poolingEnabledStr))
+                    {
+                        properties.Add(SFSessionProperty.POOLINGENABLED, "false");
+                        logger.Info("Connection pooling is disabled for external browser authentication");
+                    }
+                    else if (Boolean.TryParse(poolingEnabledStr, out var poolingEnabled) && poolingEnabled)
+                    {
+                        logger.Warn("Connection pooling is enabled for external browser authentication");
```
sfc-gh-knozderko, line 309: we should enforce false for the new pool I belive

sfc-gh-knozderko (changes requested):

PR #942, opened 2024-05-13
SNOW-1373257 Make secure password be a part of pool key

sfc-gh-knozderko:
### Description
SNOW-1373257 Make secure password be a part of pool key

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
@@ -3,7 +3,6 @@
  */
 
 using System;
-using System.Net;
 using System.Security;
 using System.Threading;
 using System.Threading.Tasks;
@@ -12,6 +11,7 @@ using Snowflake.Data.Core;
 using Snowflake.Data.Core.Session;
 using Moq;
 using Snowflake.Data.Client;
+using Snowflake.Data.Core.Tools;
 using Snowflake.Data.Tests.Util;
 
 namespace Snowflake.Data.Tests.UnitTests
@@ -22,10 +22,8 @@ namespace Snowflake.Data.Tests.UnitTests
         private readonly ConnectionPoolManager _connectionPoolManager = new ConnectionPoolManager();
         private const string ConnectionString1 = "db=D1;warehouse=W1;account=A1;user=U1;password=P1;role=R1;minPoolSize=1;";
         private const string ConnectionString2 = "db=D2;warehouse=W2;account=A2;user=U2;password=P2;role=R2;minPoolSize=1;";
-        private const string ConnectionString3 = "db=D3;warehouse=W3;account=A3;user=U3;role=R3;minPoolSize=1;";
-        private readonly SecureString _password1 = null;
-        private readonly SecureString _password2 = null;
-        private readonly SecureString _password3 = new NetworkCredential("", "P3").SecurePassword;
+        private const string ConnectionStringWithoutPassword = "db=D3;warehouse=W3;account=A3;user=U3;role=R3;minPoolSize=1;";
+        private readonly SecureString _password3 = SecureStringHelper.Encode("P3");
         private static PoolConfig s_poolConfig;
 
         [OneTimeSetUp]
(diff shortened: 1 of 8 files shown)


Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -119,6 +118,17 @@ private void CleanExpiredSessions()
...
 
+        internal void ValidateSecurePassword(SecureString password)
+        {
+            if (!ExtractPassword(Password).Equals(ExtractPassword(password)))
+            {
+                throw new Exception("Could not get a pool because of password mismatch");
+            }
+        }
+
+        private string ExtractPassword(SecureString password) =>
```
sfc-gh-mhofman, line 129: minor - this could have been part of SecureStringHelper class

Snowflake.Data/Core/Session/SFSessionProperty.cs
```diff
@@ -249,7 +250,7 @@ internal static SFSessionProperties ParseConnectionString(string connectionStrin
 
             if (password != null && password.Length > 0)
             {
-                properties[SFSessionProperty.PASSWORD] = new NetworkCredential(string.Empty, password).Password;
+                properties[SFSessionProperty.PASSWORD] = SecureStringHelper.Decode(password);
```
sfc-gh-mhofman, line 253: what worries me that for a security reason this was held encrypted (within SecureString) and now we're taking that option down

Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -119,6 +118,17 @@ private void CleanExpiredSessions()
             }
         }
 
+        internal void ValidateSecurePassword(SecureString password)
+        {
+            if (!ExtractPassword(Password).Equals(ExtractPassword(password)))
+            {
+                throw new Exception("Could not get a pool because of password mismatch");
```
sfc-gh-mhofman, line 125: can you please add logging of such a case

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
```diff
@@ -285,6 +283,71 @@ public void TestGetCurrentPoolSizeReturnsSumOfPoolSizes()
...
+            var password = SecureStringHelper.Decode(_password3);
+            var connectionStringWithPassword = $"{ConnectionStringWithoutPassword}password={password}";
+            var pool = _connectionPoolManager.GetPool(connectionStringWithPassword, _password3);
+
+            // act
+            var serializedPool = pool.ToString();
+
+            // assert
+            Assert.IsFalse(serializedPool.Contains(password));
+        }
```
sfc-gh-mhofman, line 349: should there be TestThrowsWhenPasswordNotGivenInConnectionStringNorSecurePassword ?

sfc-gh-mhofman (changes requested):

Snowflake.Data/Core/Session/SFSessionProperty.cs
sfc-gh-knozderko: this PR we does not change anything. This change is only to replace `new NetworkCredential(string.Empty, password).Password` by function execection with exactly the same code.

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: done

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
sfc-gh-knozderko: SessionPoolTest has such a test, but actually this validation is a kind of a double check. Normally in the pool manager if you search for the pool with wrong password the pool with the wrong password is created and returned.

Snowflake.Data/Core/Session/SessionPool.cs
sfc-gh-knozderko: this converts null to empty string for easy comparison. This is not what we should expect from the general purpose class like SecureStringHelper, so I would rather leave it as is. We could discuss it.

Snowflake.Data.Tests/UnitTests/ConnectionPoolManagerTest.cs
sfc-gh-knozderko: added also tests for password required

PR #985, opened 2024-06-25
example for binding an array as a variable

sfc-gh-dszmolka:
third attempt to get this into the docs ;) 
co-authored by @sfc-gh-hachouraria

### DIFF
--- a/doc/QueryingData.md
+++ b/doc/QueryingData.md
@@ -250,3 +250,37 @@ using (IDbConnection conn = new SnowflakeDbConnection())
 }
 ```
 
+Binding _an array_ As Variable
+------------------------------
+
+Directly binding an array to a variable is not supported currently. Instead, the usual method to pass local arrays as SQL arrays via bind is to use the SQL form `PARSE_JSON(?)`, and then pass a JSON encoded array as string to the variable `?`
+
+Using a stored procedure as an example, which can take an array as an input. Note, you'll need `Newtonsoft.Json` which is already a dependency of the driver.
+```cs
+using Snowflake.Data;
+using Newtonsoft.Json;
+..
+
+                    using (IDbCommand cmd = conn.CreateCommand())
+                    {
+
+                        int[] vals = new int[] { 1, 2, 3 };
+                        string array = JsonConvert.SerializeObject(vals); // alternatively you can do `vals.ToArray()` when passing it to `p1.Value`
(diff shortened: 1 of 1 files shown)


doc/QueryingData.md
```diff
@@ -250,3 +250,37 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+                        p1.Value = array; // passing the array in the bind variable. 
+                        p1.DbType = DbType.String;
+                        cmd.Parameters.Add(p1);
+
+                        IDataReader reader = cmd.ExecuteReader();
+                        while (reader.Read())
+                        {
+                            Console.WriteLine(reader.GetString(0));
+                        }
+                        conn.Close();
```
sfc-gh-pfus, line 284: I'm not a .net expert, but shouldn't the connection be closed by using `using`?

doc/QueryingData.md
```diff
@@ -250,3 +250,37 @@ using (IDbConnection conn = new SnowflakeDbConnection())
...
+using Snowflake.Data;
+using Newtonsoft.Json;
+..
+
+                    using (IDbCommand cmd = conn.CreateCommand())
+                    {
+
+                        int[] vals = new int[] { 1, 2, 3 };
+                        string array = JsonConvert.SerializeObject(vals); // alternatively you can do `vals.ToArray()` when passing it to `p1.Value`
+                        string sql = "CALL test_db.public.test(parse_json(?))"; // test SP, returns a single value
```
sfc-gh-pfus, line 269: I'm wondering how `parse_json` is related to binding structured arrays? After implementing structured arrays, it shouldn't be required anymore. I'm not sure if we should include some temporary solution, whereas it is supposed to be finished this quarter? @sfc-gh-knozderko can you take a look?

sfc-gh-dszmolka: actually it helped at least 2 people i know about, already. i'm trying to get this change in for at least 2 months - happy to hear it is becoming deprecated soon, but until then, maybe it can help even more people ? 
then if its deprecated after structured types support, the relevant doc change which will provide examples for structured type can also deprecate this change once they're in

doc/QueryingData.md
sfc-gh-dszmolka: me neither :) the opening/closing of the connection is fully irrelevant for the scope of this change which tries to demonstrate how one can bind an array which is very non trivial

sfc-gh-dprzybysz (approved):

PR #989, opened 2024-07-05
SNOW 1461701: Add test for Max LOB size parameter switch

sfc-gh-dheyman-1:
### Description
Add test for Max LOB size parameter switch

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
@@ -289,7 +289,6 @@ namespace Snowflake.Data.Tests.IntegrationTests
 
         static IEnumerable<int> LobSizeTestCases = new[]
         {
-            //SmallSize,
             OriginSize,
             MediumSize,
             LargeSize,
--- a/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
@@ -384,10 +384,10 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 command.ExecuteNonQuery();
 
                 //// Alter session max lob
-                // command.CommandText = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY = 'ENABLED'";
-                // command.ExecuteNonQuery();
-                // command.CommandText = "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true";
-                // command.ExecuteNonQuery();
+                //command.CommandText = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY = 'ENABLED'";
+                //command.ExecuteNonQuery();
+                //command.CommandText = "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true";
+                //command.ExecuteNonQuery();
             }
         }
 
--- a/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
@@ -384,8 +384,10 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 command.ExecuteNonQuery();
 
                 //// Alter session max lob
-                //command.CommandText = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY = 'ENABLED'";
-                //command.ExecuteNonQuery();
+                // command.CommandText = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY = 'ENABLED'";
+                // command.ExecuteNonQuery();
+                // command.CommandText = "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true";
+                // command.ExecuteNonQuery();
             }
         }
 
--- a/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
@@ -26,7 +26,6 @@ namespace Snowflake.Data.Tests.IntegrationTests
         private const int LargeSize = (MaxLobSize / 2);
         private const int MediumSize = (LargeSize / 2);
         private const int OriginSize = (MediumSize / 2);
-        private const int SmallSize = 16;
         private const int LobRandomRange = 100000 + 1; // range to use for generating random numbers (0 - 100000)
 
         private static string s_outputDirectory;
@@ -82,7 +81,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
                 using (var command = conn.CreateCommand())
                 {
                     // Create temp table
-                    var columnNamesWithTypes = $"{s_colName[0]} VARCHAR, {s_colName[1]} VARCHAR, {s_colName[2]} INT";
+                    var columnNamesWithTypes = $"{s_colName[0]} VARCHAR({MaxLobSize}), {s_colName[1]} VARCHAR({MaxLobSize}), {s_colName[2]} INT";
                     command.CommandText = $"CREATE OR REPLACE TABLE {t_tableName} ({columnNamesWithTypes})";
                     command.ExecuteNonQuery();
                 }
@@ -273,7 +272,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
             var c1 = GenerateRandomString(lobSize);
             var c2 = GenerateRandomString(lobSize);
             var c3 = new Random().Next(LobRandomRange);
-            t_colData = new string[]{ c1, c2, c3.ToString() };
+            t_colData = new string[] { c1, c2, c3.ToString() };
 
             PrepareTest();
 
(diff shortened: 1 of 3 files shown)


Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+
+                //initial test
+                try
+                {
+                    cmd.CommandText = "select randstr(20000000, random()) as large_str";
+                    cmd.ExecuteReader();
+                }
+                catch (SnowflakeDbException e)
+                {
+                    Assert.That(e.Message, Does.Contain("exceeds supported length"));
```
sfc-gh-knozderko, line 33: what is supposed to happen? Is this exception expected to be thrown? If so, then we could write:

```
var thrown = Assert.Throws<SnowflakeDbException>(() => cmd.ExecuteReader());
                    
Assert.That(thrown.Message, Does.Contain("exceeds supported length"));
```

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+
+                //parameter switch
+                cmd.CommandText = "alter session set ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=true";
+                cmd.ExecuteNonQuery();
+                cmd.CommandText = "select randstr(20000000, random()) as large_str";
+                var reader = cmd.ExecuteReader();
+                Assert.IsTrue(reader.Read());
+
+                //cleanup
+                cmd.CommandText = "alter session unset ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT";
```
sfc-gh-knozderko, line 44: why is this cleanup needed? If you would like this connection not be returned to pool so it could not intefere with other tests you could add `poolingEnabled=false;` to the connection string

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+                cmd.ExecuteNonQuery();
+                cmd.CommandText = "alter session unset ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT";
```
sfc-gh-knozderko, line 46: in both commands you use ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT. Shouldn't it be FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY as well?

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+                //setup
+                cmd.CommandText = "alter session set FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY='ENABLED'";
+                cmd.ExecuteNonQuery();
+                cmd.CommandText = "alter session set ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=false";
+                cmd.ExecuteNonQuery();
+
+                //initial test
+                try
+                {
+                    cmd.CommandText = "select randstr(20000000, random()) as large_str";
```
sfc-gh-knozderko, line 28: this select is repeated twice. We could extract it to a class constant.

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+                    cmd.ExecuteReader();
+                }
+                catch (SnowflakeDbException e)
+                {
+                    Assert.That(e.Message, Does.Contain("exceeds supported length"));
+                }
+
+                //parameter switch
+                cmd.CommandText = "alter session set ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=true";
```
sfc-gh-knozderko, line 37: you could consider if it would be ok to create 2 tests instead. Then you could give better (more meaningful) names for them.

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
```diff
@@ -0,0 +1,51 @@
...
+        [Test]
+        [Ignore("TODO: Enable when Max LOB size is available on the automated tests environment")]
+        public void TestIncreaseMaxLobSizeParameterSwitchSelect()
+        {
+            using (var conn = new SnowflakeDbConnection(ConnectionString))
+            {
+                conn.Open();
+                IDbCommand cmd = conn.CreateCommand();
+
+                //setup
```
sfc-gh-knozderko, line 19: maybe extract it into a method since the comment is needed for a block of code?

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
```diff
@@ -388,6 +386,8 @@ private void AlterSessionSettings(SnowflakeDbConnection conn)
                 //// Alter session max lob
                 //command.CommandText = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY = 'ENABLED'";
                 //command.ExecuteNonQuery();
+                //command.CommandText = "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true";
```
sfc-gh-knozderko, line 389: why is it commented? It would be uncommented later?

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
```diff
@@ -290,7 +289,6 @@ public void TestPutGetCommand(int lobSize)
 
         static IEnumerable<int> LobSizeTestCases = new[]
         {
-            SmallSize,
```
sfc-gh-knozderko, line 293: why did you remove this test case? Isn't it valid any more?

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
```diff
@@ -82,7 +81,7 @@ public void SetUp()
                 using (var command = conn.CreateCommand())
                 {
                     // Create temp table
-                    var columnNamesWithTypes = $"{s_colName[0]} VARCHAR, {s_colName[1]} VARCHAR, {s_colName[2]} INT";
+                    var columnNamesWithTypes = $"{s_colName[0]} VARCHAR({MaxLobSize}), {s_colName[1]} VARCHAR({MaxLobSize}), {s_colName[2]} INT";
```
sfc-gh-knozderko, line 84: why do you add size of the varchar? Is it needed for a reason?

sfc-gh-knozderko (changes requested):

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
sfc-gh-dheyman-1: No, it's not valid since this test should test different max lob sizes. Also it makes the test flaky because of the behaviour of Snowflake (for small size) when max lob size is increased.

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
sfc-gh-dheyman-1: Yes, without it the size would be default (which differs for different max lob sizes and makes the test flaky)

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
sfc-gh-dheyman-1: I added this in case if someone would like to test this for increased max LOB size. Without this change it won't work. Same goes for existing comment above.

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: I don't see much value here, would decrease readability.

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: This is enabled by default so theres no need.

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: done

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: IMHO there's no need, since after your suggestions the test is much shorter

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: Thanks for suggestion, done

Snowflake.Data.Tests/IntegrationTests/SFMaxLobSizeSwitchIT.cs
sfc-gh-dheyman-1: Yes, that's much cleaner. Done.

sfc-gh-knozderko (approved):

PR #997, opened 2024-07-29
SNOW-1333163 support for default proxy

sfc-gh-knozderko:
### Description
SNOW-1333163 Add support for default proxy.

When specifying `allowEmptyProxy=true` and `useProxy=true` but without specifying the value of `proxyHost`, the proxy would be enabled but not configured allowing to use a default proxy provided by the operating system.

### Checklist
- [x] Code compiles correctly


vc-74:

Snowflake.Data.Tests/UnitTests/SFSessionPropertyTest.cs
```diff
@@ -166,6 +166,19 @@ public void TestResolveConnectionArea(string host, string expectedMessage)
             Assert.AreEqual(expectedMessage, message);
         }
 
+        [Test]
+        [TestCase("ACCOUNT=account;USER=test;PASSWORD=test;ALLOWEMPTYPROXY=false", "false")]
+        [TestCase("ACCOUNT=account;USER=test;PASSWORD=test;ALLOWEMPTYPROXY=true", "true")]
```
sfc-gh-dstempniak, line 171: Can you add a test with `UseProxy=true`.
It would be nice to test this code: https://github.com/snowflakedb/snowflake-connector-net/blob/982a1e400f8fd67cff4d8dedb118ddfea098f328/Snowflake.Data/Core/Session/SFSessionProperty.cs#L242

sfc-gh-dstempniak: I think we don't have tests for connection string with UseProxy=true and without proxyHost. I believe that ProxyHost is required when UseProxy is true. If so, we could remove ALLOWEMPTYPROXY parameter, It won't be BCR because nobody use UseProxy without proxyHost specified.

sfc-gh-knozderko: I removed the parameter `allowEmptyProxy`. I fixed the validation for proxy parameters which did not work.

sfc-gh-dstempniak (approved):
LGTM

PR #1013, opened 2024-08-29
SNOW-1502317: Stop retry on nonrecoverable exception

sfc-gh-ext-simba-lf:
### Description
Stop retry on nonrecoverable exception

### Checklist
- [ ] Code compiles correctly


Snowflake.Data.Tests/Mock/MockInfiniteTimeout.cs
```diff
@@ -0,0 +1,45 @@
...
+                base._HttpClient = httpClient;
+            }
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage message,
+                                                              TimeSpan restTimeout,
+                                                              CancellationToken externalCancellationToken,
+                                                              string sid = "")
+        {
+            message.Properties[BaseRestRequest.HTTP_REQUEST_TIMEOUT_KEY] = Timeout.InfiniteTimeSpan;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -581,6 +584,50 @@ public void TestEnableLoginRetryOn404()
             }
         }
 
+
+        [Test]
+        public void TestNonRetryableHttpExceptionThrowsError()
```
sfc-gh-knozderko, line 589: it looks like an unit test not an integration test. You could move it to unit tests.

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -404,8 +411,19 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                         }
                         else
                         {
-                            //TODO: Should probably check to see if the error is recoverable or transient.
-                            logger.Warn("Error occurred during request, retrying...", e);
+                            Exception innermostException = e;
+                            while (innermostException.InnerException != null) innermostException = innermostException.InnerException;
+
+                            if (innermostException is AuthenticationException)
```
sfc-gh-knozderko, line 417: how is that compared to jdbc driver? Are there in jdbc driver any kinds of exceptions not retryable as well?

sfc-gh-knozderko (changes requested):

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -404,8 +411,19 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                         }
                         else
                         {
-                            //TODO: Should probably check to see if the error is recoverable or transient.
-                            logger.Warn("Error occurred during request, retrying...", e);
+                            Exception innermostException = e;
+                            while (innermostException.InnerException != null) innermostException = innermostException.InnerException;
```
sfc-gh-knozderko, line 415: I'm just thinking if we could fall into an infinitive loop if someone would throw an exception with itself assigned to inner exception. I know it would be weird but maybe we could add a condition for that in the while loop

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: JDBC does not retry if the exception is due to a [revoked certificate](https://github.com/snowflakedb/snowflake-jdbc/blob/cc073f8917b91dd0855985c3085b99c73fe066d5/src/main/java/net/snowflake/client/jdbc/RestRequest.java#L324)

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: Added check for that case

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -404,8 +411,20 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                         }
                         else
                         {
-                            //TODO: Should probably check to see if the error is recoverable or transient.
-                            logger.Warn("Error occurred during request, retrying...", e);
+                            Exception innermostException = e;
+                            while (innermostException.InnerException != null && innermostException != innermostException.InnerException)
```
sfc-gh-knozderko, line 415: maybe we could extract this logic to a method called GetInnerMostException()

sfc-gh-knozderko (approved):

Snowflake.Data/Core/HttpUtil.cs
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -517,6 +515,13 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
             }
         }
 
+        static private Exception GetInnerMostException(Exception innermostException)
```
sfc-gh-knozderko, line 518: I know that it works, but someone who reads it could be confused to see innermostException as a parameter name because it should be the result, not the input parameter.
Maybe we could rename the parameter `GetInnerMostException(Exception exception)` and then start the method with `var innermostException = exception;`

sfc-gh-ext-simba-lf: Done

PR #1022, opened 2024-09-06
SNOW-1640968 chunk downloader fix

sfc-gh-dstempniak:
### Description
Clearing memory after chunk download.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Core/SFReusableChunk.cs
+++ b/Snowflake.Data/Core/SFReusableChunk.cs
@@ -11,7 +11,7 @@ namespace Snowflake.Data.Core
     class SFReusableChunk : BaseResultChunk
     {
         internal override ResultFormat ResultFormat => ResultFormat.JSON;
-        
+
         private readonly BlockResultData data;
 
         private int _currentRowIndex = -1;
@@ -29,11 +29,18 @@ namespace Snowflake.Data.Core
             data.Reset(RowCount, ColumnCount, chunkInfo.uncompressedSize);
         }
 
+        internal override void Reset()
+        {
+            base.Reset();
+            _currentRowIndex = -1;
+            data.Reset();
+        }
+
         internal override void ResetForRetry()
         {
             data.ResetForRetry();
         }
-        
+
         [Obsolete("ExtractCell with rowIndex is deprecated", false)]
         public override UTF8Buffer ExtractCell(int rowIndex, int columnIndex)
         {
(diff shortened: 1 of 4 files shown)


Snowflake.Data/Core/SFReusableChunk.cs
```diff
@@ -74,10 +81,10 @@ private class BlockResultData
             private static readonly int NULL_VALUE = -100;
             private int blockCount;
 
-            private static int blockLengthBits = 24;
+            private static int blockLengthBits = 23;
```
sfc-gh-knozderko, line 84: why do we change this?

Snowflake.Data/Core/SFReusableChunk.cs
```diff
@@ -74,10 +81,10 @@ private class BlockResultData
             private static readonly int NULL_VALUE = -100;
             private int blockCount;
 
-            private static int blockLengthBits = 24;
+            private static int blockLengthBits = 23;
             private static int blockLength = 1 << blockLengthBits;
             int metaBlockCount;
-            private static int metaBlockLengthBits = 15;
+            private static int metaBlockLengthBits = 17;
```
sfc-gh-knozderko, line 87: why do we change this?

Snowflake.Data/Core/SFReusableChunk.cs
sfc-gh-dstempniak: To have two times smaller blocks. It is in par with jdbc: https://github.com/snowflakedb/snowflake-jdbc/blob/d26df3a367fbbde9082372e1d6f4a1c168ed31b9/src/main/java/net/snowflake/client/jdbc/JsonResultChunk.java#L449

Snowflake.Data/Core/SFReusableChunk.cs
sfc-gh-dstempniak: I did some tests and we have hundreds of meta blocks. By increasing the size of a block we reduce the number of blocks. But in JDBC there is 15, so I can revert it

sfc-gh-pbulawa (approved):

Snowflake.Data/Core/SFReusableChunk.cs
sfc-gh-knozderko: Ok, let's revert it.

sfc-gh-knozderko (approved):

PR #1042, opened 2024-10-16
SNOW-1739483 Fix S3 exception casting and flaky tests

sfc-gh-dstempniak:
### Description
Fix S3 exception casting and flaky tests
We sometimes get `System.InvalidCastException : Unable to cast object of type 'System.Net.Sockets.SocketException' to type 'Amazon.S3.AmazonS3Exception'. `

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
+++ b/Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
@@ -206,7 +206,7 @@ namespace Snowflake.Data.Core.FileTransfer.StorageClient
             }
             catch (Exception ex)
             {
-                fileMetadata = HandleFileHeaderErr(ex.InnerException, fileMetadata); // S3 places the AmazonS3Exception on the InnerException on non-async calls
+                HandleFileHeaderErr(ex.InnerException, fileMetadata); // S3 places the AmazonS3Exception on the InnerException on non-async calls
                 return null;
             }
         }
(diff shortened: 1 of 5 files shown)


sfc-gh-pbulawa (approved):

Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
```diff
@@ -519,71 +519,89 @@ private GetObjectRequest GetGetObjectRequest(ref AmazonS3Client client, SFFileMe
         /// </summary>
         /// <param name="ex">Exception from file header.</param>
         /// <param name="fileMetadata">The file metadata.</param>
-        /// <returns>The file metadata.</returns>
-        private SFFileMetadata HandleFileHeaderErr(Exception ex, SFFileMetadata fileMetadata)
+        private void HandleFileHeaderErr(Exception ex, SFFileMetadata fileMetadata)
```
sfc-gh-knozderko, line 522: for me returning fileMetadata has some sense because it emphasizes that fileMetadata parameter was modified.
But we can return void if you don't like the previous approach.

sfc-gh-dstempniak: I think it is confusing when we return SFFileMetadata and do nothing with the returned value. It seems to me that the object passed as a parameter should be immutable if we return a modified copy of it.

PR #1057, opened 2024-11-09
SNOW-834781: Remove log4net and delegate logging to consumers

sfc-gh-ext-simba-lf:
### Description
Remove the log4net dependency and replace with the Microsoft logging interface so users can plug in their own custom loggers

### Checklist
- [ ] Code compiles correctly


Snowflake.Data/Logger/SFLoggerPair.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Text;
+
+namespace Snowflake.Data.Log
+{
+    public class SFLoggerPair
+    {
+        private static SFLogger s_snowflakeLogger;
+        private static ILogger s_customLogger;
+
+        SFLoggerPair(SFLogger snowflakLogger, ILogger customLogger)
```
sfc-gh-knozderko, line 16: typo in snowflake

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -22,13 +22,13 @@ public class SnowflakeDbCommand : DbCommand
 
         private SnowflakeDbParameterCollection parameterCollection;
 
-        private SFLogger logger = SFLoggerFactory.GetLogger<SnowflakeDbCommand>();
+        private SFLoggerPair _loggerPair = SFLoggerPair.GetLoggerPair<SnowflakeDbCommand>();
```
sfc-gh-knozderko, line 25: would it be possible not to be forced to change this in all places in which we log anything?
Maybe SFLoggerPair could implement SFLogger interface?
Maybe SFLoggerFactory.GetLogger could return the same as SFLoggerPair.GetLoggerPair?
Maybe SFLoggerPair could do LogDebug -> Debug translations?
Maybe SFLoggerPair could mask secrets before passing them further?

Currently because of this 80 files are changed so it is even difficult to review.

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -55,6 +80,30 @@ public static SFLogger GetLogger<T>()
...
+        {
+            // If true, return the default/specified logger
+            if (s_isCustomLoggerEnabled)
+            {
+                // If no logger specified, use the default logger: Microsoft's console logger
+                if (s_customLogger == null)
+                {
+                    ILoggerFactory factory = LoggerFactory.Create(
+                        builder => builder
+                        .AddConsole()
```
sfc-gh-knozderko, line 94: does it mean that we create a custom logger with logging to console always enabled?
Could we defer the decision about appenders to the customer configuration?

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,181 @@
...
+        return SFLogRepository.s_rootLogger == this ?
+            _isFatalEnabled :
+            SFLogRepository.s_rootLogger.IsFatalEnabled();
+    }
+
+    public void Debug(string msg, Exception ex = null)
+    {
+        if (IsDebugEnabled())
+        {
+            msg = SecretDetector.MaskSecrets(msg).maskedText;
```
sfc-gh-knozderko, line 130: This way secrets would be masked only in SFLoggerImpl. We need to mask secrets in customer implementation as well.

sfc-gh-knozderko (changes requested):

Snowflake.Data/Logger/SFLoggerFactory.cs
sfc-gh-ext-simba-lf: This means if a customer enables custom logging without setting their own logger then a console logger will be created

If a customer sets the logger then it will return their custom logger with their chosen appenders (logging to console is not enabled unless they added it themselves)

Snowflake.Data/Logger/SFLoggerImpl.cs
sfc-gh-ext-simba-lf: Moved the masking so both loggers use the masked message

Snowflake.Data/Logger/SFLoggerPair.cs
sfc-gh-ext-simba-lf: Fixed

Snowflake.Data/Client/SnowflakeDbCommand.cs
sfc-gh-ext-simba-lf: Made the following changes:
- SFLoggerPair implements SFLogger
- Replaced SFLoggerPair.GetLoggerPair with SFLoggerFactory.GetLogger
- Renamed LogDebug to Debug and other Log functions
- Masked secrets before passing to loggers

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,51 +1,81 @@
...
 namespace Snowflake.Data.Log
 {
-    class SFLoggerFactory
+    public class SFLoggerFactory
     {
-        private static bool isLoggerEnabled = true;
+        private static bool s_isCustomLoggerEnabled = false;
 
-        private static SFLogger logger = null;
+        private static bool s_isSFLoggerEnabled = true;
```
sfc-gh-knozderko, line 14: shouldn't it be disabled by default? I thought Easy Logging feature would enable it?

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,51 +1,81 @@
...
+            s_isSFLoggerEnabled = true;
+        }
+
+        public static void DisableCustomLogger()
         {
+            s_isCustomLoggerEnabled = false;
         }
 
-        public static void disableLogger()
+        public static void EnableCustomLogger()
```
sfc-gh-knozderko, line 33: Let's say that that at the beginning customer logger was disabled.
Let's say that some class has a static logger eq. `private static readonly SFLogger logger = SFLoggerFactory.GetLogger<BasicAuthenticator>();`
then the logger would have a value of `new SFLoggerPair(..., ILoggerEmptyImpl();)`
Let's say then the customer would call `EnableCustomLogger()`.
The logger obtained in the first line would not be modified.
Am I right?
I think we need to reconfigure all the logger when `EnableCustomLogger()`, switching this field value is not enough.

Snowflake.Data/Logger/SFLoggerPair.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            s_snowflakeLogger = snowflakeLogger;
+            s_customLogger = customLogger;
+        }
+
+        public void Debug(string message, Exception ex = null)
+        {
+            message = SecretDetector.MaskSecrets(message).maskedText;
+            s_snowflakeLogger.Debug(message, ex);
+            s_customLogger.LogDebug(FormatBrackets(message), ex);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It was because only Debug has a log with brackets. It came from this [query](https://github.com/snowflakedb/snowflake-connector-net/blob/c45f1ad161dfa1c1fd0ce4e513dd17c186d9f9f7/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs#L320) which led to this [error](https://github.com/snowflakedb/snowflake-connector-net/actions/runs/12172539890/job/33954179866?pr=1057#step:8:542)

Edited to include FormatBrackets for other logging

Snowflake.Data/Logger/SFLoggerFactory.cs
sfc-gh-ext-simba-lf: That's correct

`s_isSFLoggerEnabled` is now set to false and the SF logger is disabled by default

Snowflake.Data/Logger/SFLoggerFactory.cs
sfc-gh-ext-simba-lf: That's correct, it would not register the new custom logger if the logger pair has already been created before

Edited the `SetCustomLogger` function to update the custom logger in `SFLoggerPair` so the custom logger will be updated when the user sets a new one

sfc-gh-ext-simba-lf:
> Any motion on this PR? I am finding log4net to be impossible to work with in a modern .Net application.

It's currently undergoing further review

Snowflake.Data/Logger/EasyLoggerManager.cs
```diff
@@ -75,67 +67,60 @@ private static void RemoveOtherEasyLoggingAppenders(log4net.Repository.Hierarchy
...
+                _maximumFileSizeInBytes = 1000000000, // "1GB"
+                _maxSizeRollBackups = 2,
             };
             appender.ActivateOptions();
             logger.AddAppender(appender);
             return appender;
         }
 
-        private static bool IsEasyLoggingAppender(IAppender appender)
+        private static bool IsEasyLoggingAppender(SFAppender appender)
```
sfc-gh-pbulawa, line 89: What about using a switch statement like:
```c#
switch (appender)
{
    case SFConsoleAppender a:
        return a._name != null && a._name.StartsWith(AppenderPrefix);
    case SFRollingFileAppender a:
        return a._name != null && a._name.StartsWith(AppenderPrefix);
    default:
        return false;
}
```
or even refactor `SFAppender` interface to contain a `Name` property and generalize it to:
```c#
switch (appender)
{
    case SFAppender a:
        return a.Name != null && a.Name.StartsWith(AppenderPrefix);
    default:
        return false;
}
```

sfc-gh-pbulawa (changes requested):

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-ext-simba-lf: Replaced with switch statement and added `Name` property

sfc-gh-ext-simba-lf: Ok. I replaced the variable on the switch case with `_` but let me know if you prefer another way

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
```diff
@@ -13,15 +13,15 @@ namespace Snowflake.Data.Configuration
 {
     internal class EasyLoggingConfigFinder
     {
-        private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<EasyLoggingConfigFinder>();
-        
+        private static readonly SFLogger s_logger = SFLoggerFactory.GetSFLogger<EasyLoggingConfigFinder>();
```
sfc-gh-knozderko, line 16: why not GetLogger?

Snowflake.Data/Configuration/EasyLoggingConfigParser.cs
```diff
@@ -15,7 +15,7 @@ namespace Snowflake.Data.Configuration
 {
     internal class EasyLoggingConfigParser
     {
-        private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<EasyLoggingConfigParser>();
+        private static readonly SFLogger s_logger = SFLoggerFactory.GetSFLogger<EasyLoggingConfigParser>();
```
sfc-gh-knozderko, line 18: why not GetLogger?

Snowflake.Data/Configuration/EasyLoggingLogLevel.cs
```diff
@@ -6,9 +6,10 @@
 
 namespace Snowflake.Data.Configuration
 {
-    internal enum EasyLoggingLogLevel
+    public enum EasyLoggingLogLevel
```
sfc-gh-knozderko, line 9: does it need to be public?

Snowflake.Data/Core/Session/EasyLoggingStarter.cs
```diff
@@ -14,7 +14,7 @@ namespace Snowflake.Data.Core
 {
     internal class EasyLoggingStarter
     {
-        private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<EasyLoggingStarter>();
+        private static readonly SFLogger s_logger = SFLoggerFactory.GetSFLogger<EasyLoggingStarter>();
```
sfc-gh-knozderko, line 17: why not GetLogger?

sfc-gh-knozderko (changes requested):

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,51 +1,88 @@
...
+                    logger.SetLevel(LoggingEvent.OFF); // Logger is disabled by default and can be enabled by the EasyLogging feature
+                }
+                if (useFileAppender)
                 {
-                    ILog loggerL = LogManager.GetLogger(typeof(T));
-                    return new Log4NetImpl(loggerL);
+                    var fileAppender = new SFRollingFileAppender()
+                    {
+                        _name = "RollingFileAppender",
+                        _logFilePath = Path.Combine(Directory.GetCurrentDirectory(), "test_snowflake_log.log"),
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: My mistake, this should've been removed since this appender would not have been used and end up being replaced by the one that easy logging creates

Removed since easy logging already appends a logger with the right configuration

Snowflake.Data/Logger/SFLogger.cs
```diff
@@ -1,39 +1,45 @@
...
+/*
+ * Copyright (c) 2012-2019 Snowflake Computing Inc. All rights reserved.
+ */
+
+using System;
+using System.Collections.Generic;
+
+namespace Snowflake.Data.Log
+{
+    public interface SFLogger
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: No, this was from testing but should've been changed to internal

Changed to internal

Snowflake.Data/Logger/SFLoggerPair.cs
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Snowflake.Data.Log
+{
+    public class SFLoggerPair : SFLogger
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/SFAppender.cs
```diff
@@ -0,0 +1,15 @@
+/*
+ //* Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+using System;
+
+namespace Snowflake.Data.Log
+{
+    public interface SFAppender
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,51 +1,88 @@
...
  */
 
-using log4net;
+using Microsoft.Extensions.Logging;
+using System.IO;
 
 namespace Snowflake.Data.Log
 {
-    class SFLoggerFactory
+    public class SFLoggerFactory
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+using Snowflake.Data.Log;
+using System;
+using System.Collections.Generic;
+
+public static class SFLogRepository
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,176 @@
...
+{
+    internal static SFLogger s_rootLogger = new SFLoggerImpl(typeof(SFLogRepository));
+
+    internal static SFLogger GetRootLogger()
+    {
+        return s_rootLogger;
+    }
+}
+
+public class SFLoggerImpl : SFLogger
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/ILoggerEmptyImpl.cs
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2012-2019 Snowflake Computing Inc. All rights reserved.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+using Snowflake.Data.Log;
+using System;
+using System.Collections.Generic;
+
+public static class SFLogRepository
+{
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Logger/SFAppenderImpl.cs
```diff
@@ -0,0 +1,126 @@
...
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+using Snowflake.Data.Log;
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+internal class PatternLayout
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -22,10 +22,12 @@
     <PackageReference Include="Google.Cloud.Storage.V1" Version="4.10.0" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.13.0" />
     <PackageReference Include="Azure.Storage.Common" Version="12.12.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
     <PackageReference Include="Mono.Unix" Version="7.1.0-final.1.21458.1" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="log4net" Version="2.0.12" />
     <PackageReference Include="BouncyCastle.Cryptography" Version="2.3.1" />
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.5.0" />
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It's not necessary, it was originally added to replace the log4net package but the connector has its own implementation now

Removed

Snowflake.Data/Configuration/EasyLoggingLogLevel.cs
sfc-gh-ext-simba-lf: No, changed to internal now

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
sfc-gh-ext-simba-lf: I might've missed this when replacing the other GetSFLogger calls

Replaced with GetLogger

Snowflake.Data/Configuration/EasyLoggingConfigParser.cs
sfc-gh-ext-simba-lf: Replaced with GetLogger

Snowflake.Data/Core/Session/EasyLoggingStarter.cs
sfc-gh-ext-simba-lf: Replaced with GetLogger

Snowflake.Data/Logger/EasyLoggerManager.cs
```diff
@@ -75,67 +67,54 @@ private static void RemoveOtherEasyLoggingAppenders(log4net.Repository.Hierarchy
...
         {
-            if (appender.GetType() == typeof(ConsoleAppender))
+            switch (appender)
```
sfc-gh-knozderko, line 91: I don't understand the concept.
SFLoggerImpl is a class which is configured only by easy logging feature, right?
So all of its appenders are configured by easy logging feature, right?
So this method should return always true, so it is not needed anymore, right?

Snowflake.Data/Logger/EasyLoggerManager.cs
```diff
@@ -45,27 +40,24 @@ internal static bool IsStdout(string logsPath)
...
         internal void ResetEasyLogging(EasyLoggingLogLevel easyLoggingLogLevel)
         {
-            var log4netLevel = _levelMapper.ToLog4NetLevel(easyLoggingLogLevel);
+            var sfLoggerLevel = _levelMapper.ToLoggingEventLevel(easyLoggingLogLevel);
             lock (_lockForExclusiveConfigure)
             {
-                var repository = (log4net.Repository.Hierarchy.Hierarchy)LogManager.GetRepository();
-                var rootLogger = (log4net.Repository.Hierarchy.Logger)repository.GetLogger("Snowflake.Data");
-                rootLogger.Level = log4netLevel;
+                var rootLogger = SFLogRepository.GetRootLogger();
```
sfc-gh-knozderko, line 46: I've lost. Could you explain the relation between RootLogger and SFLoggerPair.s_snowflakeLogger?
Why do we need root logger?

sfc-gh-ext-simba-lf: The `SFLoggerPair.s_snowflakeLogger` contains the type/class where the logging is being created (`SnowflakeDbConnection`, `SnowflakeDbDataReader`, etc.)

The `rootLogger` contains the current log level set from `ReconfigureEasyLogging` or `ResetEasyLogging`

If there's no `rootLogger`, then calling `ReconfigureEasyLogging` would not affect the log level of `SFLoggerPair.s_snowflakeLogger` and it would just log at the level that it was initialized

Since the original log4net used a `rootLogger` for the log level, I tried to minimize any changes by making SFlogger similar

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-ext-simba-lf: That's correct. This method could be removed but it was [another reviewer who requested to turn this into a switch statement](https://github.com/snowflakedb/snowflake-connector-net/pull/1057#discussion_r1939532531)

Is it ok for me to remove this function or would you prefer checking first with the reviewer who requested this change?

Snowflake.Data/Configuration/SFConfigurationSectionHandler.cs
```diff
@@ -2,29 +2,18 @@
...
-
 namespace Snowflake.Data.Configuration
 {
     /// <summary>
     ///     Develop In progress
     /// </summary>
-    public class SFConfigurationSectionHandler : IConfigurationSectionHandler
+    public class SFConfigurationSectionHandler
     {
         static SFConfigurationSectionHandler()
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: From my findings, this class is not being used and could be removed

I only added changes in it because it started failing when the log4net package was removed. I don't know much about it other than it was made around [6 years ago](https://github.com/snowflakedb/snowflake-connector-net/blob/7258bda76923af03b9c0dffb876d6b4ad1a79958/Snowflake.Data/Configuration/SFConfigurationSectionHandler.cs#L14) but has not been used since then

Snowflake.Data/Configuration/EasyLoggingLogLevel.cs
```diff
@@ -9,6 +9,7 @@ namespace Snowflake.Data.Configuration
     internal enum EasyLoggingLogLevel
     {
         Off,
+        Fatal,
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed `Fatal` log level

Snowflake.Data/Configuration/EasyLoggingConfigFinder.cs
```diff
@@ -14,14 +14,14 @@ namespace Snowflake.Data.Configuration
     internal class EasyLoggingConfigFinder
     {
         private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<EasyLoggingConfigFinder>();
-        
+
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

I reverted these files but for some reason `SFBlockingChunkDownloaderV3.cs` is still showing up on the "Files changed" list. I checked through a couple diff checkers and couldn't find anything. I also tried copy-pasting the original and keeping the formatting but it's still showing up on the list

Snowflake.Data/Logger/SFLoggerPair.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace Snowflake.Data.Log
+{
+    internal class SFLoggerPair : SFLogger
+    {
+        private static SFLogger s_snowflakeLogger;
+        internal static ILogger s_customLogger;
+
+        public SFLoggerPair(SFLogger snowflakeLogger, ILogger customLogger)
+        {
+            s_snowflakeLogger = snowflakeLogger;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Modified the parameters

- `s_snowflakeLogger` is changed from static to readonly since it did not need to be static
- `s_customLogger` is replaced by directly using the custom logger set in `SFLoggerFactory`

Snowflake.Data/Logger/ILoggerEmptyImpl.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Snowflake.Data.Log
+{
+    // Empty implementation of SFLogger
+    // Used when SFLoggerFactory.disableLogger() is called.
+
+    class ILoggerEmptyImpl : ILogger
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Renamed

Snowflake.Data/Logger/ILoggerEmptyImpl.cs
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) 2012-2025 Snowflake Computing Inc. All rights reserved.
+ */
+
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Snowflake.Data.Log
+{
+    // Empty implementation of SFLogger
+    // Used when SFLoggerFactory.disableLogger() is called.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,166 @@
...
+        var enabled = _level != LoggingEvent.OFF;
+        _isDebugEnabled = enabled;
+        _isInfoEnabled = enabled;
+        _isWarnEnabled = enabled;
+        _isErrorEnabled = enabled;
+        _isFatalEnabled = enabled;
+
+        if (enabled)
+        {
+            switch (_level)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added unit test

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                case LoggingEvent.TRACE:
+                case LoggingEvent.DEBUG:
+                    break;
+                case LoggingEvent.FATAL:
+                    _isErrorEnabled = false;
+                    goto case LoggingEvent.ERROR;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,51 +1,75 @@
...
-        public static void Instance(SFLogger customLogger)
-        {            
-            logger = customLogger;
+        public static void UseDefaultLogger()
+        {
+            s_customLogger = null;
         }
 
-        public static SFLogger GetLogger<T>()
+        public static void SetCustomLogger(ILogger customLogger)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Understood, the public functions intended for the user are moved to a public class in the client directory

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-ext-simba-lf: I've refactored it so the log level/appender are static and root logger is no longer necessary. But if you prefer the original way that log4net used to do it let me know

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+using Snowflake.Data.Log;
+using System;
+using System.Collections.Generic;
+
```
sfc-gh-knozderko, line 8: why no namespace?

Snowflake.Data/Logger/SFLoggerPair.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: Now `customLogger` is unused.

Snowflake.Data/Configuration/SFConfigurationSectionHandler.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: I would remove this class

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-knozderko: I think that comment was only about the style of coding not the whole sense of this logic. If this logic does not make any sense I would remove it.

Snowflake.Data/Logger/SFLoggerPair.cs
```diff
@@ -0,0 +1,99 @@
...
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Snowflake.Data.Log
+{
+    internal class SFLoggerPair : SFLogger
+    {
+        private readonly SFLogger s_snowflakeLogger;
```
sfc-gh-knozderko, line 14: it is not static but has prefix s_

Snowflake.Data/Logger/SFLoggerImpl.cs
sfc-gh-ext-simba-lf: I overlooked it. A namespace is now added

Snowflake.Data/Logger/SFLoggerPair.cs
sfc-gh-ext-simba-lf: Renamed

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-knozderko: the whole method was removed so I'm resolving this conversation.

Snowflake.Data/Logger/EasyLoggerManager.cs
sfc-gh-knozderko: I would like to understand the big picture.
1. How the user can register his logger implementation?
- I see that he needs to set SFLoggerFactory.s_customLogger by SnowflakeDbLoggerFactory.SetCustomLogger()
- so the consequence is that the user sets for all drivers's classes the same implementation - is it good?
- I know that for instance in log4net you could define different loggers for different classes
- Is there a way to set different customer loggers for different classes?
- Is there a generic way to register ILogger implementation?
2. How does Snowflake logger work?
- I see that SFLoggerPair has a field of _snowflakeLogger
- Once the SFLoggerPair is created, I don't see the way to change the snowflakeLogger
- I see that EasyLogging changes appenders for SFLoggerImpl
- but how Easy Logging would enable logging if the classes would have SFLoggerEmptyImpl implementation in their SFLoggerPair?

sfc-gh-ext-simba-lf: 1.
> I see that he needs to set SFLoggerFactory.s_customLogger by SnowflakeDbLoggerFactory.SetCustomLogger()
- That's correct
> so the consequence is that the user sets for all drivers's classes the same implementation - is it good?
- Yes, it sets the same implementation for all classes
> I know that for instance in log4net you could define different loggers for different classes
- Yes, when `log4net` was used directly by the connector it defined different loggers for different classes using `SFLoggerFactory.GetLogger<Class>()`. Now the connector does a similar thing for `SFLogger`
> Is there a way to set different customer loggers for different classes?
- There probably is, maybe a function that adds a custom logger per class. The connector could check a dictionary of custom loggers for that specific class and if not, use the custom logger from `SetCustomLogger`
> Is there a generic way to register ILogger implementation?
- This was the doc I found but it seems to be specific to HostApplicationBuilder: https://learn.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider#usage-and-registration-of-the-custom-logger


2.
> I see that SFLoggerPair has a field of _snowflakeLogger
> Once the SFLoggerPair is created, I don't see the way to change the snowflakeLogger
- It's changed through the appender
> I see that EasyLogging changes appenders for SFLoggerImpl
- That's correct
> but how Easy Logging would enable logging if the classes would have SFLoggerEmptyImpl implementation in their SFLoggerPair?
- The `SFLoggerEmptyImpl` was done for test purposes and can only be enabled internally. The `SFLoggerImpl` should be what `SFLoggerPair` always uses

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+                    if (ex != null)
+                    {
+                        writer.WriteLine(ex.Message);
+                    }
+                }
+            }
+        }
+        catch (Exception logEx)
+        {
+            Console.WriteLine($"Unable to log the following message:\n{formattedMessage}" +
```
sfc-gh-knozderko, line 46: So if we would not have permissions to write to a file, then we we would write everything on output?
I think we could use Console.Error.WriteLine to write only about the error on writing to a file, but not the error message itself.

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        var formattedMessage = _patternLayout.Format(logLevel, message, type);
+        try
+        {
+            if (LogFileIsTooLarge())
+            {
+                RollLogFile();
+            }
+
+            using (FileStream fs = new FileStream(_logFilePath, FileMode.Append, FileAccess.Write, FileShare.ReadWrite))
```
sfc-gh-knozderko, line 32: this way you don't control the log file rights for Linux and Mac.

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+                $"Due to the error: {logEx.Message}\n");
+        }
+    }
+
+    public void ActivateOptions()
+    {
+        var logDir = Path.GetDirectoryName(_logFilePath);
+        if (!Directory.Exists(logDir))
+        {
+            Directory.CreateDirectory(logDir);
```
sfc-gh-knozderko, line 56: this way you don't control permissions on Linux and Mac

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        if (!File.Exists(_logFilePath))
+        {
+            File.Create(_logFilePath).Dispose();
```
sfc-gh-knozderko, line 60: this way you don't control permissions on Linux and Mac

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
+/*
```
sfc-gh-knozderko, line 1: the header is not needed

Snowflake.Data/Logger/SFConsoleAppender.cs
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2025 Snowflake Computing Inc. All rights reserved.
```
sfc-gh-knozderko, line 2: the header is not needed

Snowflake.Data/Logger/LoggerEmptyImpl.cs
```diff
@@ -0,0 +1,28 @@
+/*
```
sfc-gh-knozderko, line 1: header is not needed

Snowflake.Data/Logger/PatternLayout.cs
```diff
@@ -0,0 +1,26 @@
+/*
```
sfc-gh-knozderko, line 1: header is not needed

Snowflake.Data/Logger/SFConsoleAppender.cs
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+using Snowflake.Data.Log;
+using System;
+
+internal class SFConsoleAppender : SFAppender
+{
+    internal string _name;
```
sfc-gh-knozderko, line 10: why the console appender has a name? When we had log4net implementation it was to distinguish between our appenders and customer appenders. Currently every object of this class is created by us so the name is not needed anymore.

Snowflake.Data/Logger/SFConsoleAppender.cs
```diff
@@ -0,0 +1,26 @@
...
+    internal PatternLayout _patternLayout;
```
sfc-gh-knozderko, line 11: let's change it to property

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+using Snowflake.Data.Log;
+using System;
+using System.IO;
+using System.Linq;
+
+internal class SFRollingFileAppender : SFAppender
+{
+    internal string _name;
```
sfc-gh-knozderko, line 12: why the appender has a name? When we had log4net implementation it was to distinguish between our appenders and customer appenders. Currently every object of this class is created by us so the name is not needed anymore.

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+    internal string _logFilePath;
+    internal long _maximumFileSizeInBytes;
```
sfc-gh-knozderko, line 14: I see that you set all these fields in when you create this appender objects. So maybe change these fields to properties. If you just set them and don't need to get them you can create only setters for them.

Snowflake.Data/Logger/PatternLayout.cs
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2025 Snowflake Computing Inc. All rights reserved.
+ */
+
+using System;
+using System.Threading;
+
+internal class PatternLayout
+{
+    internal string _conversionPattern;
```
sfc-gh-knozderko, line 10: I would change that to property.

Snowflake.Data/Logger/PatternLayout.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public PatternLayout() { }
```
sfc-gh-knozderko, line 12: do we need that constructor? The default constructor would be generated anyway?

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,85 @@
...
+    internal int _maxSizeRollBackups;
+    internal PatternLayout _patternLayout;
+
+    public SFRollingFileAppender() { }
```
sfc-gh-knozderko, line 18: do we need that constructor? The default constructor would be generated anyway.

Snowflake.Data/Logger/SFConsoleAppender.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public SFConsoleAppender() { }
```
sfc-gh-knozderko, line 13: do we need that constructor? The default constructor would be generated anyway.

sfc-gh-knozderko (changes requested):

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
```diff
@@ -0,0 +1,32 @@
+/*
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
```diff
@@ -1,235 +1,235 @@
-using System;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Reverted

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,128 @@
...
+namespace Snowflake.Data.Log
+{
+    internal class SFLoggerImpl : SFLogger
+    {
+        private readonly Type _type;
+        internal static List<SFAppender> s_appenders = new List<SFAppender>();
+        internal static LoggingEvent s_level = LoggingEvent.OFF;
+
+        private static bool _isDebugEnabled = false;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added prefix

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Replace with FileOperations

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Replace with DirectoryOperations

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Replaced with FileOperations

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/SFConsoleAppender.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/LoggerEmptyImpl.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/PatternLayout.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/SFConsoleAppender.cs
sfc-gh-ext-simba-lf: Removed name field

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Removed name field

Snowflake.Data/Logger/SFConsoleAppender.cs
sfc-gh-ext-simba-lf: Changed to property

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-ext-simba-lf: Turned to properties

Snowflake.Data/Logger/PatternLayout.cs
sfc-gh-ext-simba-lf: Changed

Snowflake.Data/Logger/PatternLayout.cs
sfc-gh-ext-simba-lf: The constructor wasn't necessary, it's now removed

Snowflake.Data/Logger/SFConsoleAppender.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/SFRollingFileAppender.cs
sfc-gh-knozderko: but you still don't validate the file rights in case someone changed it

Snowflake.Data/Logger/SFLogger.cs
```diff
@@ -1,35 +1,28 @@
...
+        void Debug(string msg, Exception ex = null);
+
+        void Info(string msg, Exception ex = null);
+
+        void Warn(string msg, Exception ex = null);
+
+        void Error(string msg, Exception ex = null);
+    }
+
+    public enum LoggingEvent
```
sfc-gh-knozderko, line 24: why public? It is used only internally.

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -51,6 +49,30 @@ public static SFLogger GetLogger<T>()
                 return new SFLoggerEmptyImpl();
             }
         }
-    }
 
+        internal static ILogger GetCustomLogger<T>()
```
sfc-gh-knozderko, line 53: what is the purpose of this method? It is only used in tests.

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.Extensions.Logging;
+using Snowflake.Data.Log;
+
+namespace Snowflake.Data.Client
+{
+    public class SnowflakeDbLoggerFactory
+    {
+        public static void DisableCustomLogger()
+        {
+            SFLoggerFactory.s_isCustomLoggerEnabled = false;
```
sfc-gh-knozderko, line 10: `s_isCustomLoggerEnabled` is only used in unused method GetCustomLogger.
I think it be rather `s_customLogger = new LoggerEmptyImpl()` or even better a call of a method of SFLoggerFactory which would wrap it.

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public static void EnableCustomLogger()
+        {
+            SFLoggerFactory.s_isCustomLoggerEnabled = true;
```
sfc-gh-knozderko, line 15: what is the purpose of this method? Someone configures a custom logger, then disables it and then after some time he would like to enable it without providing the logger (like you do in SetCustomLogger)?
At the starting point the custom logger is disabled, then the user calls `EnableCustomLogger()` and this operation would have no real effect...

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public static void ResetCustomLogger()
+        {
+            SFLoggerFactory.s_customLogger = new LoggerEmptyImpl();
```
sfc-gh-knozderko, line 20: isn't it the same as `DisableCustomLogger`?

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.Extensions.Logging;
+using Snowflake.Data.Log;
+
+namespace Snowflake.Data.Client
+{
+    public class SnowflakeDbLoggerFactory
```
sfc-gh-knozderko, line 6: Actually this class does not create any objects but rather configures the custom logger. So maybe we could change the `Factory` in its name to something else, eg. `SnowflakeDbLoggerConfig`

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        var formattedMessage = PatternLayout.Format(logLevel, message, type);
+        try
+        {
+            if (LogFileIsTooLarge())
+            {
+                RollLogFile();
+            }
+
+            FileOperations.Instance.Write(LogFilePath, formattedMessage, null, true);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Modified to create file handle and validate it. Then use same handle for writing

Snowflake.Data/Logger/SFLogger.cs
sfc-gh-ext-simba-lf: Changed to internal

Snowflake.Data/Logger/SFLoggerFactory.cs
sfc-gh-ext-simba-lf: Moved the function to the test class

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
sfc-gh-ext-simba-lf: This method was only used for testing and has been removed since there is already another method similar to this one

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
sfc-gh-ext-simba-lf: The `DisableCustomLogger` has been removed since it is pretty similar to this method

Snowflake.Data/Client/SnowflakeDbLoggerFactory.cs
sfc-gh-ext-simba-lf: Renamed to `SnowflakeDbLoggerConfig`

sfc-gh-knozderko (changes requested):

Snowflake.Data/Logger/EasyLoggerManager.cs
```diff
@@ -14,23 +11,20 @@ internal class EasyLoggerManager
 
         private readonly object _lockForExclusiveConfigure = new object();
 
-        private const string AppenderPrefix = "SFEasyLogging";
+        internal const string AppenderPrefix = "SFEasyLogging";
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed const

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    public void ActivateOptions()
+    {
+        var logDir = Path.GetDirectoryName(LogFilePath);
+        if (!DirectoryOperations.Instance.Exists(logDir))
+            DirectoryOperations.Instance.CreateDirectory(logDir);
+        if (!FileOperations.Instance.Exists(LogFilePath))
+            FileOperations.Instance.Create(LogFilePath).Dispose();
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Are there other permissions you want to add other than `FileAccessPermissions.UserRead | FileAccessPermissions.UserWrite`

The `FileOperations.Create` function calls `_unixOperations.CreateFileWithPermissions(absolutePath, FileAccessPermissions.UserRead | FileAccessPermissions.UserWrite)` for Unix

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            Console.Error.WriteLine("Encountered an error while writing log to file");
+        }
+    }
+
+    public void ActivateOptions()
+    {
+        var logDir = Path.GetDirectoryName(LogFilePath);
+        if (!DirectoryOperations.Instance.Exists(logDir))
+            DirectoryOperations.Instance.CreateDirectory(logDir);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Are there other permissions you want to add other than `FileAccessPermissions.UserReadWriteExecute`

The `DirectoryOperations.CreateDirectory` function calls `_unixOperations.CreateDirectoryWithPermissions(path, FileAccessPermissions.UserReadWriteExecute)` for Unix

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,47 +1,43 @@
...
-        }
-
-        public static void Instance(SFLogger customLogger)
-        {            
-            logger = customLogger;
+            return new SFLoggerPair(GetSFLogger<T>());
         }
 
-        public static SFLogger GetLogger<T>()
+        internal static SFLogger GetSFLogger<T>(bool useConsoleAppender = false)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed the parameter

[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed `SFLoggerEmptyImpl` from `GetSFLogger` function. It is now only used as the default value of custom logger

Snowflake.Data/Logger/SFLoggerFactory.cs
```diff
@@ -1,47 +1,43 @@
...
+            if (s_useDefaultSFLogger)
             {
-                // If no logger specified, use the default logger: log4net
-                if (logger == null)
+                var logger = new SFLoggerImpl(typeof(T));
+                if (!s_isSFLoggerEnabled)
                 {
-                    ILog loggerL = LogManager.GetLogger(typeof(T));
-                    return new Log4NetImpl(loggerL);
+                    SFLoggerImpl.SetLevel(LoggingEvent.OFF); // Logger is disabled by default and can be enabled by the EasyLogging feature
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Moved it to when the logger is being initialized. Let me know if there's another place you prefer

Snowflake.Data/Logger/EasyLoggerManager.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: It is still there...

Snowflake.Data/Logger/SFRollingFileAppender.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: Oh, I didn't notice. It's OK then.

Snowflake.Data/Logger/SFLoggerFactory.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: now `SFLoggerEmptyImpl` class is unused so you could remove it as well

Snowflake.Data/Core/Session/EasyLoggingStarter.cs
```diff
@@ -69,6 +69,7 @@ public virtual void Init(string configFilePathFromConnectionString)
                     _initTrialParameters = new EasyLoggingInitTrialParameters(configFilePathFromConnectionString);
                     return;
                 }
+                SFLoggerImpl.SetLevel(LoggingEvent.OFF); // Logger is disabled by default and can be enabled by the EasyLogging feature
```
sfc-gh-knozderko, line 72: I think we don't need this

Snowflake.Data/Logger/SFLoggerImpl.cs
```diff
@@ -0,0 +1,123 @@
...
+        internal static void SetLevel(LoggingEvent level)
+        {
+            s_level = level;
+            SetEnableValues();
+        }
+
+        private static void SetEnableValues()
+        {
+            var enabled = s_level != LoggingEvent.OFF;
+            s_isDebugEnabled = enabled;
```
sfc-gh-knozderko, line 31: Generally it works, but let's say that that someone would like to set ERROR log level.
Here you enable DEBUG level and a few lines later you switch it off.
Let's say it you disable it after a few milliseconds.
But in that milliseconds the driver will log with DEBUG log level.
It would be better if you compute all these values first and only then modify `s_*Enabled` fields.

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                File.AppendAllText(LogFilePath, formattedMessage);
+                if (ex != null)
+                    File.AppendAllText(ex.ToString(), formattedMessage);
+            }
+            else
+            {
+                var fileInfo = new UnixFileInfo(path: LogFilePath);
+                using (var handle = fileInfo.Open(FileMode.Append, FileAccess.ReadWrite, FilePermissions.S_IWUSR |  FilePermissions.S_IRUSR))
```
sfc-gh-knozderko, line 40: Look at the "CodeQL / dotnet_format run" section. It reports that this line is wrongly formatted.

sfc-gh-knozderko (changes requested):

Snowflake.Data/Core/Session/EasyLoggingStarter.cs
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/SFLoggerImpl.cs
sfc-gh-ext-simba-lf: Refactored to modify static values only once

Snowflake.Data/Snowflake.Data.csproj
```diff
@@ -24,9 +24,10 @@
     <PackageReference Include="Google.Cloud.Storage.V1" Version="4.10.0" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.13.0" />
     <PackageReference Include="Azure.Storage.Common" Version="12.12.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="9.0.1" />
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to 9.0.5

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -18,10 +18,16 @@
...
     <PackageReference Include="JunitXml.TestLogger" Version="3.1.12" />
     <PackageReference Include="Moq" Version="4.18.4" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Log4Net.AspNetCore" Version="8.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.1" />
+    <PackageReference Include="NLog.Extensions.Logging" Version="5.3.14" />
     <PackageReference Include="NUnit" Version="3.14.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
     <PackageReference Include="RichardSzalay.MockHttp" Version="6.0.0" />
+    <PackageReference Include="Serilog" Version="4.0.1" />
+    <PackageReference Include="Serilog.Extensions.Logging" Version="8.0.0" />
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to 9.0.2

Snowflake.Data.Tests/Snowflake.Data.Tests.csproj
```diff
@@ -18,10 +18,16 @@
...
+    <PackageReference Include="Serilog.Settings.Xml" Version="1.1.0" />
+    <PackageReference Include="Serilog.Sinks.File" Version="6.0.0" />
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Changed to 7.0.0

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                File.AppendAllText(LogFilePath, formattedMessage);
+                if (ex != null)
+                    File.AppendAllText(ex.ToString(), formattedMessage);
+            }
+            else
+            {
+                var fileInfo = new UnixFileInfo(path: LogFilePath);
+                using (var handle = fileInfo.Open(FileMode.Append, FileAccess.ReadWrite, FilePermissions.S_IWUSR | FilePermissions.S_IRUSR))
```
sfc-gh-knozderko, line 40: Actually this appender is used only by easy logging. And I think some users would find 600 permissions too restrictive. So I would like to make it configurable in EasyLogging config.
Currently we have the configuration file like that:
```
{
  "common": {
    "log_level": "DEBUG",
    "log_path": "/path/to/a/file.log"
  }
}
```
we could add a "dotnet" section and "log_file_unix_permissions" field in it:
```
{
  "common": {
    "log_level": "DEBUG",
    "log_path": "/path/to/a/file.log"
  },
  "dotnet": {
    "log_file_unix_permissions": "640"
  }
}
```
and then we could create the file with such permissions and verify if the permissions are not broader that specified in the config file.
If such permissions would not be configured in the config file then we could assume 600 as the default value.

sfc-gh-ext-simba-lf: I updated the PR to include the changes mentioned above:
- Add `dotnet` section and `log_file_unix_permissions` to config file
- Create file with `log_file_unix_permissions` permissions and use those permissions for validating when reading/writing the log file
- Use 600 as the default value for `log_file_unix_permissions` if not configured. Also uses 600 if the value in the config is invalid

Please let me know if there's any edits you'd like to make

Snowflake.Data/Logger/SFLogger.cs
sfc-gh-pbulawa, line 1: Why the chmod of the file is changed?

Snowflake.Data/Logger/SFLoggerFactory.cs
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: I've reverted the mode change for this one as well

Snowflake.Data/Logger/SFConsoleAppender.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+
+internal class SFConsoleAppender : SFAppender
+{
+    internal PatternLayout PatternLayout { get; set; }
+
+    public void Append(string logLevel, string message, Type type, Exception ex = null)
+    {
+        var formattedMessage = PatternLayout.Format(logLevel, message, type);
+        Console.Write(formattedMessage);
```
sfc-gh-knozderko, line 11: why not just WriteLine?

Snowflake.Data.Tests/UnitTests/Logger/SFLoggerPairTest.cs
```diff
@@ -0,0 +1,121 @@
+using NUnit.Framework;
+using Snowflake.Data.Client;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data.Tests/UnitTests/Logger/SFLoggerTest.cs
```diff
@@ -1,10 +1,11 @@
+using NUnit.Framework;
 using Snowflake.Data.Configuration;
+using Snowflake.Data.Core;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Logger/EasyLoggerManager.cs
```diff
@@ -41,97 +35,64 @@ internal static bool IsStdout(string logsPath)
...
+                RemoveOtherEasyLoggingAppenders(null);
             }
         }
 
         internal static bool HasEasyLoggingAppender()
         {
-            var repository = (log4net.Repository.Hierarchy.Hierarchy)LogManager.GetRepository();
-            var rootLogger = (log4net.Repository.Hierarchy.Logger)repository.GetLogger("Snowflake.Data");
-            return rootLogger.Appenders.ToArray().Any(IsEasyLoggingAppender);
+            return SFLoggerImpl.s_appenders.ToArray().Any();
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data.Tests/UnitTests/Logger/EasyLoggingStarterTest.cs
```diff
@@ -241,7 +286,7 @@ public void TestThatConfiguresEasyLoggingOnlyOnceForInitializationsWithoutConfig
                 t_unixOperations.Verify(u => u.CreateDirectoryWithPermissions(s_expectedLogPath,
                     FileAccessPermissions.UserReadWriteExecute), Times.Once);
             }
-            t_easyLoggerManager.Verify(manager => manager.ReconfigureEasyLogging(EasyLoggingLogLevel.Error, s_expectedLogPath), Times.Once);
+            t_easyLoggerManager.Verify(manager => manager.ReconfigureEasyLogging(EasyLoggingLogLevel.Error, s_expectedLogPath));
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Reverted, it's been a while so I'm not sure why it was changed in the first place

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                File.AppendAllText(LogFilePath, formattedMessage);
+                if (ex != null)
+                    File.AppendAllText(ex.ToString(), formattedMessage);
+            }
+            else
+            {
+                var fileInfo = new UnixFileInfo(path: LogFilePath);
+                using (var handle = fileInfo.Open(FileMode.Append, FileAccess.ReadWrite, (FilePermissions)EasyLoggingStarter.s_logFileUnixPermissions))
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Modified to check file permissions

The line after that [`SFCredentialManagerFileImpl.Instance.ValidateLogFilePermissions(handle);`](https://github.com/snowflakedb/snowflake-connector-net/pull/1057/files/e0bb3e86aa8726e4d07a7722bb72d5653e4b2000#diff-1b5b780079df1284db083a23e6feb8a73e0a28a4d715bed3df12b5a75e0d7facR43) checks for user owner, group owner, and file access permissions

Snowflake.Data/Logger/SFRollingFileAppender.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                File.AppendAllText(LogFilePath, formattedMessage);
+                if (ex != null)
+                    File.AppendAllText(ex.ToString(), formattedMessage);
+            }
+            else
+            {
+                var fileInfo = new UnixFileInfo(path: LogFilePath);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done, added an append function to UnixOperations and used that instead

Snowflake.Data.Tests/UnitTests/Logger/SFLoggerPairTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            // Log level defaults to Warn on net6.0 builds in github actions
+            // Set the root level to Debug
+            EasyLoggerManager.Instance.ReconfigureEasyLogging(EasyLoggingLogLevel.Debug, "STDOUT");
+        }
+
+        [OneTimeTearDown]
+        public static void AfterAll()
+        {
+            EasyLoggerManager.Instance.ReconfigureEasyLogging(EasyLoggingLogLevel.Warn, "STDOUT");
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It's now disabled after the tests

The original value was `Warn` when I was testing so it was reverted back to that

Snowflake.Data.Tests/UnitTests/Logger/EasyLoggingStarterTest.cs
```diff
@@ -89,6 +128,12 @@ public void BeforeEach()
                 t_environmentOperations.Object);
         }
 
+        [TearDown]
+        public void AfterEach()
+        {
+            EasyLoggingStarter.s_logFileUnixPermissions = FileAccessPermissions.UserWrite | FileAccessPermissions.UserRead;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Removed

Snowflake.Data/Core/CredentialManager/Infrastructure/SFCredentialManagerFileImpl.cs
```diff
@@ -269,16 +269,28 @@ private void ReleaseLock()
 
         internal void ValidateFilePermissions(UnixStream stream)
         {
+            ValidateUserAndGroupPermissions(stream);
             var allowedPermissions = new[]
             {
                 FileAccessPermissions.UserRead | FileAccessPermissions.UserWrite
             };
+            if (!(allowedPermissions.Any(a => stream.FileAccessPermissions == a)))
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Yeah that makes sense, I reverted the CredentialManager and made a new file specifically for logging

Snowflake.Data/Logger/EasyLoggerManager.cs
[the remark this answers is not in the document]
sfc-gh-knozderko: It still is there...

Snowflake.Data/Core/Tools/UnixOperations.cs
```diff
@@ -186,6 +185,24 @@ public virtual void WriteAllBytes(string path, byte[] bytes, Action<UnixStream>
             }
         }
 
+        public long AppendToFile(string path, string mainContent, string additionalContent, Action<UnixStream> validator, FileAccessPermissions permissions)
+        {
+            var fileInfo = new UnixFileInfo(path: path);
+            if (!fileInfo.Exists)
+                CreateFileWithPermissions(path, permissions);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: I added a method parameter to determine if it should log or not

When it's appending to a log file, that parameter will be set to false so it doesn't cause the infinite recursion

sfc-gh-mhofman (approved):

sfc-gh-mhofman:
Changes already mentioned in the changelog (in a separate PR).
Merging this since sfc-gh-knozderko requests were addressed and CI passed.

PR #1067, opened 2024-12-02
SNOW-1569293 Read encryption headers in a case insensitive way

sfc-gh-knozderko:
### Description
SNOW-1569293 Read encryption headers in a case insensitive way

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
+++ b/Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
@@ -9,6 +9,7 @@ using Amazon.S3.Model;
 using Snowflake.Data.Log;
 using System;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


Snowflake.Data/Core/FileTransfer/StorageClient/SFS3Client.cs
```diff
@@ -266,26 +267,38 @@ private GetObjectRequest GetFileHeaderRequest(ref AmazonS3Client client, SFFileM
...
 
+        private string GetMetadataCaseInsensitive(MetadataCollection metadataCollection, string metadataKey)
+        {
+            var value = metadataCollection[metadataKey];
+            if (value != null)
+                return value;
+            if (string.IsNullOrEmpty(metadataKey))
+                return null;
+            var keysCaseInsensitive = metadataCollection.Keys
+                .Where(key => $"x-amz-meta-{metadataKey}".Equals(key, StringComparison.OrdinalIgnoreCase));
```
sfc-gh-pfus, line 298: Therre was no prefixing with `x-amz-meta-` before, why is it now?

sfc-gh-knozderko: The class MetadataCollection automatically adds "x-amz-meta-" prefix to all its elements.
So if you would like to store `X-AMZ-IV` in this collection, you are really storing for the key of 'x-amz-meta-X-AMZ-IV'. That's why when I would like to find a key which is matching to `X-AMZ-IV` I need to look for 'x-amz-meta-X-AMZ-IV'.
MetadataCollection.MetaDataHeaderPrefix prefix is internal so I couldn't take the value of prefix from the AWS library and I had to duplicate the prefix value here.

sfc-gh-pfus (approved):

sfc-gh-pmotacki (approved):

PR #1083, opened 2025-01-24
Fix STDOUT logger bug

Coder3333:
### Description
Fixed a bug that prevented the STDOUT logger from ever being used, because the GetLogPath method would change the input logPath, which would prevent ReconfigureEasyLogging from recognizing the request for STDOUT.

### Checklist
- [ ] Code compiles correctly


Coder3333:

Snowflake.Data/Core/Session/EasyLoggingStarter.cs
```diff
@@ -118,6 +118,12 @@ private EasyLoggingLogLevel GetLogLevel(string logLevel)
 
         private string GetLogPath(string logPath)
         {
+            if (EasyLoggerManager.UseSTDOUT(logPath))
```
sfc-gh-knozderko, line 121: hi @Coder3333 I appreciate your work. We could fix handling of STDOUT because it could be considered as a bug.

However I'm trying to understand your case and I think that you are trying to use this feature to configure logging  in normal scenario whereas it was meant mainly for debug purpose in extraordinary situations. But of course you can use it as you want.

But maybe your real problem is that you have issue with configuring log4net? If log4net is not enabled in your application, EasyLogging will not be working anyway because it configures appenders for an assumption that log4net is working.

We are preparing this PR with removing log4net: https://github.com/snowflakedb/snowflake-connector-net/pull/1057 When this PR is merged you will be able to use any logger you want, but Easy Logging feature will be implemented by adding an extra logger, a simple logger implemented by us which might be less efficient than log4net (we don't do any buffering for messages, because the assumption was that EasyLogging is mainly for debug purpose). 

That's why I would suggest you reconsider the way you would like to use the feature and be aware of the future changes.

Coder3333: > hi @Coder3333 I appreciate your work. We could fix handling of STDOUT because it could be considered as a bug.
> 
...

@sfc-gh-knozderko , we are only interested in using this feature, because Snowflake support asked us to provide client side logs in response to a Production situation that no one can explain.  We have the exception stack trace that is collected by our logs, but Snowflake support believes more information could be collected by the .Net Connector's logs.  Once the issue is resolved, we would turn the .Net Connector logging back off.

sfc-gh-knozderko:
I'm closing this PR because I did some small changes and added a test for in this PR: https://github.com/snowflakedb/snowflake-connector-net/pull/1087
which is already merged.

PR #1100, opened 2025-02-19
Fix ConstraintException when calling DataTable.Load

sfc-gh-ext-simba-lf:
### Description
Fix the `System.Data.ConstraintException` error that occurs when querying a VARIANT column with `DataTable.Load`

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbDataReader.cs
+++ b/Snowflake.Data/Client/SnowflakeDbDataReader.cs
@@ -122,7 +122,7 @@ namespace Snowflake.Data.Client
 
                 row[SchemaTableColumn.ColumnName] = rowType.name;
                 row[SchemaTableColumn.ColumnOrdinal] = columnOrdinal;
-                row[SchemaTableColumn.ColumnSize] = (int)rowType.length;
+                row[SchemaTableColumn.ColumnSize] = rowType.length == 0 ? -1 : (int)rowType.length;
                 row[SchemaTableColumn.NumericPrecision] = (int)rowType.precision;
                 row[SchemaTableColumn.NumericScale] = (int)rowType.scale;
                 row[SchemaTableColumn.AllowDBNull] = rowType.nullable;
--- a/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
@@ -415,7 +415,7 @@ namespace Snowflake.Data.Tests.IntegrationTests
             row = dataTable.Rows[2];
             Assert.AreEqual(s_colName[2], row[SchemaTableColumn.ColumnName]);
             Assert.AreEqual(2, row[SchemaTableColumn.ColumnOrdinal]);
-            Assert.AreEqual(0, row[SchemaTableColumn.ColumnSize]);
+            Assert.AreEqual(-1, row[SchemaTableColumn.ColumnSize]);
             Assert.AreEqual(SFDataType.FIXED, (SFDataType)row[SchemaTableColumn.ProviderType]);
         }
 
--- a/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
+++ b/Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
@@ -1637,6 +1637,47 @@ namespace Snowflake.Data.Tests.IntegrationTests
             }
         }
 
+        [Test]
+        public void TestDataTableLoadOnVariantColumn()
+        {
+            using (var conn = CreateAndOpenConnection())
+            {
+                var colName = "c1";
+                var expectedVal = "{\"id\":1}"; // "{\n  \"id\": 1\n}"
+
+                CreateOrReplaceTable(conn, TableName, new[] { $"{colName} variant" });
+
+                IDbCommand cmd = conn.CreateCommand();
+
+                string insertCommand = $"insert into {TableName} select parse_json('{expectedVal}')";
+                cmd.CommandText = insertCommand;
+
+                var count = cmd.ExecuteNonQuery();
+                Assert.AreEqual(1, count);
+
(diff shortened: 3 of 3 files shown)


Snowflake.Data/Client/SnowflakeDbDataReader.cs
```diff
@@ -118,7 +118,7 @@ private DataTable PopulateSchemaTable(SFBaseResultSet resultSet)
 
                 row[SchemaTableColumn.ColumnName] = rowType.name;
                 row[SchemaTableColumn.ColumnOrdinal] = columnOrdinal;
-                row[SchemaTableColumn.ColumnSize] = (int)rowType.length;
+                row[SchemaTableColumn.ColumnSize] = IsSemiStructuredType(rowType.type) && rowType.length == 0 ? -1 : (int)rowType.length;
```
sfc-gh-knozderko, line 121: what does -1 represent?
What is the reason of the error?

Snowflake.Data/Client/SnowflakeDbDataReader.cs
```diff
@@ -365,5 +365,11 @@ public override void Close()
             resultSet.close();
             isClosed = true;
         }
+
+        private bool IsSemiStructuredType(string type)
+        {
+            type = type.ToLower();
+            return type == "array" || type == "object" || type == "variant";
```
sfc-gh-knozderko, line 372: I'm worried about structured types.
Such a way of recognising semi structured type seems not to be correct because we have structured types returned as "array", "object" or "map". Normally you would need to verify existence  fields in the response to know whether it is a structured type or semi structured type.
Could you create a similar test for structured types if it makes sense?

Could you v

Snowflake.Data/Client/SnowflakeDbDataReader.cs
sfc-gh-ext-simba-lf: -1 represents the column size is unconstrained

The reason for the error is calling `DataTable.Load()` on a `VARIANT` column with a size of 0. It looks like `VARIANT` doesn't map to a .NET data type and trying to fill the data table that has 0 column size throws the `ConstraintException`. Other types like `string` or `int` don't encounter this issue which is likely because they directly map to a .NET data type

Snowflake.Data/Client/SnowflakeDbDataReader.cs
sfc-gh-ext-simba-lf: I added tests using the structured types and without the fix, it threw the same error when calling `DataTable.Load()`

When the fix was added, the tests pass for both `array` and `object` but the `map` data type still encountered the same `ConstraintException`

sfc-gh-ext-simba-lf: There is a workaround for this issue that doesn't require any code changes by setting `EnforceConstraints = false`. Adding this made all the test pass without changing anything on the connector:
```
using (DataSet ds = new DataSet() { EnforceConstraints = false })
{
    ds.Tables.Add(dt);
    dt.Load(reader, LoadOption.OverwriteChanges);
    ds.Tables.Remove(dt);
}
```

Perhaps it would be safer to use the `EnforceConstraints` parameter instead of changing the code since the exception only happens when using the `DataTable` class?

sfc-gh-knozderko: If you reverted the fix you didn't fix anything - you are just claiming that anyone who creates a DataSet should disable constraints on it.

I think it would be better to add a fix for that and not to be forced to disabled constraints.

So the problem with your fix was that you should add "map" to the "variant", "object", "array" set and rename the method IsSemiStructuredType to let's say IsStructuredOrSemistructuredType.

The values for these objects are json strings. So I have doubts if -1 value is good for their size. Maybe it should rather be their length because they are just strings?

sfc-gh-ext-simba-lf: I renamed the function and added "map" to it

For strings, the `ColumnSize` was 16777216 which is the max allowed length for VARCHAR

Snowflake.Data.Tests/IntegrationTests/StructuredMapsIT.cs
```diff
@@ -11,6 +12,40 @@ namespace Snowflake.Data.Tests.IntegrationTests
...
+                    command.CommandText = $"SELECT {addressAsSFString} AS {colName}";
+
+                    // act
+                    using (var reader = command.ExecuteReader())
+                    {
+                        var dt = new DataTable();
+                        dt.Load(reader);
+
+                        // assert
+                        Assert.AreEqual($"{key}:{value}", dt.Rows[0][colName].ToString()
```
sfc-gh-knozderko, line 38: This unwrapping of the result is duplicated in several places.
Let's extract it and move to one place maybe to some class in Snowflake.Data.Tests.Util.
But the way you extract it is not perfect. Eg. you needed to replace "San Mateo" with "SanMateo" because you are removing all spaces from the result...
Could you use Newtonsoft to extract a value from the json instead of doing these replacements?

Snowflake.Data/Client/SnowflakeDbDataReader.cs
```diff
@@ -365,5 +367,11 @@ public override void Close()
             resultSet.close();
             isClosed = true;
         }
+
+        private bool IsStructuredOrSemistructuredType(string type)
```
sfc-gh-knozderko, line 371: maybe you could rename to IsStructuredOrSemiStructuredType ?

sfc-gh-ext-simba-lf: Renamed

Snowflake.Data.Tests/IntegrationTests/StructuredMapsIT.cs
sfc-gh-ext-simba-lf: Put the extract into its own function and use Newtonsoft to get the value instead

sfc-gh-pcyrek (approved):
lgtm

PR #1103, opened 2025-02-24
SNOW-1946997 Handle username_password_mfa and oauth authenticators in a case ins…

sfc-gh-knozderko:
…ensitive way

### Description
SNOW-1946997 Handle username_password_mfa and oauth authenticators in a case insensitive way

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/Session/SessionPool.cs
```diff
@@ -142,7 +142,8 @@ internal SFSession GetSession(string connStr, SecureString password, SecureStrin
             ValidateMinPoolSizeWithPasscode(sessionProperties, passcode);
             if (!GetPooling())
                 return NewNonPoolingSession(connStr, password, passcode);
-            var isMfaAuthentication = sessionProperties.TryGetValue(SFSessionProperty.AUTHENTICATOR, out var authenticator) && authenticator == MFACacheAuthenticator.AuthName;
+            var isMfaAuthentication = sessionProperties.TryGetValue(SFSessionProperty.AUTHENTICATOR, out var authenticator) &&
```
sfc-gh-mhofman, line 145: maybe it could be worth to check if sessionProperties during setting its values can be populated with some of their properties lowercase or uppercase? then here and everywhere it could be just simple comparison instead of a MFA-related comparator

sfc-gh-knozderko: the case of Okta authenticator is problematic because in okta case the authenticator is an url. If I would do toLower() on Url I could possibly damage it. Instead of converting toLower for some authenticators and for some others not I just made such functions in all the other authenticators so it is consistent now.

sfc-gh-mhofman: got it

sfc-gh-pbulawa (approved):

Snowflake.Data.Tests/UnitTests/AuthenticationPropertiesValidatorTest.cs
```diff
@@ -41,12 +40,13 @@ public void TestAuthPropertiesValid(string connectionString, string password)
...
-        [TestCase("authenticator=okta;", null, SFError.MISSING_CONNECTION_PROPERTY, "Error: Required property PASSWORD is not provided")]
+        [TestCase("authenticator=https://okta.com;user=;password=", null, SFError.MISSING_CONNECTION_PROPERTY, "Error: Required property PASSWORD is not provided")]
+        [TestCase("authenticator=https://okta.com;user=", null, SFError.MISSING_CONNECTION_PROPERTY, "Error: Required property PASSWORD is not provided")]
+        [TestCase("authenticator=https://okta.com;password=", null, SFError.MISSING_CONNECTION_PROPERTY, "Error: Required property PASSWORD is not provided")]
+        [TestCase("authenticator=https://okta.com;", null, SFError.MISSING_CONNECTION_PROPERTY, "Error: Required property PASSWORD is not provided")]
         [TestCase("authenticator=unknown;", null, SFError.UNKNOWN_AUTHENTICATOR, "Unknown authenticator")]
         [TestCase("authenticator=http://unknown.okta.com;", null, SFError.UNKNOWN_AUTHENTICATOR, "Unknown authenticator")]
+        [TestCase("AUTHENTICATOR=HTTPS://SOMETHING.OKTA.COM;USER=TEST;PASSWORD=TEST", null, SFError.UNKNOWN_AUTHENTICATOR, "Unknown authenticator")]
```
sfc-gh-dstempniak, line 49: I think uppercased parameters and values should be allowed. But it might be a different task.

sfc-gh-dstempniak (approved):
LGTM

Snowflake.Data.Tests/UnitTests/AuthenticationPropertiesValidatorTest.cs
sfc-gh-knozderko: it was not allowed for now, wo I don't want to change this behaviour in this PR.

PR #1115, opened 2025-03-13
SNOW-1917171 use cache and refresh token

sfc-gh-knozderko:
### Description
SNOW-1917171 use cache and refresh token

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
@@ -1,11 +1,15 @@
 using System.IO;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 using Moq;
 using NUnit.Framework;
 using Snowflake.Data.Client;
 using Snowflake.Data.Core;
 using Snowflake.Data.Core.Authenticator;
 using Snowflake.Data.Core.Authenticator.Browser;
+using Snowflake.Data.Core.CredentialManager;
+using Snowflake.Data.Core.CredentialManager.Infrastructure;
 using Snowflake.Data.Core.Session;
 using Snowflake.Data.Core.Tools;
 using Snowflake.Data.Tests.Mock;
@@ -22,7 +26,21 @@ namespace Snowflake.Data.Tests.UnitTests.Authenticator
         private static readonly string s_invalidScopeErrorMappingPath = Path.Combine(s_oauthAuthorizationCodeMappingPath, "invalid_scope_error.json");
         private static readonly string s_invalidStateErrorMappingPath = Path.Combine(s_oauthAuthorizationCodeMappingPath, "invalid_state_error.json");
         private static readonly string s_badTokenRequestErrorMappingPath = Path.Combine(s_oauthAuthorizationCodeMappingPath, "token_request_error.json");
+        private static readonly string s_refreshTokenMappingPath = Path.Combine(s_oauthAuthorizationCodeMappingPath, "refresh_token.json");
         private static readonly string s_oauthSnowflakeLoginSuccessMappingPath = Path.Combine(s_oauthMappingPath, "snowflake_successful_login.json");
+        private static readonly string s_oauthSnowflakeLoginInvalidTokenMappingPath = Path.Combine(s_oauthMappingPath, "snowflake_invalid_token_login.json");
+        private const string MasterToken = "masterToken123";
+        private const string SessionToken = "sessionToken123";
+        private const string AccessToken = "access-token-123";
+        private const string RefreshToken = "refresh-token-123";
+        private const string NewAccessToken = "new-access-token-123";
+        private const string NewRefreshToken = "new-refresh-token-123";
+        private const string SessionId = "1234567890";
+        private const string User = "testUser";
+        private const string AuthorizationScope = "session:role:ANALYST";
+        private const string TokenHost = "localhost";
+        private static readonly string s_externalAuthorizationUrl = $"http://localhost:{WiremockRunner.DefaultHttpPort}/oauth/authorize";
+        private static readonly string s_externalTokenRequestUrl = $"http://localhost:{WiremockRunner.DefaultHttpPort}/oauth/token-request";
 
         private WiremockRunner _runner;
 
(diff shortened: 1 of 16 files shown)


Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
```diff
@@ -35,32 +53,184 @@ public void BeforeAll()
...
+        }
+
+        [Test]
+        public void TestSuccessfulAuthorizationCodeFlowWithDefaultCache()
+        {
+            // arrange
+            try
+            {
+                SnowflakeCredentialManagerFactory.UseDefaultCredentialManager();
+                RemoveTokenFromCache(TokenType.OAuthAccessToken);
```
sfc-gh-astachowski, line 119: I think we should try and leave the manager in the same state it was before the test, so save these tokens and restore them in the finally block. Ideally this doesn't matter, but i think the test shouldn't have side effects.

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
```diff
@@ -35,32 +53,184 @@ public void BeforeAll()
...
-            Assert.NotNull(session._accessToken);
-            Assert.AreEqual("access-token-123", SecureStringHelper.Decode(session._accessToken.Value));
-            Assert.AreEqual("1234567890", session.sessionId);
-            Assert.AreEqual("masterToken123", session.masterToken);
-            Assert.AreEqual("sessionToken123", session.sessionToken);
+            Assert.NotNull(authenticator.AccessToken);
+            Assert.AreEqual(AccessToken, SecureStringHelper.Decode(authenticator.AccessToken));
+            Assert.AreEqual(SessionId, session.sessionId);
+            Assert.AreEqual(MasterToken, session.masterToken);
+            Assert.AreEqual(SessionToken, session.sessionToken);
```
sfc-gh-astachowski, line 84: These 3 assertions on session's fields and 2 on authenticator happen a lot (7 times I think) - maybe worth extracting them to a helper method.

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-knozderko: this test would have a site effect of removing cache for keys:
- "LOCALHOST:TESTUSER:OAUTH_ACCESS_TOKEN"
- "LOCALHOST:TESTUSER:OAUTH_REFRESH_TOKEN"
It looks like nothing important.
This side effect would be important only if on a given machine there were being executed tests from different drivers in the same time but that's not how it works.
And even if we had 2 concurrent drivers' tests at the same time (eg. .net and jdbc) it would be working wrongly even if I would remove this side effect.
I think removing this could then confuse someone who would think it was important but it wasn't.

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-knozderko: done

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
```diff
@@ -81,9 +135,61 @@ public void Authenticate()
...
+                    var realException = UnpackAggregateException(exception);
+                    throw new SnowflakeDbException(SFError.OAUTH_TOKEN_REQUEST_ERROR, realException.Message);
+                }
+                HandleAccessTokenResponse(accessTokenResponse, cacheKeys);
             }
-            base.Login();
+        }
+
+        private bool ShouldRefreshToken(SnowflakeDbException exception) =>
+            OAuthTokenErrors.IsAccessTokenExpired(exception.ErrorCode) || OAuthTokenErrors.IsAccessTokenInvalid(exception.ErrorCode);
```
sfc-gh-dstempniak, line 166: Should we refresh the token if it is invalid (not expired)? I see that in `OAuthRefreshAccessTokenRequest` RefreshToken is needed. Will this request succeed if the token is invalid?

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
```diff
@@ -35,32 +53,173 @@ public void BeforeAll()
         [SetUp]
         public void BeforeEach()
         {
+            SnowflakeCredentialManagerFactory.SetCredentialManager(new SFCredentialManagerInMemoryImpl());
```
sfc-gh-dstempniak, line 56: Can we move it to `OneTimeSetUp` and move `TestSuccessfulAuthorizationCodeFlowWithDefaultCache` test to a different file? I understand that we test code flow here, not a specific credential manager so we shouldn't call `UseDefaultCredentialManager` before a test.

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
```diff
@@ -81,9 +135,61 @@ public void Authenticate()
...
+
+        private async Task GetAccessTokenAsync(
+            BaseOAuthAccessTokenRequest accessTokenRequest,
+            OAuthCacheKeys cacheKeys,
+            CancellationToken cancellationToken)
+        {
+            var restRequester = (RestRequester) session.restRequester;
+            using (var accessTokenHttpRequest = accessTokenRequest.CreateHttpRequest())
+            {
+                var restRequest = new RestRequestWrapper(accessTokenHttpRequest, TimeSpan.FromSeconds(120));
```
sfc-gh-mhofman, line 150: fix this magic number

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
```diff
@@ -81,9 +135,61 @@ public void Authenticate()
...
+                OAuthAccessTokenResponse accessTokenResponse = null;
+                try
+                {
+                    accessTokenResponse = await restRequester.PostAsync<OAuthAccessTokenResponse>(restRequest, cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception exception)
+                {
+                    var realException = UnpackAggregateException(exception);
```
sfc-gh-mhofman, line 158: shouldn't we log this error?

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
sfc-gh-knozderko: I had the same questions at the beginning, but I discussed it with the others and they told me that I should treat invalid and expired tokens the same way.

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
```diff
@@ -53,20 +57,70 @@ public static bool IsOAuthAuthorizationCodeAuthenticator(string authenticator) =
...
+                AccessToken = SecureStringHelper.Encode(accessToken);
+            }
+
+            try
+            {
+                await LoginAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (SnowflakeDbException exception)
+            {
+                if (ShouldRefreshToken(exception))
```
sfc-gh-mhofman, line 78: i'm missing here anything in the logs  but feel free to improve it later, this is general note on lack of logging. 
Successful and most importantly failure indication that something went wrong here is a must imho.

sfc-gh-mhofman (approved):
Left some comments but you can follow on that later

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-knozderko: What could we gain by such a refactor?
Internals like constants or methods would need to be exposed to a different test class or duplicated.
Wiremock would need to be setup once again.
But I don't see much we can gain by that.
There is no problem with changing the credential manager implementation because all tests are NonParallelizable.

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
sfc-gh-knozderko: I will add logs in a future PR

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
sfc-gh-knozderko: I will add logs in a future PR

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-dstempniak: We could gain separation of code flow tests from credential manager tests.
Or there is another option - all tests could use the default manager. In such a case, all tests will additionally verify OS-based credential manager.

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
```diff
@@ -35,32 +53,173 @@ public void BeforeAll()
...
+            var session = PrepareSession();
+            var authenticator = (OAuthAuthorizationCodeAuthenticator) session.GetAuthenticator();
+
+            // act
+            session.Open();
+
+            // assert
+            Assert.NotNull(authenticator.AccessToken);
+            Assert.AreEqual(NewAccessToken, SecureStringHelper.Decode(authenticator.AccessToken));
+            Assert.AreEqual(SessionId, session.sessionId);
```
sfc-gh-astachowski, line 196: Missed one `AssertSessionSuccessfullyCreated` use here

sfc-gh-pbulawa (approved):

Snowflake.Data/Core/Authenticator/OAuthAuthorizationCodeAuthenticator.cs
```diff
@@ -81,9 +135,61 @@ public void Authenticate()
...
+                AuthorizationScope = GetAuthorizationScope(),
+                RefreshToken = refreshToken
+            };
+        }
+
+        private OAuthCacheKeys GetOAuthCacheKeys()
+        {
+            var host = new Uri(GetTokenEndpoint()).Host;
+            var user = session.properties[SFSessionProperty.USER];
+            var accessTokenKey = SnowflakeCredentialManagerFactory.GetSecureCredentialKey(host, user, TokenType.OAuthAccessToken);
```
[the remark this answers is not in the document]
sfc-gh-knozderko: I added a logic that if that if user or host is null or empty we won't use cache

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-knozderko: done

Snowflake.Data.Tests/UnitTests/Authenticator/OAuthAuthorizationCodeFlowTest.cs
sfc-gh-knozderko: we discussed to stay it as is

PR #1120, opened 2025-03-19
SNOW-1917171 Fix secret detector

sfc-gh-knozderko:
### Description
SNOW-1917171 Fix secret detector

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Logger/SecretDetector.cs
+++ b/Snowflake.Data/Logger/SecretDetector.cs
@@ -87,8 +87,8 @@ namespace Snowflake.Data.Log
         private const string PrivateKeyPropertyPrefixPattern = @"(private_key\s*=)";
         private const string ConnectionTokenPattern = @"(token|assertion content)(['""\s:=]+)([a-z0-9=/_\-+:]{8,})";
         private const string TokenPropertyPattern = @"(token)(\s*=)(.*)";
-        private const string PasswordPattern = @"(password|passcode|client_secret|pwd|proxypassword|private_key_pwd)(['""\s:=]+)([a-z0-9!""#$%&'\()*+,-./:;<=>?@\[\]\^_`{|}~]{6,})";
-        private const string PasswordPropertyPattern = @"(password|passcode|client_secret|proxypassword|private_key_pwd)(\s*=)(.*)";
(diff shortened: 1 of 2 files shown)


Snowflake.Data/Logger/SecretDetector.cs
```diff
@@ -87,8 +87,8 @@ private static string MaskCustomPatterns(string text)
         private const string PrivateKeyPropertyPrefixPattern = @"(private_key\s*=)";
         private const string ConnectionTokenPattern = @"(token|assertion content)(['""\s:=]+)([a-z0-9=/_\-+:]{8,})";
         private const string TokenPropertyPattern = @"(token)(\s*=)(.*)";
-        private const string PasswordPattern = @"(password|passcode|client_secret|pwd|proxypassword|private_key_pwd)(['""\s:=]+)([a-z0-9!""#$%&'\()*+,-./:;<=>?@\[\]\^_`{|}~]{6,})";
-        private const string PasswordPropertyPattern = @"(password|passcode|client_secret|proxypassword|private_key_pwd)(\s*=)(.*)";
+        private const string PasswordPattern = @"(password|passcode|client_?secret|pwd|proxypassword|private_key_pwd)(['""\s:=]+)([a-z0-9!""#$%&'\()*+,-./:;<=>?@\[\]\^_`{|}~]{6,})";
```
sfc-gh-mkubik, line 90: why do we have `?` here?

sfc-gh-knozderko: This way I cover "clientsecret" and "client_secret" as well.
I thought it would be nice to cover client_secret form as well...

sfc-gh-pbulawa (approved):

Snowflake.Data/Logger/SecretDetector.cs
sfc-gh-pmotacki, line 90: Is it correct to have client_secret here? Should it be changed to oauth....?

sfc-gh-pmotacki (approved):

Snowflake.Data/Logger/SecretDetector.cs
sfc-gh-knozderko: I just wanted to be more secure. Maybe clientsecret or client_secret could appear as a part of an url query parameters or some json... So I wanted cover more cases than only the most basic ones.

sfc-gh-mkubik (approved):

PR #1248, opened 2025-09-30
SNOW-2181354 Support TLS 1.3 if possible

sfc-gh-mhofman:
### Description
TLS 1.3 support and parameters: MINTLS and MAXTLS to control the version of a protocol negotiated with the server

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data.Tests/AuthenticationTests/AuthConnectionString.cs
+++ b/Snowflake.Data.Tests/AuthenticationTests/AuthConnectionString.cs
@@ -202,7 +202,7 @@ namespace Snowflake.Data.AuthenticationTests
                 using (var client = new HttpClient(new HttpClientHandler
                 {
                     CheckCertificateRevocationList = true,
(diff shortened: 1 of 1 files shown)


sfc-gh-mgemra (approved):

Snowflake.Data/Core/SslProtocolsExtensions.cs
sfc-gh-snoonan, line 1: Let's make sure we provide a way for the user to set min/max? Depending on the compliance requirements they have they might need to enforce TLS1.3, or they may be running on systems where enabling TLS1.3 causes failures.

sfc-gh-mhofman: However, I'll introduce MinTLS, MaxTLS as you suggest @sfc-gh-snoonan so it can be controlled from the config, cc @sfc-gh-dszmolka

sfc-gh-snoonan (approved):

sfc-gh-mgemra (approved):

PR #1251, opened 2025-10-03
SNOW-2391059 - simplify control of CRL checks and change the default implementation

sfc-gh-mhofman:
### Description
Removed UseDotNetCrlCheck parameter to have less variables controlling same thing.
Introduced new value of CertRevocationCheckMode = Native instead of the above switch.
This change actually switches the default CRL Check algorithm to the one provided by the driver instead of the one shipped by the .NET Framework itself. 

### Checklist
- [x] Code compiles correctly


sfc-gh-mgemra (commented):
Please update alos the Connecting.md file

sfc-gh-dszmolka:
> Please update also the `Connecting.md` file

perhaps CertificateValidation.md too, to point out CRL checks are now disabled by default? (earlier: enabled by default, with the 'old' .NET builtin CRL check method)

PR #1253, opened 2025-10-06
SNOW-2371457 Add crl configurable download timeout

sfc-gh-mgemra:
### Description
[SNOW-2371457](https://snowflakecomputing.atlassian.net/browse/SNOW-2371457): Adding download timeout parameter to the connection string. The TTL configuration remains untouched in this PR and will be handled in a separate PR.
### Checklist
- [ ] Code compiles correctly

[SNOW-2371457]: https://snowflakecomputing.atlassian.net/browse/SNOW-2371457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Snowflake.Data/Core/HttpUtil.cs
```diff
@@ -33,6 +33,7 @@ public HttpClientConfig(
             bool enableCRLDiskCaching = true,
             bool enableCRLInMemoryCaching = true,
             bool allowCertificatesWithoutCrlUrl = true,
+            int crlDownloadTimeout = 60,
```
sfc-gh-pfus, line 36: What kind of timeout it is? Is it 60 seconds? It's far too much IMHO.
1) CRLs usually are below 20MB. In that case 60s is far too much. Even if we have large CRL, like 100MB, it should still be downloaded in seconds. If you have to wait 60s in 99% it means everything hangs and it is not worth waiting that long.
2) Is there total timeout for HTTPS request (usually login request, since we should use keep alive)? If it is shorter, it doesn't make sense to have particular timeout larger.

sfc-gh-mgemra: I believe this is a total timeout (connection + read). I simply copied the hardcoded value - 60s from the existing code, but made the parameter configurable. I guess Krzysiek copied the default values from the jdbc driver (?). Still, it probably makes sense to change it. Do you think 20s will be ok?

sfc-gh-mhofman: 10s seems legit, especially better when advisory mode will be on, so we will not wait 1min when host is unreachable.
Thanks for the change

PR #1279, opened 2025-11-13
SNOW-2216772 Honor session TIMEZONE parameter for TIMESTAMP_LTZ datatype

sfc-gh-mgemra:
### Description
SNOW-2216772 This PR implements an opt-in mechanism for the `TIMESTAMP_LTZ` session timezone feature to avoid a breaking change.

A new optional connection parameter was added:
Default: false (preserves legacy behavior)
When true: TIMESTAMP_LTZ values are converted using the session timezone from Snowflake
When false: TIMESTAMP_LTZ values are converted using the local machine timezone (legacy behavior)
Usage:
```
var connectionString = "account=...;user=...;password=...;HonorSessionTimezone=true;";
```

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
     - Bug fix: Fixed `IndexOutOfRangeException` in Arrow result chunk processing by adding retry state cleanup, batch integrity validation, and defensive bounds checking in `ExtractCell()`.
 - v5.4.0
     - Added support for [DECFLOAT](https://docs.snowflake.com/en/sql-reference/data-types-numeric#decfloat) data type (returned as string to preserve full precision).
-    - Bug fix: Fixed IndexOutOfRangeException in Arrow result processing when empty batches are returned by Snowflake backend.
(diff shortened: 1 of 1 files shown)


sfc-gh-pbulawa (approved):

Snowflake.Data.Tests/IntegrationTests/SFDbDataReaderIT.cs
```diff
@@ -561,8 +565,8 @@ public void TestGetTimestampLTZ()
                 DateTimeOffset dtOffset = (DateTimeOffset)reader.GetValue(0);
                 reader.Close();
 
-                Assert.AreEqual(now, dtOffset);
-                Assert.AreEqual(now.Offset, dtOffset.Offset);
+                Assert.AreEqual(insertValue.UtcDateTime, dtOffset.UtcDateTime);
+                Assert.AreEqual(TimeSpan.FromHours(-8), dtOffset.Offset);
```
sfc-gh-mhofman: I'm wondering if during the daylight saving time of the year it might be not equal? It's local time zone right, so it's affected

PR #1314, opened 2026-02-05
SNOW-3051131: Bump AWSSDK.S3 to 4.0.18.2 to resolve GHSA-9cvc-h2w8-phrp

### LINKED ISSUE #1312: SNOW-3051131: Update AWSSDK.S3 dependency to avoid transitive dependency on vulnerable version of AWSSDK.Core
## What is the current behavior?
Currently, there is a dependency on AWSSDK.S3 set with a version of 4.0.4, but that causes a transitive dependency on AWSSDK.Core v4.0.0.14 which is known to have a security vulnerability.

## What is the desired behavior?
Updating the AWSSDK.S3 dependency to at 4.0.13.1 would then require the transitive dependency on AWSSDK.Core to be 4.0.3.3 which does not have a known vulnerability.   

## How would this improve `snowflake-connector-net`?
It allows developers to use the snowflake connector without having to manage this transitive dependency manually.

sfc-gh-daniszewski:
### Description
Update AWSSDK.S3 from 4.0.4 to 4.0.18.2, which transitively resolves
AWSSDK.Core to 4.0.3.12 (patched in 4.0.3.3+), addressing the low-severity
security advisory CVE-2026-22611 regarding improper region parameter validation.

Closes #1312

### Checklist
- [ ] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Snowflake.Data.csproj
+++ b/Snowflake.Data/Snowflake.Data.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net481;net8.0;net8.0-windows</TargetFrameworks>
     <Title>Snowflake.Data</Title>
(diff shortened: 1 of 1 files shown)


sfc-gh-dszmolka:
i think this can be paired with SNOW-3051131

PR #1345, opened 2026-04-23
SNOW-3423529 CI Stabilization effort

sfc-gh-mcepiga:
### Description
Please explain the changes you made here.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,25 @@ jobs:
       matrix:
         dotnet: ['net6.0', 'net7.0', 'net8.0', 'net9.0', 'net462', 'net471', 'net472', 'net48', 'net481']
         cloud_env: ['AZURE', 'GCP', 'AWS']
+        include:
+          - dotnet: net6.0
+            sdk: '6.0.x'
+          - dotnet: net7.0
+            sdk: '7.0.x'
+          - dotnet: net8.0
+            sdk: '8.0.x'
+          - dotnet: net9.0
+            sdk: '9.0.x'
+          - dotnet: net462
+            sdk: '6.0.x'
+          - dotnet: net471
+            sdk: '6.0.x'
+          - dotnet: net472
+            sdk: '6.0.x'
+          - dotnet: net48
+            sdk: '6.0.x'
+          - dotnet: net481
+            sdk: '6.0.x'
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -48,11 +67,7 @@ jobs:
       - name: Setup Dotnet
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: |
-            6.0.x
-            7.0.x
-            8.0.x
-            9.0.x
+          dotnet-version: ${{ matrix.sdk }}
           dotnet-quality: 'ga'
       - name: Setup Python
         uses: actions/setup-python@v5
@@ -102,11 +117,11 @@ jobs:
         with:
             token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
       - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v6
         with:
           # without the token code cov may fail because of Github limits https://github.com/codecov/codecov-action/issues/557
           token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
-          fail_ci_if_error: true
+          fail_ci_if_error: false # this is unstable, replace with true when https://github.com/codecov/codecov-action/issues/1940 is resolved
           files: Snowflake.Data.Tests\windows_${{ matrix.dotnet }}_${{ matrix.cloud_env }}_coverage.xml
 
   test-linux:
@@ -117,6 +132,15 @@ jobs:
       matrix:
         dotnet: ['net6.0', 'net7.0', 'net8.0', 'net9.0']
         cloud_env: ['AZURE', 'GCP', 'AWS']
+        include:
+          - dotnet: net6.0
+            sdk: '6.0.x'
+          - dotnet: net7.0
+            sdk: '7.0.x'
+          - dotnet: net8.0
+            sdk: '8.0.x'
+          - dotnet: net9.0
+            sdk: '9.0.x'
     steps:
       - uses: actions/checkout@v4
       - name: Setup Java
@@ -128,11 +152,7 @@ jobs:
       - name: Setup Dotnet
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: |
-            6.0.x
-            7.0.x
-            8.0.x
-            9.0.x
+          dotnet-version: ${{ matrix.sdk }}
           dotnet-quality: 'ga'
       - name: Setup Python
         uses: actions/setup-python@v5
@@ -181,11 +201,11 @@ jobs:
         with:
           token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
       - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v6
         with:
           # without the token code cov may fail because of Github limits https://github.com/codecov/codecov-action/issues/557
           token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
-          fail_ci_if_error: true
+          fail_ci_if_error: false # this is unstable, replace with true when https://github.com/codecov/codecov-action/issues/1940 is resolved
           files: Snowflake.Data.Tests/linux_${{ matrix.dotnet }}_${{ matrix.cloud_env }}_coverage.xml
 
   test-mac:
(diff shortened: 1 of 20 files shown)


.github/workflows/main.yml
```diff
@@ -102,11 +117,11 @@ jobs:
         with:
             token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
       - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v6
         with:
           # without the token code cov may fail because of Github limits https://github.com/codecov/codecov-action/issues/557
           token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
-          fail_ci_if_error: true
+          fail_ci_if_error: false # this is unstable, replace with true when https://github.com/codecov/codecov-action/issues/1940 is resolved
```
sfc-gh-daniszewski, line 124: can you create JIRA for this? so we don't forget :)

Snowflake.Data.Tests/IntegrationTests/TimeSensitiveFixtureIT.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static TestCaseData[] AsyncTestCases = GetTestCaseData(typeof(SFBaseTestAsync), typeof(SFBaseTest));
+        public static TestCaseData[] SyncTestCases = GetTestCaseData(typeof(SFBaseTest));
+
+        private static TestCaseData[] GetTestCaseData(Type assignableTo, Type notAssignableTo = null) => Assembly
+            .GetExecutingAssembly()
+            .GetTypes()
+            .Where(x => assignableTo.IsAssignableFrom(x) && (notAssignableTo == null || !notAssignableTo.IsAssignableFrom(x)))
+            .SelectMany(x => x.GetMethods(BindingFlags.Public | BindingFlags.Instance).Select(y => (x, y)))
+            .Where(x => x.Item2.GetCustomAttributes(typeof(TimeSensitiveAttribute), false).Any())
+            .Select(x => new TestCaseData(x).SetName(x.y.Name))
```
sfc-gh-daniszewski: If we'd have more than one test of the same name from different classes only one will be executed, right?

Snowflake.Data.Tests/IntegrationTests/TimeSensitiveFixtureIT.cs
```diff
@@ -0,0 +1,58 @@
...
+        public void TestSync((Type, MethodInfo) args)
+        {
+            // Arrange
+            var testClass = Activator.CreateInstance(args.Item1);
+            var startupMethod = args.Item1.GetMethods(BindingFlags.Public | BindingFlags.Instance).FirstOrDefault(x => x.GetCustomAttribute<SetUpAttribute>() != null);
+            var tearDownMethod = args.Item1.GetMethods(BindingFlags.Public | BindingFlags.Instance).FirstOrDefault(x => x.GetCustomAttribute<TearDownAttribute>() != null);
+
+            startupMethod?.Invoke(testClass, null);
+            args.Item2.Invoke(testClass, null);
+            tearDownMethod?.Invoke(testClass, null);
```
sfc-gh-daniszewski: teardown won't be called if Invoke throws

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
```diff
@@ -361,13 +376,20 @@ private void CopyIntoTable(SnowflakeDbConnection conn)
...
+                AssertOnColData(reader, 2);
                 CheckColumnMetadata(reader);
             }
         }
 
+        private static void AssertOnColData(DbDataReader reader, int index)
+        {
+            var actual = reader.GetString(index);
+            var substringLength = t_colData[index].Length > 20 ? 20 : t_colData[index].Length;
+            Assert.IsTrue(t_colData[index] == actual, $"Expected strings to be equal. Expected: '{t_colData[index].Substring(0, substringLength)}' [...], got '{actual}' instead.");
```
sfc-gh-daniszewski, line 390: isn't there something like `Assert.Equals` ?   
Also, in the failure message why you truncate expected string ?

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
```diff
@@ -1878,7 +1882,8 @@ public void TestHeartBeatWithConnectionPool()
         }
 
         [Test]
-        public void TestKeepAlive()
+        [TimeSensitive]
+        public async Task TestKeepAlive()
```
sfc-gh-daniszewski, line 1886: `SFConnectionIT` extends `SFBaseTest` - this test will be picked by SyncTestCase, right?

sfc-gh-daniszewski (commented):
I left some comments/questions, PTAL

.github/workflows/main.yml
sfc-gh-mcepiga: SNOW-3437365

Snowflake.Data.Tests/IntegrationTests/TimeSensitiveFixtureIT.cs
sfc-gh-mcepiga: No. Both of them would be executed, they'd just be labelled with the same name. (Which is not ideal, but trivially to fix with just naming them differently, so I decided against adding code handling it)

Snowflake.Data.Tests/IntegrationTests/TimeSensitiveFixtureIT.cs
sfc-gh-mcepiga: fair point, fix'd

Snowflake.Data.Tests/IntegrationTests/SFConnectionIT.cs
sfc-gh-mcepiga: good observation, whilst Sync -> Async would work, this way we'd get fire-and-forget situation. Will be fix'd.

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
sfc-gh-mcepiga: Assert Equals by default will print out the unequal operands, which I don't want here. I truncate explicitly, because the output is a lot of bytes that clutter the test execution log and are not exactly of any value to be read. See example: 
image
(initially I wanted to record screen scrolling through this, but it took ~3 minutes of scrolling)

Snowflake.Data.Tests/IntegrationTests/TimeSensitiveFixtureIT.cs
sfc-gh-daniszewski: > No. Both of them would be executed

👍 

> trivially to fix with just naming them differently

fix is trivial, but IMO detection isn't - how would you find out that you added a test with the same name? I sincerely doubt one would check it. Maybe we can add some precommit or CI rule.

I won't block on this, but I'm afraid it will come back and kick us at some point

Snowflake.Data.Tests/IntegrationTests/MaxLobSizeIT.cs
sfc-gh-daniszewski: exquisite verbosity :) OK, I get the intention

sfc-gh-daniszewski (approved):
LGTM, but please fix CI following before merging:
- add label indicating no changelog entry
- fix formatting

Snowflake.Data/Core/Revocation/CrlParser.cs
sfc-gh-boler, line 77: I think there is similar bug to what we had fixed in UD recently, where the leading 0x00 byte is not stripped leading to a failing check.

Snowflake.Data/Core/Revocation/CrlParser.cs
```diff
@@ -54,7 +55,7 @@ public Crl Create(X509Crl crl, DateTime now)
                 IssuerName = crl.IssuerDN.ToString(),
                 IssuerNameRawData = crl.IssuerDN.GetEncoded(),
                 IssuerDistributionPoints = ReadIdpFromCrl(crl),
-                RevokedCertificates = crl.GetRevokedCertificates().Select(cert => ConvertToHexadecimalString(cert.SerialNumber)).ToList(),
+                RevokedCertificates = crl.GetRevokedCertificates()?.Select(cert => ConvertToHexadecimalString(cert.SerialNumber)).ToList() ?? new List<string>(),
```
sfc-gh-boler, line 58: a hash set seems more appropriate, because the set of revoked certs can be pretty large.

Snowflake.Data/Core/Revocation/CrlParser.cs
sfc-gh-boler, line 82: I see we have ToUpper here, but do we normalize the lookup key too?

sfc-gh-boler (dismissed):
While not introduced here, we should fix the leading 0 byte issue.

Snowflake.Data/Core/Revocation/CrlParser.cs
sfc-gh-mcepiga: this item will be handled in SNOW-3445796

sfc-gh-mcepiga: this item will be handled in SNOW-3445796

Snowflake.Data/Core/Revocation/CrlParser.cs
sfc-gh-mcepiga: this item will be handled in SNOW-3445796

PR #1347, opened 2026-04-24
SNOW-3397010 390111 repooling

sfc-gh-mcepiga:
### Description
When the Snowflake server responds with vendor code 390111 ("session no longer exists"), the driver now detects this
   condition, throws a SnowflakeDbException with SFError.SESSION_GONE, and marks the session as invalid for pooling.
  Previously, sessions that the server had already terminated could be returned to the connection pool, causing
  subsequent consumers to fail on reuse.

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/Session/SFSession.cs
```diff
@@ -791,7 +790,7 @@ internal void InvalidateForPooling()
         {
             if (_invalidatedForPooling)
                 return;
-            logger.Info($"Session {sessionId} invalidated for pooling due to authentication failure.");
+            logger.Info($"Session {sessionId} invalidated for pooling due to authentication failure or session being gone.");
```
sfc-gh-mhofman: session destroyed? or session destroyed server-side (being gone sounds a bit weird and not precise)

sfc-gh-mhofman (approved):

Snowflake.Data/Core/Session/SFSession.cs
sfc-gh-mcepiga: changed to "servier destroyed server-side"

PR #1360, opened 2026-05-08
SNOW-3437480 replace out-of-process wiremock for unit tests with in-process .net version

sfc-gh-mcepiga:
Summary

  - Replaced the out-of-process Java-based WireMock (launched as a java -jar subprocess) with the in-process .NET
  library WireMock.Net (v2.2.0) for unit tests
  - Eliminated the Java runtime dependency, JAR download logic, startup polling, and the ca-cert.jks keystore file
  - Converted WiremockRunner to a sealed class using WireMockServer directly, with automatic free-port selection
  instead of hardcoded ports
  - Added typed deserialization models (WiremockModels/) to parse existing JSON mapping files and translate them into
  WireMock.Net fluent API calls
  - Updated all authenticator and HTTP unit tests to use the new runner's Url/SslUrl properties and simplified
  StringTransformations

  Test plan

  - All existing unit tests that use WiremockRunner pass without Java installed
  - OAuth, PAT, and redirect tests exercise both HTTP and HTTPS endpoints
  - Verify no port conflicts by running tests in parallel (ports are now dynamically allocated)

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/UnitTests/Authenticator/BaseOAuthFlowTest.cs
```diff
@@ -8,28 +8,42 @@
...
+        public void BaseBeforeAll()
+        {
+            Runner = WiremockRunner.NewWiremock();
+        }
+
+        [TearDown]
+        public void BaseAfterAll()
+        {
+            Runner.Stop();
+        }
```
sfc-gh-daniszewski: Isn't `SetUp` and `TearDown` called before/after **each** test? If so, `BeforeAll` is confusing

Snowflake.Data.Tests/Util/WiremockRunner.cs
```diff
@@ -1,257 +1,203 @@
...
+
+        public void Dispose() => Stop();
+
+        private static int FindFreePort()
+        {
+            var listener = new TcpListener(IPAddress.Loopback, 0);
+            listener.Start();
+            var port = ((IPEndPoint)listener.LocalEndpoint).Port;
+            listener.Stop();
+            return port;
```
sfc-gh-daniszewski, line 218: I think we might have TOCTOU issue here, at least in theory it might happen that at the time of check the port is free, but when WireMock binds it's already used. 
I'm wondering if we couldn't hardcode the ports to something with low collision chance.

Snowflake.Data.Tests/UnitTests/Tools/WiremockRunnerTest.cs
```diff
@@ -18,22 +18,10 @@ public class WiremockRunnerTest
...
-            _runner = WiremockRunner.NewWiremock();
-        }
-
         [SetUp]
         public void BeforeEach()
         {
-            _runner.ResetMapping();
-        }
-
-        [OneTimeTearDown]
```
sfc-gh-daniszewski: shouldn't we have `[Teardown]` since we have `[Setup]`?

Snowflake.Data.Tests/Util/WiremockModels/Mapping.cs
```diff
@@ -0,0 +1,22 @@
...
+        public MappingResponse Response { get; set; }
+
+        [JsonProperty("scenarioName")]
+        public string ScenarioName { get; set; }
+
+        [JsonProperty("requiredScenarioState")]
+        public string RequiredScenarioState { get; set; }
+
+        [JsonProperty("newScenarioState")]
+        public string NewScenarioState { get; set; }
```
sfc-gh-daniszewski: shouldn't the JSON properties be accordingly `whenStateIs` and `setStateTo` ?

sfc-gh-daniszewski (changes requested):
Great direction, but I think few things needs to be fixed before merging

Snowflake.Data.Tests/UnitTests/Authenticator/BaseOAuthFlowTest.cs
sfc-gh-mcepiga: fix'd

Snowflake.Data.Tests/UnitTests/Tools/WiremockRunnerTest.cs
sfc-gh-mcepiga: sure

Snowflake.Data.Tests/Util/WiremockModels/Mapping.cs
sfc-gh-mcepiga: Fix'd now

Snowflake.Data.Tests/Util/WiremockRunner.cs
```diff
@@ -16,250 +24,198 @@ public interface IWiremockRunner : IDisposable
...
-                        }
-                    }
-
-                    return runner;
-                }
-                retries++;
-                Thread.Sleep(RetryInterval);
+                Urls = [$"http://{Localhost}:{httpPort}"],
+                StartAdminInterface = true
+            });
```
sfc-gh-ddas, line 53: Can't this result in a race condition?

sfc-gh-ddas (approved):
One potential problem, please ignore if race is not possible on testing infra.

PR #1363, opened 2026-05-15
SNOW-3373902 Inbound Telemetry in dotnet driver

sfc-gh-mcepiga:
## Summary
  - Adds client-side telemetry based on `System.Diagnostics.Activity` (OTel-compatible) that instruments all command
  execution methods and sends structured telemetry to Snowflake via `/telemetry/send`
  - Introduces public `SnowflakeDbCommand.StartActivity()` API for custom client-defined activities with session
  context enrichment
  - Telemetry is controlled by `CLIENT_TELEMETRY_ENABLED` connection string property (defaults to `true`)

  ## Architecture
  - **ActivityStarter** — creates `Activity` instances from two sources: `Snowflake_dotnet_activity` (internal) and
  `Client_custom_activity` (public API)
  - **SnowflakeTelemetryModule** — static coordinator that owns a global `ActivityListener` and routes completed
  activities to per-session modules. Both internal and custom client's activities are forwarded to the server.
  - **SessionTelemetryModule** — per-session buffer + flush pipeline that converts activities to telemetry payloads
  and POSTs to `/telemetry/send` on timer tick, buffer threshold, or session close
  - **ActivityExtensions** — `SetSuccess()`, `SetException()`, `AddTelemetryEvent()` with secret masking and
  `SnowflakeDbException` error code extraction (including `AggregateException` unwrapping)

  ## Key design decisions
  - Telemetry is opt-out (enabled by default) via connection string property `CLIENT_TELEMETRY_ENABLED`
  - Custom activities (public API) are sent to Snowflake — they stay client-side on a separate activity source. They might be used by other  OTel subscribers as well.
  - Internal activities are session-scoped and filtered by session ID to prevent cross-session contamination in
  parallel test runs
  - Flush is triggered by buffer size threshold, periodic timer, or session close (dispose)

  ## Test plan
  - [x] Unit tests: `ActivityStarterTest` — activity creation, success/error status, exception unwrapping, secret
  masking, null safety
  - [x] Unit tests: `SessionTelemetryModuleTest` — buffer/flush lifecycle, auto-flush threshold, periodic flush,
  disable on server error, concurrent flush guard
  - [x] Unit tests: `SnowflakeTelemetryModuleTest` — session registration/unregistration, activity routing, token
  refresh
  - [x] Wiremock Integration Tests: `ClientTelemetryWiremockTest` — full pipeline through wiremock verifying sync/async commands,
  session tags, driver metadata, auth token, error status, disabled telemetry, public StartActivity API with nested
  activities
  - [x] Integration tests: `SFDbCommandIT` telemetry test cases — verify against real Snowflake with
  session-ID-scoped activity capture

something is going through
image

### DIFF
--- a/Snowflake.Data/Client/SnowflakeDbCommand.cs
+++ b/Snowflake.Data/Client/SnowflakeDbCommand.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Snowflake.Data.Log;
+using Snowflake.Data.Telemetry;
 
 namespace Snowflake.Data.Client
 {
@@ -161,78 +162,135 @@ namespace Snowflake.Data.Client
 
         public override int ExecuteNonQuery()
         {
-            logger.Debug($"ExecuteNonQuery");
-            SFBaseResultSet resultSet = ExecuteInternal();
-            long total = 0;
-            do
+            logger.Debug("ExecuteNonQuery");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteNonQuery);
+            try
             {
-                if (resultSet.IsDQL()) continue;
-                int count = resultSet.CalculateUpdateCount();
-                if (count < 0)
+                SFBaseResultSet resultSet = ExecuteInternal();
+                long total = 0;
+                do
                 {
-                    // exceeded max int, return -1
-                    return -1;
+                    if (resultSet.IsDQL())
+                    {
+                        activity?.AddTelemetryEvent(TelemetryEvents.DqlResultSetSkipped);
+                        continue;
+                    }
+                    int count = resultSet.CalculateUpdateCount();
+                    if (count < 0)
+                    {
+                        // exceeded max int, return -1
+                        activity?.AddTelemetryEvent(TelemetryEvents.RowCountNegative);
+                        return -1;
+                    }
+                    total += count;
+                    if (total > int.MaxValue)
+                    {
+                        activity?.AddTelemetryEvent(TelemetryEvents.RowCountOverflow);
+                        return -1;
+                    }
                 }
-                total += count;
-                if (total > int.MaxValue)
-                {
-                    return -1;
-                }
-            }
-            while (resultSet.NextResult());
+                while (resultSet.NextResult());
 
-            return (int)total;
+                activity?.SetSuccess();
+                return (int)total;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         public override async Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
         {
-            logger.Debug($"ExecuteNonQueryAsync");
-            cancellationToken.ThrowIfCancellationRequested();
-
-            var resultSet = await ExecuteInternalAsync(cancellationToken).ConfigureAwait(false);
-            long total = 0;
-            do
+            logger.Debug("ExecuteNonQueryAsync");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteNonQueryAsync);
+            try
             {
-                if (resultSet.IsDQL()) continue;
-                int count = resultSet.CalculateUpdateCount();
-                if (count < 0)
-                {
-                    // exceeded max int, return -1
-                    return -1;
-                }
-                total += count;
-                if (total > int.MaxValue)
-                {
-                    return -1;
-                }
-            }
-            while (await resultSet.NextResultAsync(cancellationToken).ConfigureAwait(false));
+                cancellationToken.ThrowIfCancellationRequested();
 
-            return (int)total;
+                var resultSet = await ExecuteInternalAsync(cancellationToken).ConfigureAwait(false);
+                long total = 0;
+                do
+                {
+                    if (resultSet.IsDQL())
+                    {
+                        activity?.AddTelemetryEvent(TelemetryEvents.DqlResultSetSkipped);
+                        continue;
+                    }
+                    int count = resultSet.CalculateUpdateCount();
+                    if (count < 0)
+                    {
+                        activity?.AddTelemetryEvent(TelemetryEvents.RowCountNegative);
+                        return -1;
+                    }
+                    total += count;
+                    if (total > int.MaxValue)
+                    {
+                        activity?.AddTelemetryEvent(TelemetryEvents.RowCountOverflow);
+                        return -1;
+                    }
+                }
+                while (await resultSet.NextResultAsync(cancellationToken).ConfigureAwait(false));
+
+                activity?.SetSuccess();
+                return (int)total;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         public override object ExecuteScalar()
         {
-            logger.Debug($"ExecuteScalar");
-            SFBaseResultSet resultSet = ExecuteInternal();
+            logger.Debug("ExecuteScalar");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteScalar);
+            try
+            {
+                var resultSet = ExecuteInternal();
 
-            if (resultSet.Next())
-                return resultSet.GetValue(0);
-            else
-                return DBNull.Value;
+                object result;
+                if (resultSet.Next())
+                    result = resultSet.GetValue(0);
+                else
+                    result = DBNull.Value;
+
+                activity?.SetSuccess();
+                return result;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         public override async Task<object> ExecuteScalarAsync(CancellationToken cancellationToken)
         {
-            logger.Debug($"ExecuteScalarAsync");
-            cancellationToken.ThrowIfCancellationRequested();
+            logger.Debug("ExecuteScalarAsync");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteScalarAsync);
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
 
-            var result = await ExecuteInternalAsync(cancellationToken).ConfigureAwait(false);
+                var resultSet = await ExecuteInternalAsync(cancellationToken).ConfigureAwait(false);
 
-            if (await result.NextAsync().ConfigureAwait(false))
-                return result.GetValue(0);
-            else
-                return DBNull.Value;
+                object result;
+                if (await resultSet.NextAsync().ConfigureAwait(false))
+                    result = resultSet.GetValue(0);
+                else
+                    result = DBNull.Value;
+
+                activity?.SetSuccess();
+                return result;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         /// <summary>
@@ -259,21 +317,34 @@ namespace Snowflake.Data.Client
 
         protected override DbDataReader ExecuteDbDataReader(CommandBehavior behavior)
         {
-            logger.Debug($"ExecuteDbDataReader");
-            SFBaseResultSet resultSet = ExecuteInternal();
-            return new SnowflakeDbDataReader(this, resultSet);
+            logger.Debug("ExecuteDbDataReader");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteDbDataReader);
+            try
+            {
+                SFBaseResultSet resultSet = ExecuteInternal();
+                activity?.SetSuccess();
+                return new SnowflakeDbDataReader(this, resultSet);
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         protected override async Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
         {
-            logger.Debug($"ExecuteDbDataReaderAsync");
+            logger.Debug("ExecuteDbDataReaderAsync");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteDbDataReaderAsync);
             try
             {
                 var result = await ExecuteInternalAsync(cancellationToken).ConfigureAwait(false);
+                activity?.SetSuccess();
                 return new SnowflakeDbDataReader(this, result);
             }
             catch (Exception ex)
             {
+                activity?.SetException(ex);
                 logger.Error("The command failed to execute.", ex);
                 throw;
             }
@@ -286,9 +357,19 @@ namespace Snowflake.Data.Client
         /// <returns>The query id.</returns>
         public string ExecuteInAsyncMode()
         {
-            logger.Debug($"ExecuteInAsyncMode");
-            SFBaseResultSet resultSet = ExecuteInternal(asyncExec: true);
-            return resultSet.queryId;
+            logger.Debug("ExecuteInAsyncMode");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteInAsyncMode);
+            try
+            {
+                SFBaseResultSet resultSet = ExecuteInternal(asyncExec: true);
+                activity?.SetSuccess();
+                return resultSet.queryId;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         /// <summary>
@@ -299,9 +380,19 @@ namespace Snowflake.Data.Client
         /// <returns>The query id.</returns>
         public async Task<string> ExecuteAsyncInAsyncMode(CancellationToken cancellationToken)
         {
-            logger.Debug($"ExecuteAsyncInAsyncMode");
-            var resultSet = await ExecuteInternalAsync(cancellationToken, asyncExec: true).ConfigureAwait(false);
-            return resultSet.queryId;
+            logger.Debug("ExecuteAsyncInAsyncMode");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.ExecuteAsyncInAsyncMode);
+            try
+            {
+                var resultSet = await ExecuteInternalAsync(cancellationToken, asyncExec: true).ConfigureAwait(false);
+                activity?.SetSuccess();
+                return resultSet.queryId;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         /// <summary>
@@ -311,8 +402,19 @@ namespace Snowflake.Data.Client
         /// <returns>The query status.</returns>
         public QueryStatus GetQueryStatus(string queryId)
         {
-            logger.Debug($"GetQueryStatus");
-            return _queryResultsAwaiter.GetQueryStatus(connection, queryId);
+            logger.Debug("GetQueryStatus");
+            using var activity = connection?.SfSession?.StartActivity(TelemetryActivities.GetQueryStatus);
+            try
+            {
+                var status = _queryResultsAwaiter.GetQueryStatus(connection, queryId);
+                activity?.SetSuccess();
+                return status;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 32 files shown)


Snowflake.Data/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,68 @@
...
+internal static class TelemetryEvents
+{
+    internal const string RowCountOverflow = "RowCountOverflow";
+    internal const string RowCountNegative = "RowCountNegative";
+    internal const string DqlResultSetSkipped = "DqlResultSetSkipped";
+}
+
+internal static class TelemetryTransport
+{
+    internal const int TelemetryFlushSize = 10;
```
sfc-gh-pbulawa: Isn't 10 too small? I imagine that we will hit that very often causing unnecessary network transports.

Snowflake.Data/Telemetry/ActivityStarter.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <returns>A started <see cref="Activity"/>, or <c>null</c> if no listener is sampling.</returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown when the connection is not open or client telemetry is not enabled.
+    /// </exception>
+    public static Activity StartActivity(this SnowflakeDbCommand command, string name)
+    {
+        if (command?.Connection is not SnowflakeDbConnection connection || !connection.IsOpen())
+            throw new ArgumentException($"{nameof(StartActivity)} can only be called on open connections!");
+
+        if (!connection.SfSession.IsClientTelemetryEnabled())
```
sfc-gh-pbulawa, line 40: The ActivityStarter could be a no-op implementation when telemetry is disabled. This would allow us to skip the `try / catch` blocks when used with the changes around the main implementation.

Snowflake.Data/Telemetry/ActivityStarter.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </param>
+    /// <param name="name">The operation name for the activity.</param>
+    /// <returns>A started <see cref="Activity"/>, or <c>null</c> if no listener is sampling.</returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown when the connection is not open or client telemetry is not enabled.
+    /// </exception>
+    public static Activity StartActivity(this SnowflakeDbCommand command, string name)
+    {
+        if (command?.Connection is not SnowflakeDbConnection connection || !connection.IsOpen())
+            throw new ArgumentException($"{nameof(StartActivity)} can only be called on open connections!");
```
sfc-gh-pbulawa, line 38: IMO we should aim for the `Activity` to be a silent-fail with debug message, as it is not crucial for the main flow.

Snowflake.Data/Telemetry/ActivityExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+#endif
+        activity.SetTag(TelemetryTags.StatusCode, "ERROR");
+        activity.SetTag(TelemetryTags.StatusDescription, description);
+
+        if (exception != null)
+        {
+            var tagsCollection = new ActivityTagsCollection
+            {
+                { TelemetryTags.Exception, exception.GetType().FullName },
+                { TelemetryTags.ExceptionMessage, description },
```
sfc-gh-pbulawa: This is tricky. We cannot guarantee that customer-restricted data is filtered out 100% reliably from the message. I'd stick with the exception name only.

Snowflake.Data/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,68 @@
...
+    internal const int TelemetryFlushIntervalInMs = 1000 * 60;
```
sfc-gh-pbulawa: Can we make it configurable by the user?

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+namespace Snowflake.Data.Telemetry;
+
+internal sealed class SessionTelemetryModule : ISessionTelemetryModule
+{
+    private static readonly SFLogger s_logger = SFLoggerFactory.GetSFLogger<SessionTelemetryModule>();
+
+    private static readonly int s_flushSize = TelemetryTransport.TelemetryFlushSize;
+    private static readonly TimeSpan s_flushInterval = TimeSpan.FromMilliseconds(TelemetryTransport.TelemetryFlushIntervalInMs);
+    private static readonly TimeSpan s_sendTimeout = TimeSpan.FromSeconds(30);
+    private static readonly TimeSpan s_sendHttpTimeout = TimeSpan.FromSeconds(10);
```
sfc-gh-pbulawa: Let's use the timeouts we have available for the session connections.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+
+    internal void UpdateToken(string sessionToken) => _sessionToken = sessionToken;
+
+    internal void OnActivityStoppedImpl(Activity activity)
+    {
+        try
+        {
+            if (!IsTelemetryEnabled())
+            {
+                s_logger.Debug($"Recorded activity, but client telemetry is disabled. Activity will be ignored.");
```
sfc-gh-pbulawa: Wouldn't this cause a lot of noise in the logs? IMO it's not that important, consider removing it or switching to `TRACE`

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            s_logger.Debug($"Failed to process activity telemetry for session {_sessionId}: {ex.Message}.");
+        }
+    }
+
+    internal async Task FlushAsync(CancellationToken cancellationToken)
+    {
+        if (!IsTelemetryEnabled())
+        {
+            s_logger.Debug("Trying to flush telemetry, but client telemetry is disabled.");
```
sfc-gh-pbulawa: Same as above, logging this provides no actual value. Consider removing it or switching to `TRACE`.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        try
+        {
+            var request = BuildSendRequest(toSend);
+            var response = await _restRequester.PostAsync<NullDataResponse>(request, cancellationToken).ConfigureAwait(false);
+
+            if (response is not { success: true })
+            {
+                s_logger.Warn($"Non-success response from telemetry server for session {_sessionId}. Disabling telemetry.");
```
sfc-gh-pbulawa, line 132: `Info` or even `Debug`?

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+                DisableTelemetry();
+                return;
+            }
+
+            s_logger.Debug($"Successfully sent {toSend.Count} telemetry logs for session {_sessionId}.");
+        }
+        catch (Exception ex)
+        {
+            s_logger.Warn($"Failed to send telemetry for session {_sessionId}: {ex.Message}. Disabling telemetry.");
```
sfc-gh-pbulawa, line 141: `Info` or even `Debug`?

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+            DisableTelemetry();
+        }
+    }
+
+    private void Flush(bool isCallingFromDispose = false)
+    {
+        if (!_isServiceAvailable || (!isCallingFromDispose && IsDisposed()))
+        {
+            s_logger.Debug("Trying to flush telemetry, but client telemetry is disabled.");
```
sfc-gh-pbulawa: Logging this provides no actual value. Consider removing it or switching to `TRACE`.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        try
+        {
+            var request = BuildSendRequest(toSend);
+            var response = _restRequester.Post<NullDataResponse>(request);
+
+            if (response is not { success: true })
+            {
+                s_logger.Warn($"Non-success response from telemetry server for session {_sessionId}. Disabling telemetry.");
```
sfc-gh-pbulawa, line 171: I don't think that it's a good idea to disable the telemetry when we don't receive a success message. This is more like a fire and forget, as the endpoint may have transient issue and when it will be back, we won't send anything.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+                DisableTelemetry();
+                return;
+            }
+
+            s_logger.Debug($"Successfully sent {toSend.Count} telemetry logs for session {_sessionId}.");
+        }
+        catch (Exception ex)
+        {
+            s_logger.Warn($"Failed to send telemetry for session {_sessionId}: {ex.Message}. Disabling telemetry.");
```
sfc-gh-pbulawa, line 180: `Info` or even `Debug`?

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        if (!IsDisposed())
+            _flushTimer.Change(Timeout.Infinite, Timeout.Infinite);
+    }
+
+    private void FlushTimerCallback()
+    {
+        if (!IsTelemetryEnabled())
+        {
+            s_logger.Debug($"Timer tries to flush client telemetry, but client telemetry is disabled.");
```
sfc-gh-pbulawa: Logging this provides no actual value. Consider removing it or switching to `TRACE`.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
```diff
@@ -0,0 +1,297 @@
...
+            return;
+        }
+
+        try
+        {
+            Flush();
+        }
+        catch (Exception ex)
+        {
+            s_logger.Warn($"Periodic flush failed for session {_sessionId}: {ex.Message}.");
```
sfc-gh-pbulawa, line 236: `Info` or even `Debug`?

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
```diff
@@ -0,0 +1,153 @@
...
+            Sample = Sample,
+        };
+        ActivitySource.AddActivityListener(listener);
+    }
+
+    internal static void Register(SFSession session)
+    {
+        if (!session.IsClientTelemetryEnabled())
+        {
+            s_logger.Debug("Session tried to register to telemetry module, but it has client telemetry disabled!");
```
sfc-gh-pbulawa: Logging this provides no actual value. Consider removing it or switching to `TRACE`.

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
```diff
@@ -0,0 +1,153 @@
...
+        module = concreteModule;
+        return true;
+    }
+
+    private static bool TryRemoveSessionModule(string sessionId, out SessionTelemetryModule module)
+    {
+        module = null;
+        if (string.IsNullOrEmpty(sessionId))
+        {
+            s_logger.Warn("Tried to unregister session without providing any id!");
```
sfc-gh-pbulawa: `Debug`?

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
```diff
@@ -0,0 +1,153 @@
...
+        if (!s_activitySources.Contains(activity.Source.Name))
+        {
+            s_logger.Debug($"Activity {activity.Source.Name} will not be handled by this component!");
+            return;
+        }
+
+        var sessionId = activity.GetTagItem(TelemetryTags.SessionId) as string;
+        if (string.IsNullOrEmpty(sessionId))
+        {
+            s_logger.Warn($"Activity with no {TelemetryTags.SessionId} recorded!");
```
sfc-gh-pbulawa: `Debug`

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
```diff
@@ -0,0 +1,153 @@
...
+            return;
+        }
+
+        SessionTelemetryModule module;
+        lock (s_lock)
+        {
+            if (!s_sessions.TryGetValue(sessionId, out module))
+            {
+                s_logger.Warn($"Activity with no matching session telemetry module recorded! SessionId: {sessionId}");
```
sfc-gh-pbulawa: `Debug`

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        module.OnActivityStoppedImpl(activity);
+    }
+
+    private static void OnActivityStarted(Activity activity)
+    {
+        var sessionId = activity.GetTagItem(TelemetryTags.SessionId) as string;
+        s_logger.Info($"Activity {activity.DisplayName} for session: {sessionId} started..");
```
sfc-gh-pbulawa: `Debug`

Snowflake.Data/Client/SnowflakeDbCommand.cs
```diff
@@ -161,78 +162,135 @@ public override void Cancel()
...
+                while (resultSet.NextResult());
 
-            return (int)total;
+                activity?.SetSuccess();
+                return (int)total;
+            }
+            catch (Exception ex)
+            {
+                activity?.SetException(ex);
+                throw;
```
sfc-gh-pbulawa, line 200: The telemetry should not fail any action.

sfc-gh-pbulawa (changes requested):

Snowflake.Data/Telemetry/TelemetryConstants.cs
sfc-gh-mcepiga: what about 100?

Snowflake.Data/Telemetry/ActivityStarter.cs
sfc-gh-mcepiga: I don't think I fully understand, could you elaborate on this?

Snowflake.Data/Telemetry/ActivityStarter.cs
sfc-gh-mcepiga: this method is intended to be used by client. For the client's purposes it's better to have it complain loud and clear something is not supported instead of silently swallow configuration mishaps. I agree with the sentiment in terms of activities started by the driver internally - and the internal method driver uses won't throw (unless provided with empty-name activity, which is not supported, i.e. truly an exception)

Snowflake.Data/Telemetry/ActivityExtensions.cs
sfc-gh-mcepiga: fair, there's little value in handling it and some risks associated. I'll adjust this per your suggestion.

Snowflake.Data/Telemetry/TelemetryConstants.cs
sfc-gh-mcepiga: Do we need this to be that configurable? That could be considered future extension if needed.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: SendTimeout is now shared between this and session. Http timeout however has different settings in different code paths (okta idp has 16 seconds, other okta call has no timeout, chunk download has 1 hour, MFA call has 60 seconds etc). I decied to drop the second param and rely on the default RestRequest value- let me know if you have other perspective on this

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: depends on what "DEBUG" is supposed to represent, but generally I agree. Changed to trace.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: fix'd

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: This has potentially long-lasting impact, as telemetry for this session will be disabled and there's no self-heal. While this is negligible in terms of the basic driver health it is a case of something going wrong and should be visible to the client. If they would consider that to be polluting the logs, they can define logging level for this logger to hide warnings.

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: as above

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: changed to trace

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: This was modeled after JDBC driver and how they react to the telemetry failure. Underlying HttpClient has retry handler that should handle truly transient issues (unless configured to not use retries, but that's fine).

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: see above

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: Changed to TRACE

Snowflake.Data/Telemetry/SessionTelemetryModule.cs
sfc-gh-mcepiga: as above

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
sfc-gh-mcepiga: changed to TRACE

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
sfc-gh-mcepiga: okay

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
sfc-gh-mcepiga: okay

Snowflake.Data/Telemetry/SnowflakeTelemetryModule.cs
sfc-gh-mcepiga: okay

Snowflake.Data/Client/SnowflakeDbCommand.cs
sfc-gh-mcepiga: this try-catch block was added so that we can catch the exception and add it to telemetry and re-throw it in case it would be thrown, regardless if there is or there isn't a telemetry setup. Previously exception would also be thrown, it'd just be unhandled here.

doc/Client/Telemetry.md
```diff
@@ -0,0 +1,110 @@
...
+
+// Start a custom activity — requires an open connection with telemetry enabled
+using var activity = command.StartActivity("MyCustomOperation");
+
+// Add your own tags
+activity?.SetTag("app.module", "billing");
+activity?.SetTag("app.batch_size", "500");
+
+// Add named events within the activity
+activity?.AddTelemetryEvent("ValidationComplete");
```
sfc-gh-jmartinezgaray, line 60: Thanks for adding this! It would help us to have a cleaner implementation + an abstraction we can rely on when moving to Solscite soon.

I would like to point out that some events can be more complex than others, which may require adding more metadata to the payload.

I think the `AddTelemetryEvent()` signature is pretty good for "heartbeat" events, but I would suggest we overload the signature to support a more detailed payload

```C#
public void AddTelemetryEvent(string eventName, object payload);
// Or
public void AddTelemetryEvent(object payload);

```

Also, I would love to see how the final payload is built before sending it to the endpoint 👀 

Is it possible to include it in the markdown?

Let me know if this makes sense, thanks!

sfc-gh-mcepiga: @sfc-gh-jmartinezgaray thank you for your input. Telemetry.md file has been updated with more examples and details, alongside extended api for more complex events.

Snowflake.Data/Telemetry/TelemetryConstants.cs
sfc-gh-pbulawa: In case if telemetry will cause slowdown in some environment it would be easier to debug/mitigate with balancing the params than preparing a new version. IMO this should be configurable by the user.

sfc-gh-pbulawa (changes requested):
The Telemetry config parameters should be configurable by the user.

sfc-gh-mcepiga:
> The Telemetry config parameters should be configurable by the user.

done

sfc-gh-pbulawa (approved):

PR #1364, opened 2026-05-18
SNOW-3422038 Add Retry from transient errors when sf returns invalid, truncated response

sfc-gh-mcepiga:
Summary

  - Add a single retry when Snowflake returns HTTP 200 with a truncated/invalid JSON body (transient server-side
  issue observed in production)
  - Extract DeserializeResponseAsync<T> helper and centralize JsonSerializer instance in JsonUtils
  - Fix Wiremock test mapping files (missing scenario state, missing mappings wrapper)

  Changes

  - RestRequester.SendAsync<T> catches JsonReaderException on first deserialization attempt and re-sends the request
  once before propagating
  - JsonUtils.Serializer — shared JsonSerializer static field (thread-safe for reads) replaces per-call construction
  - New unit tests (Moq-based) covering: valid JSON without retry, retry-then-succeed, both-fail-throws for both GET
  and POST
  - New Wiremock integration tests covering full end-to-end retry through SnowflakeDbConnection
  - Minor cleanup: var usage in mocks, field rename _HttpClient → HttpClient

  Test plan

  - RestRequesterTest — 10 unit tests pass (net6.0, net10.0)
  - RestRequesterWiremockTest — 2 integration tests pass (net6.0, net10.0)
  - Verify no regression in existing integration test suite
  - Confirm retry behavior under real transient conditions (manual/staging)

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Changelog
 - v5.6.0
+    - Bug fix: Fixed handling of transient server issues resulting in sending truncated JSON response.
     - Added .NET 10 support. Changed LangVersion to C#13.
     - Added `DbType.AnsiStringFixedLength` to the set of types mapped to Snowflake `TEXT`, matching existing support for `AnsiString`, `String`, and `StringFixedLength`.
     - Extended login-request telemetry with libc detection (`LIBC_FAMILY`, `LIBC_VERSION`). On Linux, the driver now reports whether the runtime uses glibc and includes the library version.
--- a/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
@@ -81,7 +81,7 @@ namespace Snowflake.Data.Tests.UnitTests
 
         private string BuildConnectionString()
         {
-            var uri = new Uri(_runner.Url);
+            var uri = new Uri(_runner.WiremockBaseHttpsUrl);
             return new StringBuilder()
                 .Append("account=")
                 .Append("testaccount")
--- a/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
@@ -81,7 +81,7 @@ namespace Snowflake.Data.Tests.UnitTests
 
         private string BuildConnectionString()
         {
-            var uri = new Uri(_runner.WiremockBaseHttpsUrl);
+            var uri = new Uri(_runner.WiremockBaseHttpUrl);
             return new StringBuilder()
                 .Append("account=")
                 .Append("testaccount")
--- a/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
@@ -1,0 +1,102 @@
+using System;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using Snowflake.Data.Client;
+using Snowflake.Data.Tests.Util;
+
+namespace Snowflake.Data.Tests.UnitTests
+{
+    [TestFixture, NonParallelizable]
+    public sealed class RestRequesterWiremockTest
+    {
+        private static readonly string s_mappingPath = Path.Combine("wiremock", "RestRequester");
+        private static readonly string s_loginMapping = Path.Combine(s_mappingPath, "login_success.json");
+        private static readonly string s_queryTruncatedThenValid = Path.Combine(s_mappingPath, "query_truncated_then_valid.json");
+        private static readonly string s_queryTruncatedAlways = Path.Combine(s_mappingPath, "query_truncated_always.json");
+
+        private WiremockRunner _runner;
+
+        [OneTimeSetUp]
+        public void BeforeAll()
+        {
+            _runner = WiremockRunner.NewWiremock();
+        }
+
+        [SetUp]
+        public void BeforeEach()
+        {
+            _runner.ResetMapping();
+        }
+
+        [OneTimeTearDown]
+        public void AfterAll()
+        {
+            _runner.Stop();
+        }
+
+        [Test]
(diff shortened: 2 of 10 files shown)


Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using Snowflake.Data.Client;
+using Snowflake.Data.Tests.Util;
+
+namespace Snowflake.Data.Tests.UnitTests
```
sfc-gh-daniszewski, line 11: Is this unit test if it uses wiremock?

Snowflake.Data/Core/RestRequester.cs
```diff
@@ -54,49 +56,50 @@ public T Post<T>(IRestRequest request)
...
+            // in rare instances server may return invalid response body (e.g. truncated one), whilst returning 2XX response. Those are mostly transient. See SNOW-3422038 for further details.
+            using var retryResponse = await SendAsync(method, request, cancellationToken).ConfigureAwait(false);
+            return await DeserializeResponseAsync<T>(retryResponse, cancellationToken).ConfigureAwait(false);
+        }
+
+        private Task<HttpResponseMessage> SendAsync(HttpMethod method,
+                                                          IRestRequest request,
+                                                          CancellationToken externalCancellationToken)
+        {
+            var message = request.ToRequestMessage(method);
```
sfc-gh-daniszewski, line 101:
```suggestion
            using var message = request.ToRequestMessage(method);
```
otherwise we have leaked resource for unsuccessful calls.

Snowflake.Data/Core/RestRequester.cs
```diff
@@ -179,6 +163,29 @@ internal static bool HasUnauthorizedStatusCode(Exception ex)
...
+            return uri.ToString();
+#endif
+        }
+
+        private static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage response, CancellationToken cancellationToken)
+        {
+            #if NET6_0_OR_GREATER
+            await using var stream = await response.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(false);
+            using var streamReader = new StreamReader(stream);
+            await using var jsonReader = new JsonTextReader(streamReader);
```
sfc-gh-daniszewski, line 180: is this branch covered in tests? do `JsonTextReader` doesn't implement `IAsyncDisposable`?

Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
```diff
@@ -0,0 +1,102 @@
...
+            conn.ConnectionString = BuildConnectionString();
+            conn.Open();
+
+            using var cmd = conn.CreateCommand();
+            cmd.CommandText = "SELECT 1";
+
+            // act & assert
+            Assert.ThrowsAsync<JsonReaderException>(
+                () => cmd.ExecuteReaderAsync(CancellationToken.None));
+        }
```
sfc-gh-daniszewski, line 80: Is it possible that `JsonReaderException` will be wrapped into something else within `ExecuteReaderAsync` ?

sfc-gh-daniszewski (changes requested):
One sanity question - there is no back-off for the first retry, this is intentional, right?

Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
sfc-gh-mcepiga: that's an excellent question that was answered before my time by e.g. RedirectUnitTest.

Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
sfc-gh-mcepiga: unless the implementation changes, no.

dotnet async state machine unwinds exception accordingly, so unless we do something like Task.ContinueWith(...) (like we do in e.g. session component), we'll get the JsonReaderException here.

Snowflake.Data/Core/RestRequester.cs
sfc-gh-mcepiga: IAsyncDisposable was introduced later, so .net framework has no IAsyncDisposable by default. This branch is covered by tests on platforms using .net >= 6

Snowflake.Data.Tests/UnitTests/RestRequesterWiremockTest.cs
sfc-gh-daniszewski: ok, let's sort this later

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 # Changelog
 - v5.6.0
+    - Bug fix: Fixed handling of transient server issues resulting in sending truncated JSON response.
```
sfc-gh-daniszewski, line 5: forgot earlier - pls move bug fixes after new features

Snowflake.Data/Core/RestRequester.cs
sfc-gh-mcepiga: see line 136.

If we'd add using here, we'd end up breaking the component, cause when we'd wrap up unfinished task from SendAsync (likely at starting the web call), we'd dispose the request message which is used internally to load up the body.

there is a code path in which we don't explicitly dispose the message - when the response throws, but from the perspective of this component, that's unhandled exception, so there's a lot of extra resources allocation then and functional case to worry about, and the message will get picked up by the gc eventually.

sfc-gh-mcepiga:
> One sanity question - there is no back-off for the first retry, this is intentional, right?

yes, it was considered to be transient issue by the server that should not be idempotent, so just the overhead of network roundtrip is considered to be enough of the back-off here.

CHANGELOG.md
sfc-gh-mcepiga: done

Snowflake.Data/Core/RestRequester.cs
sfc-gh-daniszewski: I see, thx!

sfc-gh-daniszewski (approved):
LGTM

PR #1369, opened 2026-05-22
SNOW-3445796 CRL handling minor improvements

sfc-gh-mcepiga:
### Description
Summary

  - Changed Crl.RevokedCertificates from List<string> to HashSet<string> with OrdinalIgnoreCase comparer for O(1)
  case-insensitive revocation lookups
  - Modernized ConvertToHexadecimalString to use Convert.ToHexString on .NET 6+ (with fallback for older TFMs)
  - Added unit tests covering case-insensitive serial matching and DER leading-zero-byte padding consistency between
  BouncyCastle and .NET = shows we don't face 0x00 crl padding issue that happened in UD 

  Test plan

  - TestIsRevokedIsCaseInsensitive — verifies uppercase/lowercase/mixed hex serial numbers all match
  - TestIsRevokedHandlesLeadingZeroBytePadding — proves BouncyCastle BigInteger.ToByteArray() and
  X509Certificate2.SerialNumber produce identical hex for serials with and without high-bit-set padding
  - TestConvertBigIntegerToHexString — data-driven boundary cases (1, 127, 128, 255, 256, 32768) including sign-byte
  edge cases
  - Existing CRL tests continue to pass

### Checklist
- [x] Code compiles correctly


Snowflake.Data.Tests/UnitTests/Revocation/CrlTest.cs
```diff
@@ -128,5 +137,92 @@ public void TestConvertBigIntegerToHexString(string stringValue, string expected
...
+            // Generate certs until we get one with/without the 0x00 prefix (depending on test case).
+            // CreateSelfSigned uses a random 64-bit serial; roughly half will have the high bit set.
+            X509Certificate2 certificate;
+            var expectedBytesCount = serialHasHighBitSet ? 9 : 8;
+            var counter = 0;
+            do
+            {
+                certificate = BuildSelfSignedCertificate(counter);
+                Assert.InRange(counter++, int.MinValue, 20);
+            } while (certificate.SerialNumberBytes.Length != expectedBytesCount); // Expected value of number of loop iterations here = 2
```
sfc-gh-boler, line 186: Could we somehow create a cert with our own serial? That would avoid this expensive and non-deterministic loop.

Snowflake.Data.Tests/UnitTests/Revocation/CrlTest.cs
```diff
@@ -128,5 +137,92 @@ public void TestConvertBigIntegerToHexString(string stringValue, string expected
...
+            // assert
+            Assert.True(isRevoked);
+        }
+
+        [SFTheory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void TestIsRevokedHandlesLeadingZeroBytePadding(bool serialHasHighBitSet)
+        {
+#if !NET8_0_OR_GREATER
```
sfc-gh-boler, line 166: This trailing curly bracket makes it a bit hard to read. Could we use the preprocesor to create a bool that's then passed to Sip.When?

sfc-gh-boler (approved):
lgtm, but I'm curious if we can improve the two areas I commented on.

Snowflake.Data.Tests/UnitTests/Revocation/CrlTest.cs
sfc-gh-mcepiga: create it easily through test code? It's not possible.

We could avoid this non-deterministic approach by e.g. having a hardcoded cer, we could've generated using this loop here or by any other means. We could then continue with the rest of the logic of the test. 

This however introduces layer of obfuscation - what is exactly in this hardcoded cert, where did it come from, why is it like this, if something changes in logic, is is still valid to use or maybe we should provide a different one, if so - how to create one. Given this is rather small corner case, which we go with the default implementation and test shows we *don't* have a problem we faced in other drivers, I decided against complicating this further than this. 

Let me know if you have other perspective!

Snowflake.Data.Tests/UnitTests/Revocation/CrlTest.cs
sfc-gh-mcepiga: Well, we can't really, as the older TFM has different contract, so if preprocessor directive covered only bool flag, the rest of the test method, even though it's unreachable during runtime, would fail to compile.

We can however move this curly bracket down, so it's outside of the preprocessor if-else, and it will improve readability. It's done now.

sfc-gh-boler (approved):
lgtm

PR #1371, opened 2026-05-22
SNOW-2918936 Incorrect parsing of timepart before unix time fixed

sfc-gh-mcepiga:
Summary

  - Fix GetTicksFromSecondAndNanosecond to negate the fractional part for negative (pre-1970) timestamps. Without
  this, a wire value like -0.876543300 would incorrectly add the nanoseconds instead of subtracting them, producing a
   time after the epoch instead of before it.
  - The decimal portion is parsed independently from the integer sign, so it must explicitly inherit the sign from
  the leading - character (srcVal.Buffer[srcVal.offset]).
  - Add unit tests covering pre-epoch timestamps with fractional seconds across NTZ and LTZ paths, including
  hardcoded wire-value assertions.

  Detail

  For timestamps before 1970-01-01, Snowflake returns a negative seconds.nanoseconds value (e.g. -0.876543300 for
  1969-12-31 23:59:59.1234567). FastParseInt64 only parses the digits after the dot, producing a positive value. The
  integer part (-0) correctly becomes 0, but the decimal part needs to be negated to produce the correct negative
  tick offset.

  Test plan

  - TestConvertDatetime — extended with 3 pre-epoch dates (1969, 1960, 1900) with fractional seconds
  - TestConvertTimestampLtzPreEpochWithFractionalSeconds — round-trips pre-epoch DateTimes through the LTZ wire
  format
  - TestConvertTimestampNtzPreEpochWithNegativeWireValue — hardcoded negative wire values asserted against exact
  expected DateTimes
  - Full SFDataConverterTest suite passes (177 tests, net6.0 + net9.0)

### Checklist
- [x] Code compiles correctly


sfc-gh-daniszewski (approved):
LGTM!

PR #1405, opened 2026-06-30
SNOW-3705005 Fix transitive builds with NativeLibrary added.

sfc-gh-mcepiga:
SNOW-3705005 Fix transitive builds with NativeLibrary added

  What it does: Differentiates Snowflake's own native libraries (tagged IsSnowflakeNativeLibrary="True") from
  third-party ones added transitively. Missing Snowflake libs produce an <Error> (build breaks), missing third-party
  libs produce a <Warning>. The <Copy> now skips non-existent files instead of failing.

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
   -  AWS Workload Identity Federation attestation now defaults to a SigV4-presigned `GetCallerIdentity` request.
       The STS `GetWebIdentityToken` path (returning a signed JWT) is available as an opt-in by setting the
       `SNOWFLAKE_ENABLE_AWS_WIF_OUTBOUND_TOKEN=true` environment variable.
-  -  Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
+  - Bug fix: Fixed non-windows builds with added NativeLibrary items in their transitively built projects that were no longer available to copy to output directory.
+  - Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
   -  Bug fix: Fixed incorrect DateTime conversion for timestamps preceding Unix epoch (1970-01-01) when fractional seconds are
     present.
 - v5.7.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@
       `SNOWFLAKE_ENABLE_AWS_WIF_OUTBOUND_TOKEN=true` environment variable.
   -  `CloseAsync` now throws the original exception on failure instead of wrapping it in an `AggregateException`.
   -  Bug fix: Fixed non-windows builds with added NativeLibrary items in their transitively built projects that were no longer available to copy to output directory.
-  -  Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
   -  Bug fix: `CloseAsync` method of `SnowflakeDbConnection` now resets its state to `Closed` on cancellation and `Broken` on failures.
   -  Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
   -  Bug fix: Fixed incorrect DateTime conversion for timestamps preceding Unix epoch (1970-01-01) when fractional seconds are
--- a/Snowflake.Data/build/Snowflake.Data.targets
+++ b/Snowflake.Data/build/Snowflake.Data.targets
@@ -1,21 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <!-- Determine the target architecture -->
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
   -  AWS Workload Identity Federation attestation now defaults to a SigV4-presigned `GetCallerIdentity` request.
       The STS `GetWebIdentityToken` path (returning a signed JWT) is available as an opt-in by setting the
       `SNOWFLAKE_ENABLE_AWS_WIF_OUTBOUND_TOKEN=true` environment variable.
-  -  Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
+  - Bug fix: Fixed non-windows builds with added NativeLibrary items in their transitively built projects that were no longer available to copy to output directory.
+  - Bug fix: Fixed session creation token leak when `GetSessionAsync` is cancelled.
```
sfc-gh-pbulawa: It seems that this does not belong to this PR. Can you verify?

sfc-gh-pbulawa (dismissed):
I don't see anywhere the usage of IsSnowflakeNativeLibrary. Can you confirm how this works?

Snowflake.Data/build/Snowflake.Data.targets
```diff
@@ -1,21 +1,23 @@
...
 
-  <!-- Copy native library for ALL Windows targets -->
-  <ItemGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
-    <NativeLibrary Include="$(MSBuildThisFileDirectory)..\runtimes\win-x64\native\sf_mini_core.dll" Condition="'$(SfMiniCoreArch)' == 'x64'" />
-    <NativeLibrary Include="$(MSBuildThisFileDirectory)..\runtimes\win-arm64\native\sf_mini_core.dll" Condition="'$(SfMiniCoreArch)' == 'arm64'" />
-    <NativeLibrary Include="$(MSBuildThisFileDirectory)..\runtimes\win-x86\native\sf_mini_core.dll" Condition="'$(SfMiniCoreArch)' == 'x86'" />
-  </ItemGroup>
+    <!-- Copy native library for ALL Windows targets -->
+    <ItemGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
+        <NativeLibrary Include="$(MSBuildThisFileDirectory)..\runtimes\win-x64\native\sf_mini_core.dll" Condition="'$(SfMiniCoreArch)' == 'x64'" IsSnowflakeNativeLibrary="True"/>
```
sfc-gh-mhofman, line 13: _SnowflakeNativeLibrary or SfNativeLibrary wouldn't grab anything other libraries introduce here as the NativeLibrary is a share one, is this by design?

CHANGELOG.md
sfc-gh-mhofman: I agree, please remove the 2nd changed line

sfc-gh-mhofman (changes requested):

CHANGELOG.md
sfc-gh-mcepiga: this line is in master already, it was line 11 in base revision, it is line 12 now.

Snowflake.Data/build/Snowflake.Data.targets
sfc-gh-mcepiga: could you rephrase this question? I'm having hard time understanding what you mean.

sfc-gh-mcepiga:
> I don't see anywhere the usage of IsSnowflakeNativeLibrary. Can you confirm how this works?

Snowflake.Data/build/Snowflake.Data.targets lines 19-20. This is signalling flag based on which we either warn or error out during the build - if we're missing "IsSnowflakeNativeLibrary"=true items to copy, we error out. If we're missing other items (e.g. when client application add some NativeLibrary items in their csprojs), we just warn and proceed.

Snowflake.Data/build/Snowflake.Data.targets
sfc-gh-mhofman: There's a suggestion not to rely on NativeLibrary which is commonly used one and may contain other artifacts than just our sf_core. So something SF specific instead of NativeLibrary would allow us to copy only the intended stuff. Does it make sense?

sfc-gh-mcepiga: Yeah, it does now. It would work, but I'm afraid to change it that way as that would be straying away from the "standard" way that some client integration may depend on in some manner already - they lived with NativeLibrary item for quite some time now and it *is* in fact, a part of the NativeLibrary dependency list of the application they build. If you think this concern is overstated - let me know, with it out of the way, we could go for custom Item definition.

sfc-gh-mhofman (approved):

PR #1415, opened 2026-07-15
SNOW-3768995 Propagated `CancellationToken` through the chunk download and parsing pipeline 

sfc-gh-mcepiga:
### Description

  - Propagated `CancellationToken` through the chunk download and parsing pipeline (`SFBlockingChunkDownloaderV3` →
  `IChunkParser` → `ArrowChunkParser`/`ReusableChunkParser`), enabling callers to cancel long-running result-set
  fetches.
  - Introduced `FastStreamWrapper` — a 32 KB buffered async stream wrapper used by `ReusableChunkParser` to support
  cancellation-aware byte-level reads without per-byte syscalls.
  - Added XML documentation for new public/internal APIs (`IChunkParser.ParseChunkAsync`, `FastStreamWrapper`).
  - Changed TestPutGetGcsDownscopedCredential test and associated wiremock setup to improve it's stability to reduce CI failures.

 
  ## Test plan
  - [x] Unit tests verify `FastStreamWrapper` buffering, EOF, and cancellation behavior
  - [x] Unit tests verify `ReusableChunkParser` throws `OperationCanceledException` mid-parse
  - [x] Integration tests verify cancellation during JSON and Arrow chunk downloads against live Snowflake
  - [x] Verify no performance regression on large result sets (multi-MB JSON chunks)

  Pull Request description generated with [Claude Code](https://claude.ai/code)

### Checklist
- [x] Code compiles correctly

### DIFF
--- a/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
+++ b/Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
@@ -172,7 +172,13 @@ namespace Snowflake.Data.Core
                 }
                 catch (Exception e)
                 {
-                    if (!downloadContext.CancellationToken.IsCancellationRequested && (maxRetry <= 0 || retryCount < maxRetry))
+                    if (e is OperationCanceledException || downloadContext.CancellationToken.IsCancellationRequested)
+                    {
+                        s_logger.Error($"Operation has been canceled. {e.Message}");
+                        throw;
+                    }
+
+                    if (maxRetry <= 0 || retryCount < maxRetry)
                     {
                         s_logger.Debug($"Retry {retryCount}/{maxRetry} of parse stream to chunk error: " + e.Message);
                         retry = true;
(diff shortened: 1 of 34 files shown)


Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
```diff
@@ -131,105 +117,104 @@ public async Task<BaseResultChunk> GetNextChunkAsync()
...
-                            if (backOffInSec > HttpUtil.MAX_BACKOFF)
-                            {
-                                backOffInSec = HttpUtil.MAX_BACKOFF;
-                            }
-                        }
-                        else
+                        await ParseStreamIntoChunkAsync(stream, chunk, downloadContext.CancellationToken).ConfigureAwait(false);
+                    }
+                }
+                catch (Exception e)
```
sfc-gh-pbulawa, line 173: When OperationCanceledException is received, it will also be retried. Shouldn't we handle it separately?

sfc-gh-pbulawa (changes requested):

Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
sfc-gh-mcepiga: well, it won't, because of the awaited backoff in line 182 but you're right in general and we should definitely be more explicit in handling this. Will introduce explicit catch handling bit for that case.

sfc-gh-pbulawa (approved):

PR #1420, opened 2026-07-27
SNOW-3768995 Chunk reads timeouts

sfc-gh-mcepiga:
## Summary
  - Adds `IdleTimeoutReadStream`, a stream decorator that enforces configurable idle and per-read timeouts on async
  reads during result set chunk downloads
  - Introduces two environment variables: `SF_CHUNK_DOWNLOAD_IDLE_TIMEOUT` (default 180s) detects stalled connections
   between reads; `SF_CHUNK_DOWNLOAD_READ_TIMEOUT` (default disabled) sets a per-read deadline
  - Refactors retry/exception handling in `SFBlockingChunkDownloaderV3` to use exception filters for clearer control
  flow
  - Adds comprehensive environment variables documentation (`doc/Client/EnvironmentVariables.md`)

  ## Motivation
  Long-running queries with large result sets can hang indefinitely when the server stops sending chunk data
  mid-transfer. Previously there was no mechanism to detect or recover from stalled connections during chunk parsing.
   This change provides configurable timeouts that abort stuck downloads, allowing the retry logic to re-attempt the
  transfer.
  ## Design Notes
  - The idle timeout fires **between** read operations — it detects when no data has arrived for N seconds after the
  last successful read. It does not interrupt a single blocked read call.
  - The read timeout applies to **each individual** `ReadAsync` call independently via a linked
  `CancellationTokenSource`.
  - For full protection against all stall scenarios, both timeouts should be configured. The documentation notes this
   explicitly.
  - Environment variables are read once at class load (`static readonly`) to avoid per-chunk syscall overhead.
  - Set either timeout to `0` to disable it.

  ## Test plan
  - [x] Unit tests for `IdleTimeoutReadStream` (11 tests): normal reads, idle timeout fires between reads, idle
  resets after successful reads, read timeout fires, caller cancellation propagation, both timeouts configured, both
  disabled as passthrough, dispose safety, sync read delegation
  - [x] Unit tests for `EnvironmentFacade.GetTimeSpan` (12 test cases): valid seconds, null/empty/whitespace,
  overflow, invalid formats, default TimeSpan.Parse mode
  - [x] Unit tests for `GetInt`, `GetBool`, `GetString` with edge cases
  - [x] Manual validation on slow/throttled network with large result sets
  - [x] Verify `SF_CHUNK_DOWNLOAD_IDLE_TIMEOUT=0` disables the feature (no behavioral change from previous driver
  versions)

  Pull Request description generated with [Claude Code](https://claude.ai/code)

### Checklist
- [x] Code compiles correctly


Snowflake.Data/Core/IdleTimeoutReadStream.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            throw new TimeoutException($"Read took longer than {_readTimeout.TotalSeconds} seconds", ex);
+        }
+        finally
+        {
+            readCts?.Dispose();
+        }
+    }
+
+    private void ResetIdleTimer()
```
sfc-gh-pbulawa, line 86: Shouldn't we reset the `_idleExpired` too?.

Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
```diff
@@ -170,36 +174,30 @@ private async Task<BaseResultChunk> DownloadChunkAsync(DownloadContextV3 downloa
...
+                }
                 catch (Exception e)
                 {
-                    if (e is OperationCanceledException || downloadContext.CancellationToken.IsCancellationRequested)
+                    if (maxRetry > 0 && retryCount >= maxRetry)
                     {
-                        s_logger.Error($"Operation has been canceled. {e.Message}");
-                        throw;
+                        s_logger.Error($"Failed retries of parse stream to chunk error: {e.Message}");
+                        throw new Exception($"Parse stream to chunk error: {e.Message}");
```
sfc-gh-pbulawa, line 187: Shouldn't we use more specific exception here?

sfc-gh-pbulawa (changes requested):

Snowflake.Data/Core/IdleTimeoutReadStream.cs
sfc-gh-mcepiga: no, this is a sort of heartbeat reset - it keeps the idle clock running. If the idle time gets exceeded, there's no heartbeat as the patient is dead. Maybe this method could be named better - any ideas?

Snowflake.Data/Core/SFBlockingChunkDownloaderV3.cs
sfc-gh-mcepiga: this is not new, it's just reordered from old revision: lines 200-202 into lines 186-188.

Technically changing this exception would be a breaking change. Since we've had one recently, I'd prefer not to go that route

sfc-gh-pbulawa (approved):

PR #1422, opened 2026-07-28
SNOW-3857631 Write token cache as UTF-8 without BOM on Unix/macOS

sfc-gh-mhofman:
## Summary
- Stop writing a UTF-8 BOM when persisting the token cache (and log append path) on Linux/macOS, matching Windows and other drivers so shared cache JSON stays parseable.

## Test plan
- [x] Unit test `TestWriteAllTextDoesNotEmitUtf8Bom` asserts raw bytes have no `EF BB BF` and round-trips via `ReadAllText`
- [ ] Verify on Linux/macOS that a freshly written credential cache file starts with `{`, not a BOM

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   -  Bug fix: Fixed incorrect DateTime conversion for timestamps preceding Unix epoch (1970-01-01) when fractional seconds are
     present.
   -  Bug fix: Fixed an unnecessary second PUT (stage re-resolution) per file during GCS uploads when the server scopes upload credentials with an access token.
+  -  Bug fix: Token cache file on Linux/macOS is now written as UTF-8 without a BOM for cross-driver compatibility.
 - v5.7.0
     - Improved input handling in `ChangeDatabase` by using parameterized queries.
     - Improved input validation in `QueryResultsAwaiter` with stricter UUID format checks.
--- a/Snowflake.Data.Tests/UnitTests/Tools/UnixOperationsTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/Tools/UnixOperationsTest.cs
@@ -320,6 +320,22 @@ namespace Snowflake.Data.Tests.UnitTests.Tools
         }
 
         [SFFact(SkipCondition.SkipOnWindows)]
+        public void TestWriteAllTextDoesNotEmitUtf8Bom()
+        {
+            // arrange
+            var content = "{\"token\":\"abc\"}";
+            var filePath = Path.Combine(_fixture.WorkingDirectory, $"no_bom_{Path.GetRandomFileName()}");
+
+            // act
+            _fixture.UnixOperations.WriteAllText(filePath, content, _ => { });
+
+            // assert
+            var bytes = File.ReadAllBytes(filePath);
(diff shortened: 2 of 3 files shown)


Snowflake.Data.Tests/UnitTests/Tools/UnixOperationsTest.cs
```diff
@@ -319,6 +319,22 @@ public void TestNestedDir()
...
+            // arrange
+            var content = "{\"token\":\"abc\"}";
+            var filePath = Path.Combine(_fixture.WorkingDirectory, $"no_bom_{Path.GetRandomFileName()}");
+
+            // act
+            _fixture.UnixOperations.WriteAllText(filePath, content, _ => { });
+
+            // assert
+            var bytes = File.ReadAllBytes(filePath);
+            Assert.False(bytes.Length >= 3 && bytes[0] == 0xEF && bytes[1] == 0xBB && bytes[2] == 0xBF);
```
sfc-gh-mcepiga: If we have a regression and this test fails at this assert, it will produce very unhelpful message "Expected False but was True". (also if it fails because bytes.Length < 3 this is unexpected result which may lead to false negative in regression coverage). 

This is unit test, so it's not the end of the world as any failure should be easily reproducible, but we could do a little better.

sfc-gh-pbulawa (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
   -  Bug fix: Fixed incorrect DateTime conversion for timestamps preceding Unix epoch (1970-01-01) when fractional seconds are
     present.
   -  Bug fix: Fixed an unnecessary second PUT (stage re-resolution) per file during GCS uploads when the server scopes upload credentials with an access token.
+  -  Bug fix: Token cache file on Linux/macOS is now written as UTF-8 without a BOM for cross-driver compatibility.
```
sfc-gh-mcepiga: it won't be included in v6.0.0, please put it in 6.1.0 section

Snowflake.Data/Core/Tools/UnixOperations.cs
```diff
@@ -14,6 +14,7 @@ internal class UnixOperations
     {
         public static readonly UnixOperations Instance = new UnixOperations();
         private static readonly SFLogger s_logger = SFLoggerFactory.GetLogger<UnixOperations>();
+        private static readonly Encoding Utf8NoBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
```
sfc-gh-mcepiga, line 17: shouldn't this also be used in ReadAllText method?

sfc-gh-mcepiga (changes requested):

CHANGELOG.md
sfc-gh-mhofman: done

Snowflake.Data.Tests/UnitTests/Tools/UnixOperationsTest.cs
sfc-gh-mhofman: fixed

Snowflake.Data/Core/Tools/UnixOperations.cs
sfc-gh-mhofman: StreamReader strips BOM when detected so it's intentional for driver to read old written files.

sfc-gh-mcepiga (approved):

Issue #1428, opened 2026-08-18
SNOW-3977560: TIMESTAMP_TZ values with non-whole-hour UTC offsets are silently corrupted when converted to DateTimeOffset


sdt-rkety:
### Summary
Reading a TIMESTAMP_TZ column whose UTC offset is not a whole number of hours (e.g. +05:30) returns a
DateTimeOffset with the correct absolute UTC instant but the WRONG local time and offset — silent data
corruption, not an exception.

### Root cause
Snowflake.Data/Core/SFDataConverter.cs, ConvertToDateTimeOffset, TIMESTAMP_TZ case:

    int offset = FastParser.FastParseInt32(...);   // wire value = actual offset-in-minutes + 1440
    TimeSpan offSetTimespan = new TimeSpan((offset - 1440) / 60, 0, 0);
    return new DateTimeOffset(UnixEpoch.Ticks + GetTicksFromSecondAndNanosecond(timeVal), TimeSpan.Zero)
        .ToOffset(offSetTimespan);

`(offset - 1440) / 60` uses integer division. For a +05:30 offset (330 minutes, wire value 1770):
(1770 - 1440) / 60 = 330 / 60 = 5 — the 30-minute remainder is silently discarded, producing a
TimeSpan of exactly +05:00:00 instead of +05:30:00.

Because the absolute UTC instant is computed correctly and independently (from timeVal), and only the
re-expression offset is wrong, the bug is easy to miss: the UTC instant is right, but .ToOffset() then
re-expresses it at the truncated offset, shifting the local time to match. E.g. source "12:00:00.000+05:30"
(06:30 UTC) is returned as a DateTimeOffset equal to "11:30:00.000+05:00" — same instant, wrong local
time/offset pair.

### Repro
Given a TIMESTAMP_TZ column containing '2028-02-29 12:00:00.000 +05:30':

    using var cmd = connection.CreateCommand();
    cmd.CommandText = "SELECT '2028-02-29 12:00:00.000 +05:30'::TIMESTAMP_TZ AS ts";
    using var reader = await cmd.ExecuteReaderAsync();
    await reader.ReadAsync();
    var value = (DateTimeOffset)reader.GetValue(0);
    // Expected: 2028-02-29 12:00:00 +05:30
    // Actual:   2028-02-29 11:30:00 +05:00

### Impact
[...]

PR #1429, opened 2026-08-24
SNOW-3977560 Fix TIMESTAMP_TZ sub-hour offset truncation in JSON result format

sfc-gh-dszmolka:
This is for https://github.com/snowflakedb/snowflake-connector-net/issues/1428

### Description

**Problem:** In the JSON result format, `ConvertToDateTimeOffset` in `SFDataConverter.cs` built the TIMESTAMP_TZ timezone offset with integer whole-hour division (`new TimeSpan((offset - 1440) / 60, 0, 0)`), discarding the sub-hour minutes, so offsets that are not a whole number of hours — `+05:30`, `-03:30`, `+05:45`, `+12:45` — were truncated to the whole hour.

**Impact:** The returned `DateTimeOffset` kept the correct UTC instant (`.UtcDateTime`) but had a wrong `.Offset`, `.DateTime`, and `.LocalDateTime`; for example a value that should be `2028-02-29 12:00:00 +05:30` came back as `2028-02-29 11:30:00 +05:00`. Only the JSON path was affected — Arrow was already correct — so the corruption showed up intermittently depending on which result format the server returned.

**Fix:** Use `TimeSpan.FromMinutes(offset - 1440)`, which preserves the full-minute offset and matches the Arrow decode, array-bind decode, and C#-to-wire encode paths in this same driver that were already correct.

**Test:** Added the JSON-path theory `TestConvertTimestampTzToDateTimeOffsetPreservesSubHourOffset` covering `+05:30`, `-03:30`, `+05:45`, `+12:45`, `+01:00`, and UTC and asserting `.Offset`, `.UtcDateTime`, and `.DateTime` — it fails on the four sub-hour cases with the old code and passes after the fix; also extended the existing Arrow test with `+05:30`/`-03:30` for symmetry. Full unit-test suite: 1287 passed, 6 skipped, 0 failed.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing (`dotnet test`) — full unit-test suite green (integration tests require a live account)

### DIFF
--- a/Snowflake.Data.Tests/UnitTests/SFDataConverterTest.cs
+++ b/Snowflake.Data.Tests/UnitTests/SFDataConverterTest.cs
@@ -334,6 +334,13 @@ namespace Snowflake.Data.Tests.UnitTests
             return (tickDiff / 10000000.0m).ToString(CultureInfo.InvariantCulture);
         }
 
+        private string DateTimeToTzWireFormat(DateTime utcDateTime, int offsetMinutes)
+        {
+            // TIMESTAMP_TZ wire format: "<seconds since epoch, in UTC> <offsetMinutes + 1440>"
+            var seconds = DateTimeToLtzWireFormat(utcDateTime);
+            return $"{seconds} {offsetMinutes + 1440}";
+        }
+
         [SFFact]
         public void TestConvertTimestampLtzToDateTimeOffsetWithLocalTimezone()
         {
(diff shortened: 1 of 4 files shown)


Snowflake.Data.Tests/UnitTests/SFDataConverterTest.cs
```diff
@@ -334,6 +334,13 @@ private string DateTimeToLtzWireFormat(DateTime utcDateTime)
             return (tickDiff / 10000000.0m).ToString(CultureInfo.InvariantCulture);
         }
 
+        private string DateTimeToTzWireFormat(DateTime utcDateTime, int offsetMinutes)
```
sfc-gh-mcepiga: nit: inline this method, this is very simple, used only once and disrupts the structure of other test methods, there's no reason not to inline it.

sfc-gh-mcepiga (approved):
approved, nit comment is nice to have.

Snowflake.Data.Tests/UnitTests/SFDataConverterTest.cs
sfc-gh-dszmolka: right - inlined it

